PR #1031, opened 2015-02-04
Regression test fix

sidheshenator:
Three changes have been made
Diff Util used the HTML report files comparison instead of the chunk comparisons.
ReportGenerator.java is modified to sort the keyword hits and hash set hits so that the report is in the same order each time.
Comments are added in Bundle.properties to notify dependence of regression testing script.


sidheshenator:
The output of diff util is written to HTML-Report-Diff.txt.

PR #1038, opened 2015-02-10
Multiple aliases for blackboard_attributes removed

sidheshenator:
blackboard_attributes had multiple aliases in he SQL query (att1. att2, att3). Replaced it by single alias- att.


rcordovano:
This is not correct. There are three different attribute rows to be fetched.

PR #1073, opened 2015-02-20
android ja localization/translation update

mrtizmo:


rcordovano:
Does not automatically merge. Please address and resubmit.

PR #1084, opened 2015-02-24
Moved the _copy_diff_files() call to appropriate place

sidheshenator:


sidheshenator:
Initially, _copy_diff_files was called only once after ingesting the last image from the image list. Now, it is called for every image in the image list.

PR #1088, opened 2015-03-05
Python ingest modules settings

sidheshenator:


rcordovano:
Please modify as discussed: Jython ingest module job settings are serialized to and deserialized from files of the same name, and the deserialization code branches to use the correct object stream based on settings file name rather than using exceptions to determine behavior.

PR #1098, opened 2015-03-11
Duplicate tagging isue

sidheshenator:


Core/src/org/sleuthkit/autopsy/actions/AddContentTagAction.java
```diff
@@ -109,7 +112,16 @@ else if (file.getName().equals("..")) {
                         continue;
                     }                    
                 }
-                
+                // check if the same tag is being added for the same abstract file.
+                List<ContentTag> contentTagList = Case.getCurrentCase().getServices().getTagsManager().getAllContentTags();
```
bcarrier: Use getContentTagsByContent(file) instead of getAllContentTags().

sidheshenator:
Implemented your comment.

PR #1107, opened 2015-03-17
View File in Directory available for local and derived files

sidheshenator:


millmanorama:
I suspect if you are going to make LocalFileNode extend FileNode there is more that could be refactored to reduce duplicated code.

sidheshenator:
@millmanorama  - Refactored that class.

PR #1110, opened 2015-03-18
Rule Name field made optional

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSetRulePanel.java
```diff
@@ -135,12 +135,18 @@ private void populatePathFilterComponents(FilesSet.Rule rule) {
...
         // The rule must have a name.
+        // If rule name is absent, set it as the Name Pattern since nameTextField cannot be left empty.
         if (this.ruleNameTextField.getText().isEmpty()) {
-            NotifyDescriptor notifyDesc = new NotifyDescriptor.Message(
-                    NbBundle.getMessage(FilesSetPanel.class, "FilesSetRulePanel.messages.filesSetRulesMustBeNamed"),
-                    NotifyDescriptor.WARNING_MESSAGE);
-            DialogDisplayer.getDefault().notify(notifyDesc);
-            return false;
+            String ruleName = this.nameTextField.getText();
+            if (ruleName != null && !ruleName.isEmpty()) {
```
millmanorama: it is probably not worth changing this pull request But i have been using StringUtils.notBlank() with a static import instead of the `ruleName  != null && !ruleName.isEmpty()`. I think it is more concise and readable.

in this example line 141 would become:
`if (notBlank(ruleName){`

sidheshenator:
@millmanorama using StringUtils.notBlank is useful in case of string containing only blank spaces need to be handled. StringUtils.isNotBlank(" ") = false.
I'll change it.

rcordovano:
Please modify the code to permit an empty string in the place of the rule name.

PR #1116, opened 2015-03-19
Recent Case Button and Label customized before loading cuebanner and aft...

sidheshenator:
...er closing case


rcordovano:
The CueBannerPanel.customizeComponents() method should not be public. Exposing this method breaks the encapsulation of the class and puts an unnecessary burden on the clients of CueBannerPanel. Instead, CueBannerPanel.customizeComponents() needs to be a private method called in the CueBannerPanel constructor after the generated initComponents() method is called.

PR #1117, opened 2015-03-19
Interesting file rule names optional

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/interestingitems/InterestingItemDefsPanel.java
```diff
@@ -306,7 +306,7 @@ private void doFilesSetRuleDialog(FilesSet.Rule selectedRule) {
                 rules.remove(selectedRule.getName());
             }
             FilesSet.Rule newRule = new FilesSet.Rule(panel.getRuleName(), panel.getFileNameFilter(), panel.getMetaTypeFilter(), panel.getPathFilter());
-            rules.put(newRule.getName(), newRule);
+            rules.put(newRule.toString(), newRule);
```
sidheshenator: HashMap rules<String, Rule> needs a unique key. newRule.toString is a unique string and an ideal key.

rcordovano:
Please use hashCode (inherited or overridden) as the key for storing rules in the hash map instead of toString. Also, please leave the check for a null rule name in the Rule constructor - it can be empty, but not null. The javadocs for the ctor and getName() should mention this. Additionally, the code in InterstingItemDefsManager that synthesizes a rule name should be removed.

PR #1119, opened 2015-03-19
Interesting file rule names optional

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/interestingitems/InterestingItemDefsPanel.java
```diff
@@ -306,7 +306,7 @@ private void doFilesSetRuleDialog(FilesSet.Rule selectedRule) {
                 rules.remove(selectedRule.getName());
             }
             FilesSet.Rule newRule = new FilesSet.Rule(panel.getRuleName(), panel.getFileNameFilter(), panel.getMetaTypeFilter(), panel.getPathFilter());
-            rules.put(newRule.getName(), newRule);
+            rules.put(Integer.toString(newRule.hashCode()), newRule);
```
sidheshenator: Used hashCode (inherited) as the key for storing rules in the hash map instead of toString

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
```diff
@@ -150,8 +150,8 @@ public String toString() {
          * @param pathFilter A file path filter, may be null.
          */
         Rule(String ruleName, FileNameFilter fileNameFilter, MetaTypeFilter metaTypeFilter, ParentPathFilter pathFilter) {
-            if ((ruleName == null) || (ruleName.isEmpty())) {
-                throw new IllegalArgumentException("Interesting files set rule name cannot be null or empty");
+            if (ruleName == null) {
```
sidheshenator: Left the check for a null rule name in the Rule constructor

Core/src/org/sleuthkit/autopsy/modules/interestingitems/InterestingItemDefsManager.java
```diff
@@ -402,11 +402,6 @@ private static void readFilesSet(Element setElem, Map<String, FilesSet> filesSet
         private static String readRuleName(Element elem) {
             // The rule must have a name. 
             String ruleName = elem.getAttribute(FilesSetXML.NAME_ATTR);
-            if (ruleName.isEmpty()) {
```
sidheshenator: The code in InterstingItemDefsManager that synthesizes a rule name is removed.

PR #1138, opened 2015-03-25
Duplicate rule set names not allowed

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
```diff
@@ -617,6 +618,7 @@ public boolean textMatches(String subject) {
              */
             CaseInsensitivePartialStringComparisionMatcher(String textToMatch) {
                 this.textToMatch = textToMatch;
+                this.pattern = Pattern.compile(Pattern.quote(textToMatch), Pattern.CASE_INSENSITIVE);
```
sidheshenator: Pattern compiled only once rather than for every function call.

Core/src/org/sleuthkit/autopsy/modules/interestingitems/InterestingItemDefsPanel.java
```diff
@@ -255,6 +256,14 @@ private void doFileSetsDialog(FilesSet selectedSet) {
             option = JOptionPane.showConfirmDialog(null, panel, NbBundle.getMessage(FilesSetPanel.class, "FilesSetPanel.title"), JOptionPane.OK_CANCEL_OPTION, JOptionPane.PLAIN_MESSAGE);
         } while (option == JOptionPane.OK_OPTION && !panel.isValidDefinition());
 
+        // If rule set with same name already exists, do not add to the filesSets hashMap.
```
sidheshenator: Rule set name duplicity check.

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
```diff
@@ -441,7 +441,7 @@ public boolean textMatches(String textToMatch) {
              */
             @Override
             public boolean passes(AbstractFile file) {
-                return this.textMatches(file.getParentPath());
+                return this.textMatches(file.getParentPath() + "/");
```
sidheshenator: compiled pattern can match the last directory in the path.

PR #1139, opened 2015-03-25
used non deprecated complete(ReportStatus)

sidheshenator:


Core/src/org/sleuthkit/autopsy/casemodule/StartupWindow.java
```diff
@@ -63,7 +63,7 @@ private void init() {
         setLocation((screenDimension.width - w) / 2, (screenDimension.height - h) / 2);
         
         CueBannerPanel welcomeWindow = new CueBannerPanel();
-        welcomeWindow.refresh();
+        CueBannerPanel.refresh();
```
sidheshenator: C:\autopsy\Core\src\org\sleuthkit\autopsy\casemodule\StartupWindow.java:66: warning: [static] static method should be qualified by type name, CueBannerPanel, instead of by an expression
welcomeWindow.refresh(); compiler WARNING fixed.

PR #1145, opened 2015-03-27
getOneHitPerObject() optimized

sidheshenator:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/QueryResults.java
```diff
@@ -167,17 +166,26 @@ KeywordSearchQuery getQuery() {
...
+     * @return Collection<KeywordHit> containing KeywordHits with lowest SolrObjectID-ChunkID pairs.
+     */
+    private Collection<KeywordHit> getOneHitPerObject(Keyword keyword) {
+
+        HashMap<Long, KeywordHit> hits = new HashMap();
+
+        // create a list of KeywordHits. KeywordHits with lowest chunkID is added the the list.
+        for(KeywordHit hit: getResults(keyword)) {
+            if(!hits.containsKey(hit.getSolrObjectId())) {
+                hits.put(hit.getSolrObjectId(), hit);
```
millmanorama: I think this if chain becomes a one liner if you use Map.merge() with an appropriate remapper function

PR #1155, opened 2015-03-30
fixed naming consistency - BlackboardDumpErrors instead of SortedErrors

sidheshenator:


rcordovano:
SortedErrors is not a dump of blackboard errors - we don't have such a thing. There should be an attempt to diff the SEVERE log messages generated in the gold and the test runs.

PR #1159, opened 2015-03-31
Save the last export location

sidheshenator:


rcordovano:
Going to go with a different solution, i.e., per user settings.

PR #1162, opened 2015-04-01
Solr upgrade

esaunders:
This pull request upgrades Solr to 4.9.1, Jetty to 8.1 and Tika to 1.5


kefir-:
Tika 1.5 was released more than a year ago, and solr 4.9.1 was released in September 2014. Neither of these are the latest releases, which are Tika v1.7 and Solr v.4.10.4 or v.5.0.0. As I read the diff, the previous versions were Tika 1.2 and Solr 4.0.0, so it'd seem to make more sense to go "all the way" (at least to solr 4.10.4 if 5.0.0 means more work) now that the upgrade is done anyway.

esaunders:
The reason Solr 4.9.1 and its related dependencies was chosen was to base
our development on something that had a chance to have had some use in the
wild. If we have time down the line we can consider upgrading to a later
version. Looking at the release notes I don't see anything in the newer
versions that make the later versions more attractive. The reasons for not
going to 5.0.0 at this point were a) it had only just been released when we
started looking into upgrading and b) there were concerns about possible
backwards compatibility issues.

PR #1166, opened 2015-04-02
Messaging prototype

rcordovano:


Core/nbproject/project.properties
```diff
@@ -1,6 +1,7 @@
+file.reference.activemq-all-5.11.1.jar=C:\\autopsy\\Core\\release\\modules\\ext\\activemq-all-5.11.1.jar
```
karlmortensen: Relative path?

rcordovano: I thought I had fixed that already. Good catch, thanks!

PR #1185, opened 2015-04-09
Image gallery fixes

bcarrier:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -391,14 +391,6 @@ private boolean initializeDB() {
             return false;
         }
 
-        // @@@ BC: I don't think this is needed because it is the primary key. 
-        try (Statement stmt = con.createStatement()) {
```
millmanorama: does it hurt to have this.  presumably it is a know op if the primary key counts as an index. but yes it seems redundant

bcarrier: My impression is that it gets maintained as an additional index and we are therefore wasting time on inserts  and space.

PR #1212, opened 2015-04-27
Messaging

rcordovano:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -223,8 +205,16 @@ private Case(String name, String number, String examiner, String configFilePath,
         this.caseType = type;
         this.db = db;
         this.services = new Services(db);
-        this.eventSubscriber = new EventSubscriber();
-        this.messenger = new Messenger(this.name, eventPublisher);
+        if (CaseType.MULTI_USER_CASE == this.caseType) {
+            try {
+                eventPublisher.openRemoteEventChannel(name);
```
karlmortensen: Would it be better to do this in an init() after the case is open? That way we're not subscribed for incoming messages until we're fully constructed. And, if this fails, the user is unable to edit anything with the case, right? e.g. they can't add a data source or anything... right?

Core/src/org/sleuthkit/autopsy/events/AutopsyEventPublisher.java
```diff
@@ -83,29 +120,22 @@ synchronized public void removeSubscriber(Collection<String> eventNames, Autopsy
...
-                try {
-                    subscriber.receiveEvent(event);
-                } catch (Exception ex) {
-                    logger.log(Level.SEVERE, "Exception thrown by subscriber", ex);
-                }
-            }
+    synchronized public void publish(AutopsyEvent event) throws JMSException {
+        event.setSourceType(AutopsyEvent.SourceType.LOCAL);
+        localPublisher.publish(event);
+        if (null != remotePublisher) {
```
karlmortensen: I guess I don't understand why we publish it back locally too. I was under the impression that only the remote node(s) would need these events. What am I missing?

Core/src/org/sleuthkit/autopsy/events/LocalEventPublisher.java
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.events;
+
+import java.beans.PropertyChangeListener;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
```
karlmortensen: Which logger do we want to use?

Core/src/org/sleuthkit/autopsy/casemodule/Bundle.properties
```diff
@@ -136,6 +136,7 @@ Case.createCaseDir.exception.cantCreate=Cannot create case dir\: {0}
 Case.createCaseDir.exception.cantCreateCaseDir=Could not create case directory\: {0}
 Case.createCaseDir.exception.cantCreateModDir=Could not create modules output directory\: {0}
 Case.createCaseDir.exception.gen=Could not create case directory\: {0}
+Case.OpenEventChannel.ErrMsg=Case opening process failed to connect to nodes collaborating on case {0}.
```
karlmortensen: While ErrMsg follows the existing style, it doesn't save much typing. :-)

Core/src/org/sleuthkit/autopsy/events/AutopsyEventPublisher.java
rcordovano: I'm not sure what you mean by "publish it back locally."  When an object, say the IngestManager, publishes an event (fires a property change) it needs to be sent to the subscribers on the local or source node, as well as to the subscribers on other nodes, the remote nodes. The LocalEventPublisher (an improved version of PropertyChangeSupport) handles the publication of events to subscribers on the local or source node via calls to propertyChange(), and the RemoteEventPublisher handles the publication of outgoing events to the remote nodes by sending JMS messages. This is what happens in AutopsyEventPublisher.publish(). When a node receives a JMS event message, it needs to be published to the subscribers on the local node. That's why the RemoteEventPublisher gets a reference to the LocalEventPublisher which, like PropertyChangeSupport, encapsulates the local subscribers. The action happens in RemoteEventPublisher.MessageReceiver.onMessage() in a JMS thread. If I have not answered your question, please rephrase it.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
rcordovano: Actually, the AutopsyEventPublisher for the Case class is static - this allows subscribers to subscribe once, instead of every time a new case is opened. So there actually is no leaking of a partially constructed object happening. However, some Case events are published in static methods, others are published in instance methods, and some are published by instance methods that call the static changeCase() method, so it is all rather messy and confusing. Originally, I had the calls to RemoteEventListener.openEventChannel and closeEventChannel() in the static changeCase() method. I will put them back there, instead of in the constructor/int()/closeCase() troika of instance methods. All roads lead to changeCase().

Note that I will do this in the next pull request for this feature, along with other work on the Case class and its events.

Core/src/org/sleuthkit/autopsy/casemodule/Bundle.properties
rcordovano: You speak the truth! I can hear myself saying the same thing about other abbreviations, like "evt" for "event." I should change it for the sake of consistency with self! I can get that in the next pull request for this feature.

Core/src/org/sleuthkit/autopsy/events/AutopsyEventPublisher.java
karlmortensen: Okay. This makes sense. I had thought that it had already been published locally and we might be publishing it a second time, but that is not the case.

PR #1237, opened 2015-05-08
Fixed compare() override. TEXT is now a sortedField

sidheshenator:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/LuceneQuery.java
```diff
@@ -477,7 +477,11 @@ public int compare(SolrDocument left, SolrDocument right) {
                 rightID = rightID.substring(0, index);
             }
 
-            return leftID.compareTo(rightID);
+            if(Integer.parseInt(leftID) < Integer.parseInt(rightID))
+                return -1;
+            if(Integer.parseInt(leftID) > Integer.parseInt(rightID))
```
millmanorama: if the point here was to compares the parsed integer values, rather than strings you can still use compareTo:

Integer.parseInt(leftID).compareTo(Integer.parseInt(rightID))

PR #1266, opened 2015-05-20
File type identification

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/sevenzip/SevenZipIngestModule.java
```diff
@@ -657,24 +661,7 @@ private boolean isSupported(AbstractFile file) {
      * @return true if zip file, false otherwise
      */
     private boolean isZipFileHeader(AbstractFile file) {
-        if (file.getSize() < readHeaderSize) {
-            return false;
-        }
-
```
millmanorama: maybe this is what tika does under the hood, but this method name is misleading now...

sidheshenator: @millmanorama  I'll rename it to isZipFile().

PR #1267, opened 2015-05-21
Add data source

eugene7646:
* Added case folder and data source path validation for multi-user cases
* Path to multi-user cases and data sources can't be on C: drive
* Added WizardPathValidator interface for custom path validators
* Modified all file/folder selection panels


bcarrier:
Re @kefir comments: The idea is a basic sanity check for new "multi-user" cases that will be part of Autopsy 3.2.   

A C: check is easy and mostly effective. Yes, you could have a local D: and E: drive that it would not check, but Java doesn't provide us with an easy way to differentiate the true source of any drive letter.  So, in that regard, a check for C: may not be restrictive enough. 

For some of our initial setups though, this check is too restrictive because you could have a folder on C: that is mapped to shared drive. 

This can certainly be expanded on over time.  The basic idea was a simple sanity check. 

The current requirement is that all systems working in a multi-user cluster must have access to the same network-based storage and it must be at the same location for each node. 

With regard to performance, I'd propose that the image should always be available in a shared location from all nodes and we could then modify Autopsy in the future to make a local copy of it from the shared location when the case is opened.

PR #1268, opened 2015-05-21
Image resizing exception handled. Image cropped instead

sidheshenator:


Core/src/org/sleuthkit/autopsy/coreutils/ImageUtils.java
```diff
@@ -269,7 +270,14 @@ private static BufferedImage generateIcon(Content content, int iconSize) {
             BufferedImage biScaled = ScalrWrapper.resizeFast(bi, iconSize);
 
             return biScaled;
-        } catch (OutOfMemoryError e) {
+        } catch (IllegalArgumentException e) {
+            // if resizing does not work due to extremely small height/width ratio,
+            // crop the image instead.
+            BufferedImage biCropped = ScalrWrapper.cropImage(bi, iconSize > bi.getWidth() ?
+                    bi.getWidth() : iconSize, iconSize > bi.getHeight() ? bi.getHeight() : iconSize);
```
millmanorama: i would use Math.min here over a ternary

sidheshenator: Yeah. That's way more readable. Thanks.  


     On Thursday, May 21, 2015 11:41 AM, Jonathan Millman <notifications@github.com> wrote:
   

 In Core/src/org/sleuthkit/autopsy/coreutils/ImageUtils.java:> @@ -269,7 +270,14 @@ private static BufferedImage generateIcon(Content content, int iconSize) {
>              BufferedImage biScaled = ScalrWrapper.resizeFast(bi, iconSize);
>  
>              return biScaled;
> -        } catch (OutOfMemoryError e) {
> +        } catch (IllegalArgumentException e) {
> +            // if resizing does not work due to extremely small height/width ratio,
> +            // crop the image instead.
> +            BufferedImage biCropped = ScalrWrapper.cropImage(bi, iconSize > bi.getWidth() ?
> +                    bi.getWidth() : iconSize, iconSize > bi.getHeight() ? bi.getHeight() : iconSize);
i would use Math.min here over a ternary—

PR #1279, opened 2015-05-27
Bug fix in new sample Python FileIngestModule

jgrover:
Added an empty settings object as an argument when creating a SampleJythonFileIngestModule object. Was getting 'incorrect number of arguments error' without it.


jgrover:
Turns out this is not a bug.  I had some old sample code that didn't mix well with the new sample code.  Disregard.

PR #1285, opened 2015-05-29
Catch case where Content is null

APriestman:


rcordovano:
We are going to investigate this further.

PR #1295, opened 2015-06-01
Add ds wizard

bcarrier:


rcordovano:
Leaving restoring this capability out for now. A Basis Technology JIRA story to consider options suitable for apps built on an Autopsy platform has been created.

PR #1325, opened 2015-06-08
Collaboration monitor

rcordovano:


Core/src/org/sleuthkit/autopsy/casemodule/CollaborationMonitor.java
```diff
@@ -0,0 +1,700 @@
...
+
+        /**
+         * Monitor the availability of collaboration resources
+         */
+        @Override
+        public void run() {
+            synchronized (lock) {
+                CaseDbConnectionInfo dbInfo = UserPreferences.getDatabaseConnectionInfo();
+                try {
+                    DriverManager.getConnection("jdbc:postgresql://" + dbInfo.getHost() + ":" + dbInfo.getPort() + "/" + "postgres", dbInfo.getUserName(), dbInfo.getUserName()); // NON-NLS
```
karlmortensen: It would be nice to be able to use the connection pool to do this.

karlmortensen: And the cut and paste error: repeated dbInfo.getUserName() should be dbInfo.getPassword().

rcordovano: Yes, it would be nice to use the connection pool for this sort of check. But I'm not sure we want to expose connections to the case database in such direct fashion, and since this code is connecting to the postgres database, connection pooling is probably not needed. Perhaps there could be a test method on SleuthkitCase, and we could change the test in the collaboration monitor.

Setting that idea aside for now, I'm not sure how line 524 and 526 are doing almost the same thing. UserPreferences just constructs a CaseDbConnectinoInfo. However, I do see that CaseDbConnectinoInfo has a  settingsValid() method that does the same sort of thing, so calling that would be better than duplicating code, is that what you are suggesting? If so, sounds good to me, but the method name settingsValid() and the docs don't really make it clear that the validation mechanism is making a connection - it could just be checking the form of the settings, for example. We should probably change the name of settingsValid() to something like canConnect(). And CaseDbConnectinoInfo.settingsValid() is also not closing the connection, so we will need to fix that.

As for the copy-paste error, thanks, that slipped by my testing because I was using the default user name and password, which are the same. Good catch!

karlmortensen: Yeah. We probably don't want to expose the connection pool like that. I was
trying to indicate settingsValid(). A different name may help.

In the pull request that is sitting out there for connection pools, it does
close the connection. I added the close() methods, I think everywhere
(ignoring SQLite in imagegallery and timeline, of course).

On Tue, Jun 9, 2015 at 11:59 AM, Richard Cordovano <notifications@github.com

PR #1339, opened 2015-06-11
forced overwrites, updating schema files

karlmortensen:


rcordovano:
Yes, there is a story about caseschema.xsd.

PR #1363, opened 2015-06-15
jump to hex offset implemented

sidheshenator:


kefir-:
Jump to relative (hex and decimal) offsets would also be nice. For example, prepending a plus to an offset could jump that far ahead relative to the cursor. That'd be useful, and wouldn't require any particular GUI development. A prepended minus could jump that far backwards.

sidheshenator:
@kefir- forward(+someHexValue) and backward(-someHexValue) jump relative to the current cursor position implemented.

kefir-:
Nice!! :-) Thanks!

PR #1375, opened 2015-06-18
Inform user if database errors are critical or not

karlmortensen:


karlmortensen:
Going to put this check in Sleuthkit.

PR #1403, opened 2015-06-29
Delete report

sidheshenator:
1. delete report/s feature implemented.
2. originalNode.accept(getActionsDIV)'s null value checked.
3. ReportKML's BufferedReader closed.


Core/src/org/sleuthkit/autopsy/report/ReportKML.java
```diff
@@ -321,6 +321,7 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
                     // read the next line
                     line = reader.readLine();
                 }
```
millmanorama: I would use a try-with block to close the reader.

PR #1408, opened 2015-06-30
Send event when datasource processing is cancelled before ingest begins

APriestman:


APriestman:
Included in other pull request

PR #1413, opened 2015-07-01
make the new tag events work in collaborative autopsy.

millmanorama:
most of this pull request is  Image Gallery Stuff that motivated this change in the first place and is also in develop.

The significant changes are in the non Image Gallery files


millmanorama:
requires https://github.com/sleuthkit/sleuthkit/pull/484 in sleuthkit

millmanorama:
updated to move the new Tag events into org.sleuthkit.autopsy.casemodule.events and make them explicitly implement Serializable.

PR #1422, opened 2015-07-02
Include all drawable mimetypes

millmanorama:


millmanorama:
I am closing this temporarily while I make some more improvements

PR #1434, opened 2015-07-14
unnecessary menu items removed from menu bar

sidheshenator:
Unnecessary Configure Window menu items removed.


rcordovano:
Too many items are removed from the menus by the changes in this pull request. Consult with Brian Carrier and me to determine what can and can't be deleted.

PR #1441, opened 2015-07-17
Current open case not displayed in Open Recent Cases

sidheshenator:


rcordovano:
The code in getRecentCasePaths() is comparing the current case name with case paths. Please correct and resubmit.

PR #1452, opened 2015-07-21
exceptions handled in case usersitepackage paths contain nonascii cha…

sidheshenator:
…racters


kefir-:
Will autopsy find the user's python modules with this change, if the user's home directory contains non-ascii characters?

sidheshenator:
Autopsy can handle non-ascii paths to python modules even without this change.
The problem was with jython-standalone.2.7.0.jar/Lib/site.py throwing exception while searching usersite packages which are optionally used for instantiating the PythonInterpreter. The PR fixes this problem.

PR #1453, opened 2015-07-22
Sqlitemodule

mhmdfy:


rcordovano:
This code was incorporated in another version of this sample module that has been pulled into the baseline, so this particular pull request is OBE.

PR #1478, opened 2015-07-30
Timeline hash set filter

millmanorama:


millmanorama:
going to move the getHashSetNamesForFile method to Content

PR #1495, opened 2015-08-05
Upgrade sqlite jar

mhmdfy:
This is parallel to the pull request on sleuthkit/sleuthkit#493


millmanorama:
if we strip the version off the jar when we add it to AutopsyCore we won't have to change any netbeans configuration if we do another upgrade...

PR #1516, opened 2015-08-20
include tiff, wav along with existing jpeg as exif supported mimetypes

sidheshenator:


Core/src/org/sleuthkit/autopsy/modules/exif/ExifParserFileIngestModule.java
```diff
@@ -65,8 +66,12 @@
     private long jobId;
     private static final IngestModuleReferenceCounter refCounter = new IngestModuleReferenceCounter();
     private FileTypeDetector fileTypeDetector;
+    private final HashSet supportedMimeTypes = new HashSet();
```
millmanorama: sid, can you add the type parameter, so this doesn't generate compiler warnings.

sidheshenator: @millmanorama - Done. :-)

PR #1534, opened 2015-08-26
Lock log before writing

karlmortensen:


rcordovano:
Closing this as we are going to use a different locking technique.

PR #1537, opened 2015-08-27
face detection feature added

sidheshenator:
1. parse makernotes exif tags
2. upgraded metadata-extractor and xmp-core jars


millmanorama:
are the jars things that could be pulled down by ivy?

PR #1547, opened 2015-08-28
save a Set of mimeTypes with different signatures

sidheshenator:


rcordovano:
I am rejecting this pull request for the reasons described below.

FileTypeIdGlobalSettingsPanel.java:

- The replacement of the HashMap with a HashSet is incorrect. The HashMap ensured that there was one file type definition per mime type. The desired change was to allow multiple file type definitions per mime type. The HashSet substitution trivially accomplishes this, but at the cost of injecting bugs. Specifically, because the FileType class does not override Object.equals(), the only effect of the HashSet is ensuring that there is at most one reference to a given FileType object in the collection. At the same time, HashSet makes no guarantees about the iteration  order of the set. Consequences include:
-- saveTypeButtonActionPerformed() is broken. It constructs a new FileType object and stores it into the HashSet. Previously, this implemented a combined new/edit capability (in conjunction with the field clearing New button), because it replaced the FileType for the mime type. Now it just adds another file type every time it is pressed. There is nothing to prevent duplicate file type definitions and the behavior has changed: editing must now be done by a delete and recreate operation.     
-- populateTypeDetailsComponents() is broken. It depended on each file type having a unique mime type. This is no longer true, so the loop over the values in the set matches the first file type object with a particular mime type. It may or may not be the one the user has selected in the list.
- deleteTypeButtonActionPerformed() is broken. 
-- It depended on each file type having a unique mime type. This is no longer true, so the loop over the values in the set matches the first file type object with a particular mime type. It may or may not be the one the user has selected in the list.
-- One cannot remove() from a collection while iterating with a for loop, one must use an Iterator. This code will throw ConcurrentModificationException.

UserDefinedFileTypesManager.java:

- The replacement of the HashMaps with HashSets in this class breaks file type identification.  The original design was for user defined types to overwrite other file type definitions, as indicated in this commented code:
{code}
       /**
         * Load the predefined types first so that they can be overwritten by
         * any user-defined types with the same names.
         */
        loadPredefinedFileTypes();
        loadUserDefinedFileTypes();
{code}
Because of the incorrect use of HashSet with a FileType class that does not override Object.equals(), the overwrite effect is destroyed. Because of the lack of a guaranteed iteration of a HashSet, a user-defined type taking precedence over another type definition is a matter of chance.

PR #1553, opened 2015-09-02
Fix 7zip

mhmdfy:
The inner part will only be called on non-null sized files.


rcordovano:
Closing for additional work.

PR #1565, opened 2015-09-10
NPE fix

mhmdfy:


mhmdfy:
If getName() is null -> error popup and don't display the module
if getName() is empty -> error popup and don't display the module
if getRelativeFilePath() is null -> error popup and don't display the module
if getRelativeFilePath() is empty -> proceed normally, but don't display link

PR #1581, opened 2015-09-18
create a folder for time output instead of leaving them in script folder

mhmdfy:


rcordovano:
Please place the time file in the results folder for the image instead.

PR #1584, opened 2015-09-22
Email parser: NullPointerException patch and embedded messages handling

robertolarcher:
Embedded messages aren't part of the .pst btree structure so some properties are missing (filename, creation time, node id etc.).

When an embedded message is parsed, a NullPointerException is thrown in _PstParser.java_ at line 222 (_attach.getCreationTime()_ returns 0).
This behaviour is related with _getDescriptorNodeId()_ in java-libpst. Java-libpst is already patched in devel branch (https://github.com/rjohnsondev/java-libpst/commit/66ef450fa759f34fa7b4d4ea5555bd9ad22693f4). 

In order to remove this exception in Autopsy, embedded messages are now parsed separately. The extracted artifacts are added to the results (and to the blackboard).

Please note that the embedded message itself is not saved to disk at the moment (a placeholder is saved) but its attachments, if any, are parsed too. Since no _node id_ is available for embedded messages, there is no _Message ID_ in the blackboard and attachments are saved with the same prefix of the parent email message (the container of the embedded message).


mhmdfy:
Roberto, if possible can you provide some data (such as pst files with embedded emails) for testing this feature?

robertolarcher:
Hi. Sorry being late. I've just updated my development environment to Autopsy 4.0.0 after some issues with my PC. Since mailparser is changed since september, I'll try to understand if the problem is still there. If it is, I'll try to set up a sample.

robertolarcher:
Hi. The problem is still present at the moment. Unfortunately I cannot provide a pst file but I have setup a procedure to create a pst file with a sample. These are the steps:

1) Outlook >> new Email
2) embed an image from your HD in the body using the menu: Insert >> Picture
3) send the email to someone and ask to have a replay
4) export the two emails (sent and received) into a pst

Running Autopsy pst ingest module you should have an null exception on the email that you have sent

rcordovano:
Hi Roberto,

We have been hampered in testing and integrating the fix by not having
Outlook here at Basis. Perhaps someone else could volunteer to create the
test data?

Sincerely,
Richard Cordovano
Autopsy Development Team Lead

robertolarcher:
I see. May I send you an email with a .pst test file?

rcordovano:
Yes, that would be great!

rcordovano:
Roberto, we have processed the sample file, but it does not appear to have embedded emails, just two email messages, each with an  attachment (a copy of the same PNG file). Your new code is never executed for this sample.

We also tried using the API described in this StackOverflow article:
http://stackoverflow.com/questions/34723626/extracting-pst-file-containing-mail-as-an-attachment-of-pstmessage
i.e., we tried if (attach.getAttachMethod() == 5) instead of looking for an empty file name. The result was the same. So, unfortunately, we don't have test data for your proposed code change.

Inspecting the code, I have the following concerns:

- extractAttachments is treating the embedded email as **_both_** an email and an attachment and it appears that for each embedded email there will be an 8 byte binary file named uniqueFilenameRoot + "-" + "placeholder for embedded message" that will be added as a derived file and passed through the ingest module pipeline; if there is more than one embedded email, each derived file will overwrite the previous one. I am thinking that when the call to extractEmbeddedEmailMessage in extractAttachments returns, extractEmbeddedEmailMessage should also return.
- In extractAttachments, if the file name is empty and embeddedEmail is null, filename will still be the empty string. Is this possible?
- It appears possible that attach.getAttachMethod() may be correct, rather than the empty file name test.
- extractEmbeddedEmailMessage is a copy-paste of extractEmailMessage with one difference:
            String uniqueFilenameRoot = "" + msg.getDescriptorNodeId();
vs
            String uniqueFilenameRoot = descriptorNodeId;
but the same PSTMessage is being used in every recursive call, so each EmailMessage object will be the same and it looks like there will be infinite recursion given that extractEmbeddedEmailMessage calls extractAttachments with the same PSTMessage that extractEmailMessage uses to call extractAttachments.

Also note that this PR will not auto-merge because of the inclusion of the two auto-updated bundle.properties files

robertolarcher:
Hi Richard,
you are right: the file that I sent was created with the procedure described in my [previous comment]( https://github.com/sleuthkit/autopsy/pull/1584#issuecomment-173841293) and covers only the null pointer exception.

As I wrote may be better to split the fix removing the null pointer exception (easier to test) and then manage the embedded message. If you agree I'll close this pr and open another one only for the null exception. Unfortunately I have little time to test and code.

rcordovano:
Hi Roberto, I agree with you; in fact, I already had an in-house developer
bring in your two bug fixes (I also had him fix a bug that I found while
inspecting this code, an area of the code base I had not previously
explored!). If you or someone else in the community has time to get us some
test data with embedded emails, we can take over coding up the solution. It
is however, too late to get this in Autopsy 4.0.1.

Thank you very much for your contribution!

rcordovano:
Closing this pull request because we have been unable to verify the functionality and it no longer merges. The source repository is unknown, so the branch cannot be copied to this repository. We do have a branch in which the code was hand merged into a branch of develop. I am pushing that branch (embedded_email_per_larcher) to this repository for later reference.

PR #1600, opened 2015-10-01
added feature to hide individual clusters/descriptions

millmanorama:


millmanorama:
@rcordovano I fixed up the suite.properties, so this should be ready to go.

rcordovano:
Cool, thanks!

PR #1603, opened 2015-10-02
change _time.txt directory

mhmdfy:


mhmdfy:
It is still unclear to me why we create them and send them to the input directory in the first place

PR #1608, opened 2015-10-08
Run ingest module on selected part of filesystem

robertolarcher:
As requested by rpav [here](http://forum.sleuthkit.org/viewtopic.php?f=7&t=2574) this pull request simply enables the context menu on directories.

Limitations: not all ingest modules work correctly on a (sub) directory. _Recent Activity_ and _Android Analyzer_ raise errors.

_Keyword Search_ module has been tested extensively and no problem has been reported.


bcarrier:
Thanks. The IngestManager code assumes that a data source is passed into (even though, as you found out, you can also pass in a Directory).   The modules you mentioned also assume they will get passed in a data source.  The fix to this is either:
1) We change the logic in the dialog to not display the modules that will only work on a full data source.
2) We change the logic in INgestManager to not run those modules (but this is a bit confusing because the user will have thought that they would be running. 

1 seems like the better approach.

robertolarcher:
Hi, sorry for replying late. I think that the better approach would be 1 & 2 :)
I mean: 1) seems the better but I would also prevent IngestManager to run ingest modules that are not suitable for directories.
As far as I am concerned, this feature is really important because we may have multiple accounts on a PC (think about normal users and administrator(s)) and only one user is under investigation. Since the ingest modules correctly output an error when they fail, I think that even without option 1 and/or 2, this feature should be enabled.

rcordovano:
The complete solution has been integrated via PR #1710, so closing this. Thanks!

PR #1633, opened 2015-10-19
fix stack overflow when zooming way out by disabling zooming out furt…

millmanorama:
…her than the entire case...  other minor cleanup


rcordovano:
This does not auto-merge. Please address and resubmit.

PR #1635, opened 2015-10-19
Multi sig mime type

mhmdfy:


Core/src/org/sleuthkit/autopsy/modules/filetypeid/FileType.java
```diff
@@ -102,6 +103,31 @@ boolean alertOnMatch() {
...
     }
+    
+    @Override
+    public String toString() {
+        return this.mimeType;
+    }
+    
+    @Override
+    public boolean equals(Object other) {
+        if(other != null && other instanceof FileType) {
```
millmanorama: I think this equals/hashcode pair is broken.  Two filetypes that are equal based on their comparison of mimetype and signature could have different hashcodes becuase the hash code also includes the interesting file set name and alert boolean.

Core/src/org/sleuthkit/autopsy/modules/filetypeid/FileType.java
```diff
@@ -193,6 +219,25 @@ boolean containedIn(final AbstractFile file) {
                 return false;
             }
         }
+        
+        @Override
+        public boolean equals(Object other) {
+            if (other != null && other instanceof Signature) {
+                Signature that = (Signature) other;
+                if(Arrays.equals(this.getSignatureBytes(), that.getSignatureBytes()) && this.getOffset() == that.getOffset())
```
millmanorama: same issue with equals and hashCode here, I think

Core/src/org/sleuthkit/autopsy/modules/filetypeid/UserDefinedFileTypesManager.java
```diff
@@ -231,17 +231,47 @@ private void loadUserDefinedFileTypes() throws UserDefinedFileTypesException {
...
+
+    /**
+     * Adds given FileType to given map:
+     * if the mimetype exists, add fileType to that mimetype's list.
+     * otherwise, create a new mimetype with the new fileType.
+     * @param map The map to be modified.
+     * @param fileType The added FileType
+     */
+    void addFileTypeToMap(Map<String, List<FileType>> map, FileType fileType) {
+        String mimeType = fileType.getMimeType();
```
millmanorama: guava provides a great MultiMap (see https://code.google.com/p/guava-libraries/wiki/NewCollectionTypesExplained) that is quite usefull in this kind of situation.

PR #1641, opened 2015-10-21
refactor Timeline initialization code to align with ImageGallery and …

millmanorama:
…use more constructor style dependency injection.


rcordovano:
Does not auto-merge. Please address and re-submit.

PR #1685, opened 2015-10-30
Extend kws core check to multi user

rcordovano:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/KeywordSearch.java
```diff
@@ -126,7 +126,7 @@ public void propertyChange(PropertyChangeEvent evt) {
                     } catch (Exception ex) {
                         String caseName = closedCase.getName();
                         logger.log(Level.SEVERE, String.format("Failed to close core for %s", caseName), ex);
-                        MessageNotifyUtil.Notify.error(NbBundle.getMessage(KeywordSearch.class, "KeywordSearch.closeCore.notification..msg"), ex.getCause().getMessage());
+                        MessageNotifyUtil.Notify.error(NbBundle.getMessage(KeywordSearch.class, "KeywordSearch.closeCore.notification.msg"), ex.getMessage());
```
karlmortensen: Are you sure we don't want getCause() here?

rcordovano: There is not always a cause, which leads to an NPE. I decided to standardize the scheme rather than have a branch on whether or not getCause() returned null. Especially given that we decided to hide the raw exceptions with our own class. The log should show the gory details.

rcordovano: To clarify, what I mean by standardize the scheme is that the methods called directly or indirectly by KeywordSearch$CaseChangeListener.propertyChange all construct their KeywordSearchModuleException with a user-friendly message, which is shown as the details of the notification. If there was a lower level exception, it is set as the cause of the KeywordSearchModuleException . This way, the logging calls in propertyChange will show the cause in the stack traces. The motivation for this was:
- Making the lower level methods throw to the top-level which then consistently handles both logging and notification;
- Trying for user-friendly messages in the details section instead of raw Solr exception messages, etc.;
- The catch blocks in propertyChange are firewalls, i.e., catching Exception, another reason to not assume that there would be a non-null cause object

PR #1747, opened 2015-12-01
Tl busy indicators and background thread optimization

millmanorama:


rcordovano:
This does not auto-merge. Please resolve and resubmit, thanks!

PR #1782, opened 2015-12-18
if no gold dumps are provided, generete them

mhmdfy:


mhmdfy:
If they are NOT provided*

PR #1823, opened 2016-01-19
Negative array size exception

robertolarcher:
Same problem as pr #1556 (some other pr reintroduced the exception?)
Slight different patch

This excepion is raised when the attachment size is a multiple of 8176
(the buffer size).

InputStream read methos return -1 if there is no more data because the
end of the stream has been reached

A simple check if result != -1 is added

refs:
http://forum.sleuthkit.org/viewtopic.php?f=6&t=2563
http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read()

### DIFF
--- a/branding/core/core.jar/org/netbeans/core/startup/Bundle.properties
+++ b/branding/core/core.jar/org/netbeans/core/startup/Bundle.properties
@@ -1,5 +1,5 @@
 #Updated by build script
-#Mon, 21 Dec 2015 06:26:49 -0500
+#Tue, 19 Jan 2016 12:12:04 +0100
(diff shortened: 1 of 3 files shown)


rcordovano:
Roberto, this is ready for integration, but it will not auto-merge, i.e., there are conflicts to resolve. I am pretty certain the problem is simply the date changes in the two bundle.properties files. It is annoying that NetBeans updates the dates on these configuration-controlled files; I have a git alias for:

git checkout -- branding 

that I use to before every commit out of habit now.

Please address the merge conflict.

rcordovano:
The fixes in this pull request have been added in PR #1900.

PR #1835, opened 2016-01-26
IG split nav panel

millmanorama:
after pr # 1833

refactor the nav area into seperate Tree and hash hit list areas.


rcordovano:
Does not auto-merge, please fix and resubmit.

PR #1851, opened 2016-01-29
1501 ja imagegallery

mrtizmo:
I had to re-work parts of ImageGalleryTopComponent.java. Look closely to make sure everything is to your satisfaction.

This file is one of a small number of cases where the @Messages{} was used by the developer instead of the Bundle.properties files. I most often see this happen with TopComponent setups, so it may be something that NetBeans is auto-generating.


millmanorama:
Most of the @Messages annotation that you removed were indeed auto-generated bu NetBeans, but at least one was not.  I have started to use them(you will notice them more in timeline, especially), for two reasons. 1) it keeps the strings close to where they are used in the source code.  i find this easier for me to track down when i need to find where a string in the UI comes from. 2)  It auto-generates a Bundle.java class with getters for all the properties.  I find it easier to avoid mistakes when using these than trying to match the bundle keys.  My understanding was that this approach was compatible with localized properties and resulted in a distributed Bunde.properties file that was the composition of the one in the source tree and the properties from the annotations.  

Please let me and @rcordovano know if I need to avoid using the @messages annotations.

Otherwise this looks fine.

PR #1876, opened 2016-02-01
1501 ja image gallery - redo 2

mrtizmo:
Fixed bundle tags. Removed one empty string tag. Fixed duplicate tag issue. 

This fix by itself will not auto-merge because you reverted all of the changes from "redo 1" that this code depends on. However, if you can re-apply the changes from "redo 1" and then apply these changes on top of that, it should all be ready to go.


rcordovano:
After reverting the revert of "redo 1" and merging "red 2" the image gallery package does not build. I need to revert again, sorry.

PR #1881, opened 2016-02-01
Bb user defs in reports

BasisOlivers:


Core/src/org/sleuthkit/autopsy/report/ArtifactSelectionDialog.java
```diff
@@ -66,21 +66,25 @@ public ArtifactSelectionDialog(java.awt.Frame parent, boolean modal) {
...
+                    BlackboardArtifact.ARTIFACT_TYPE.TSK_GEN_INFO.getLabel(),
+                    BlackboardArtifact.ARTIFACT_TYPE.TSK_GEN_INFO.getDisplayName()));
+            doNotReport.add(new BlackboardArtifact.Type(BlackboardArtifact.ARTIFACT_TYPE.TSK_TOOL_OUTPUT.getTypeID(),
+                    BlackboardArtifact.ARTIFACT_TYPE.TSK_TOOL_OUTPUT.getLabel(),
+                    BlackboardArtifact.ARTIFACT_TYPE.TSK_TOOL_OUTPUT.getDisplayName())); // output is too unstructured for table review
+
+            artifacts = Case.getCurrentCase().getSleuthkitCase().getArtifactTypesInUse();
             artifacts.removeAll(doNotReport);
-            Collections.sort(artifacts, new Comparator<BlackboardArtifact.ARTIFACT_TYPE>() {
+            Collections.sort(artifacts, new Comparator<BlackboardArtifact.Type>() {
```
millmanorama: I don't think you need to specify the comparator here, since Artifact.Type is Comparable.

BasisOlivers: Only attributes are comparable at the moment, though that may be something we want to switch.

millmanorama: ah, ok, I assumed artifact types were too. Then I would use Comparator.comparing(BlackboardArtifact.Type::getDisplayName) here, but that is mostly stylistic.

PR #1900, opened 2016-02-02
Fixed lack of unique file name

BasisOlivers:


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/PstParser.java
```diff
@@ -215,14 +217,14 @@ private void extractAttachments(EmailMessage email, PSTMessage msg) {
...
-                String uniqueFilename = msg.getDescriptorNodeId() + "-" + filename;
+                String uniqueFilename = fileID + "-" + msg.getDescriptorNodeId() + "-" + filename;
                 String outPath = outputDirPath + uniqueFilename;
                 saveAttachmentToDisk(attach, outPath);
 
                 EmailMessage.Attachment attachment = new EmailMessage.Attachment();
 
-                long crTime = attach.getCreationTime().getTime() / 1000;
-                long mTime = attach.getModificationTime().getTime() / 1000;
+                long crTime = attach.getCreationTime() != null ? attach.getCreationTime().getTime() : 0;
```
karlmortensen: It looks like we just went from crTime being in seconds to crTime being in milliseconds. Was that deliberate?

BasisOlivers: No, that was a mistake. Thank you for the catch.

PR #1946, opened 2016-02-04
fix NullPointerException due to null getFont().

mrtizmo:
The Exception seems to indicate that .getFont() was returning NULL. This change makes the jSplitPane1 get the font from leftPane, which will have a valid font.

### DIFF
--- a/KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/DropdownListSearchPanel.java
+++ b/KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/DropdownListSearchPanel.java
@@ -200,7 +200,7 @@ class DropdownListSearchPanel extends KeywordSearchPanel {
 
         setFont(getFont().deriveFont(getFont().getStyle() & ~java.awt.Font.BOLD, 11));
 
-        jSplitPane1.setFont(jSplitPane1.getFont().deriveFont(jSplitPane1.getFont().getStyle() & ~java.awt.Font.BOLD, 11));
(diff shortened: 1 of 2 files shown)


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/DropdownListSearchPanel.java
```diff
@@ -200,7 +200,7 @@ private void initComponents() {
 
         setFont(getFont().deriveFont(getFont().getStyle() & ~java.awt.Font.BOLD, 11));
 
-        jSplitPane1.setFont(jSplitPane1.getFont().deriveFont(jSplitPane1.getFont().getStyle() & ~java.awt.Font.BOLD, 11));
+        jSplitPane1.setFont(leftPane.getFont());
```
millmanorama: this looks wrong to me.  the font for leftpane is not assigned until the next couple of lines

mrtizmo:
On line ~193 the leftPane variable is assigned a new JScrollPane:

`leftPane = new javax.swing.JScrollPane();`

As of that point, the leftPane has a default font that can be accessed using getFont().

If you look at line ~205, where the font of the leftPane is modified, you'll see that we are using the existing font of the leftPane object as the basis to change the font. This is exactly the same thing that is happening when we use the leftPane as the argument to setFont() on the jSplitPane1, except in one case we using the default font of the leftPane as-is, and in the other case, we are taking the default font and applying modifications to it.

Does that clarify what is going on?

millmanorama:
makes perfect sense when you explain it!

PR #1951, opened 2016-02-05
mark as deprecated

karlmortensen:


rcordovano:
Deprecating this method was a mistake. The deprecation will be removed in my next commit and the warning will go away.

PR #1995, opened 2016-03-01
Updated for new constructor for attribute type

BasisOlivers:


millmanorama:
@rcordovano , @BasisOlivers  why not do the same thing for BlackboardArtifact.Type?

PR #2012, opened 2016-03-09
Disabled cancel button in reports panel on completion

BasisOlivers:


Core/src/org/sleuthkit/autopsy/report/ReportProgressPanel.java
```diff
@@ -44,6 +47,12 @@
         ERROR
     }
 
+    public enum Events {
+
+        COMPLETED,
+        NOT_COMPLETED
```
millmanorama: NOT_COMPLETED is kind of a weird event name...  what about STARTED> or IN_PROGRESS?

millmanorama: Or is it supposed to be an error?

BasisOlivers: Only completed is used, so I'm just going to get rid of it. Don't know why I have it in the first place.

PR #2031, opened 2016-03-21
refactor FileTypeDetector so that detect() and getFileType() have con…

millmanorama:
…sistent behavior except for adding the results to the case db.


millmanorama:
i made those two changes you wanted @rcordovano

PR #2032, opened 2016-03-21
Made cancel button disable when pressed in reports

BasisOlivers:


rcordovano:
I added the relevant code from this pull request in two commits to get around merge conflicts.

PR #2037, opened 2016-03-23
Mime type in file search by attributes

BasisOlivers:


Core/src/org/sleuthkit/autopsy/filesearch/FilterArea.java
```diff
@@ -92,20 +92,22 @@ private void init() {
...
+        BoxLayout filtersPanelLayout = new BoxLayout(this, BoxLayout.Y_AXIS);
+        this.setLayout(filtersPanelLayout);
 
-        for (FileSearchFilter f : filters) {
+        for (int i = 0; i < filters.size(); i++) {
+            FileSearchFilter f = filters.get(i);
             JComponent filterComponent = f.getComponent();
             filterComponent.setAlignmentX(Component.LEFT_ALIGNMENT);
-            filterComponent.setBorder(new EmptyBorder(0, 0, 20, 0));
-            filtersPanel.add(filterComponent);
```
millmanorama: this looks like a hack,  if the components need differnet borders (based on their filter type) they should set them themselves

BasisOlivers: It's not by what kind of component it is, it's whether it's the last component or not. This is to make it clear that some parts are grouped together.

PR #2054, opened 2016-03-31
Fixed log to show complete stack trace.

BasisOlivers:


Core/src/org/sleuthkit/autopsy/coreutils/Logger.java
```diff
@@ -90,17 +90,21 @@ private static FileHandler createFileHandlerWithTraces(String logDirectory) {
             fileHandler.setFormatter(new Formatter() {
                 @Override
                 public String format(LogRecord record) {
-                    if (record.getThrown() != null) {
+                    Throwable thrown = record.getThrown();
+                    if (thrown != null) {
```
millmanorama: With the while loop, I think the entire if statement is unneeded.

BasisOlivers: Doing so creates extra whitespace (because there is a \n after stackTrace), which is undesirable.

millmanorama: why do we need to add the final newline?

BasisOlivers: Unsure, will consult Richard on that

rcordovano: I don't know if the newline is required. As long as I get my full stack
traces with no white space between log entries, I'm a happy camper!

PR #2085, opened 2016-04-20
Fixed global settings panels' ui

BasisOlivers:


rcordovano:
Still needs some refinement, please resubmit.

PR #2087, opened 2016-04-21
include datasource name in prompt to rebuild TL db when ingest done

millmanorama:


rcordovano:
Does not auto-merge, please address and re-submit.

PR #2100, opened 2016-04-29
TL only snapshot visualization

millmanorama:
take after https://github.com/sleuthkit/autopsy/pull/2097

rearranges scene-graph so that the SaveSnapshotAsReport action can clip the snapshot to the bounds of the visualization (including axes) and exclude the surrounding UI.


rcordovano:
Please add method documentation - at least for the new methods.

PR #2106, opened 2016-05-03
Cleaned up reports code to only allow one report module at a time.

BasisOlivers:


rcordovano:
Please do additional refinement of this solution per our discussion.

PR #2107, opened 2016-05-03
TL Fix back action; cleanup pushInterval

millmanorama:
clean up plus fixing back action that was broken by refactoring in ZoomSettingsPane


millmanorama:
I am going to seperate the actual fix from the cleanup

PR #2124, opened 2016-05-11
Fixed expansion behavior on options panel.

BasisOlivers:


rcordovano:
Looks good, but does not auto-merge. Please address and resubmit.

PR #2128, opened 2016-05-11
Added mime type to file metadata content viewer.

BasisOlivers:


rcordovano:
Looks good, but need to handle error label reset as discussed. Please address and re-submit.

PR #2134, opened 2016-05-12
Made stix report properly dispose upon failure, add to db.

BasisOlivers:


rcordovano:
Does not auto-merge. Please address and resubmit.

PR #2159, opened 2016-05-23
TL change visualizaion -> view in code, ui, and comments

millmanorama:


rcordovano:
Does not auto-merge, please update and re-submit.

PR #2160, opened 2016-05-24
783 remove zoom in list mode

millmanorama:


rcordovano:
The zoom controls are still there. In testing, I did a full clean and rebuild. Please re-submit.

PR #2162, opened 2016-05-24
788 lazy context menus

millmanorama:
make context menus on demand rather than upfront


rcordovano:
Does not auto-merge. Please address and re-submit.

PR #2163, opened 2016-05-24
787 reset selection on mode change

millmanorama:


rcordovano:

PR #2167, opened 2016-05-25
793 collapse file system events in list view

millmanorama:
collapse file system events in list view
adds multi select support as necessary consequence of collapsing events,


rcordovano:

PR #2168, opened 2016-05-25
792 truncate descriptions in list view

millmanorama:


rcordovano:

PR #2171, opened 2016-05-25
Changed default path for import hash database dialog

eugene7646:


Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbImportDatabaseDialog.java
```diff
@@ -251,6 +253,8 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
     private void openButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_openButtonActionPerformed
         if (ModuleSettings.settingExists(ModuleSettings.MAIN_SETTINGS, LAST_FILE_PATH_KEY)) {
             fileChooser.setCurrentDirectory(new File(ModuleSettings.getConfigSetting(ModuleSettings.MAIN_SETTINGS, LAST_FILE_PATH_KEY)));
+        } else {
+            fileChooser.setCurrentDirectory(new File(Paths.get(PlatformUtil.getUserConfigDirectory(), "HashDatabases").toString()));
```
rcordovano: This does not work on the first pass through after a clean and build, i.e., with a new userdir. It appears that the default directory is used, although the documentation for JFileChooser suggests otherwise. I think we should create the default directory if it does not exist.

PR #2189, opened 2016-05-27
Added timeline to the toolbar.

BasisOlivers:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1523,7 +1523,7 @@ private static void doCaseChange(Case toChangeTo) {
                 CallableSystemAction.get(CaseCloseAction.class).setEnabled(true);
                 CallableSystemAction.get(CasePropertiesAction.class).setEnabled(true);
                 CallableSystemAction.get(CaseDeleteAction.class).setEnabled(true); // Delete Case menu
-
+                CallableSystemAction.get(OpenTimelineAction.class).setEnabled(true);
                 if (toChangeTo.hasData()) {
```
rcordovano: Looking at this, it seems strange that the enabling of the actions is not being dispatched to the EDT queue with SwingUtilities.invokeLater. Please do so, as for the case closed branch of the conditional.

Core/src/org/sleuthkit/autopsy/core/layer.xml
```diff
@@ -387,11 +387,11 @@
             <file name="org-sleuthkit-autopsy-report-ReportWizardAction.shadow">
                 <attr name="displayName" bundlevalue="org.sleuthkit.autopsy.report.Bundle#Toolbars/Reports/org-sleuthkit-autopsy-report-ReportWizardAction.shadow"/>
                 <attr name="originalFile" stringvalue="Actions/Tools/org-sleuthkit-autopsy-report-ReportWizardAction.instance"/>
-                <attr name="position" intvalue="102"/>
+                <attr name="position" intvalue="104"/>
             </file>
         </folder>
         <folder name="Ingest">
-            <attr intvalue="103" name="position"/>
+            <attr intvalue="104" name="position"/>
```
rcordovano: It looks like two items are being given the same position here, which will probably cause a new warning to be emitted.

rcordovano:
Is the intent to rename the Timeline icon? It looks like it is just added again with a new name, leaving the file with the old name in place. Also, there are a couple of file changes related to the deferred image gallery work that should not be included.

rcordovano:
In tests, the Timeline toolbar button is truncated and does not have an icon. Please address these issues and re-submit.

PR #2212, opened 2016-06-06
Handle JFIF .tec files as JPEG with a leading EOI marker

karlmortensen:


Core/src/org/sleuthkit/autopsy/coreutils/ImageUtils.java
```diff
@@ -726,8 +753,20 @@ private void saveThumbnail(BufferedImage thumbnail) {
                 if (image.isError() == false) {
                     return image;
                 }
-                //fall through to default image reading code if there was an error
+            } else if (file.getNameExtension().toLowerCase().compareTo("tec") == 0) {
```
millmanorama: why not just :
  } else if (file.getNameExtension().equalsIgnoreCase("tec") ) {

karlmortensen: Awesome. Will do.

karlmortensen:
Submitting to develop instead.

Core/src/org/sleuthkit/autopsy/coreutils/ImageUtils.java
```diff
@@ -726,8 +753,20 @@ private void saveThumbnail(BufferedImage thumbnail) {
...
+                int offset = 0;
+                if (isJfifFileHeaderWithLeadingEOIMarker(file)) {
```
millmanorama: I think using the Image constructor directly means we are missing out of the progress indicator.  I wonder if there is a way redesign things for gif and tec so that we still get it?

millmanorama:
have you tested this?  You might also want to file an issue with TwelveMonkeys at https://github.com/haraldk/TwelveMonkeys/issues if this is something they could be supporting.  Given how easy it was for you to work around, I suspect they could and would like to.

karlmortensen:
Yes, it has been tested. I wrote an issue at TwelveMonkeys. We'll see what happens.

PR #2214, opened 2016-06-06
Support .tec JFIF files with leading EOI marker

karlmortensen:


rcordovano:
Does not auto-merge, please address and resubmit. Also please advise regarding test input via private email, thanks.

PR #2219, opened 2016-06-08
Made run ingest modules dialog accessible via directory listing.

BasisOlivers:


Core/src/org/sleuthkit/autopsy/datamodel/VirtualDirectoryNode.java
```diff
@@ -80,13 +87,28 @@ public VirtualDirectoryNode(VirtualDirectory ld) {
...
         List<Action> actions = new ArrayList<>();
+
         actions.add(new NewWindowViewAction(
                 NbBundle.getMessage(this.getClass(), "VirtualDirectoryNode.getActions.viewInNewWin.text"), this));
         actions.add(null); // creates a menu separator
         actions.add(ExtractAction.getInstance());
         actions.add(null); // creates a menu separator
+        
+        if (this.content.isDataSource()) {
+            actions.add(new FileSearchAction(
```
rcordovano: I have verified that we want the ability to run ingest modules on non-data source virtual directories. Please remove the isDataSource check.

PR #2225, opened 2016-06-09
Show meta information for interesting file hits in HTML report

karlmortensen:


Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
```diff
@@ -899,23 +899,32 @@ public int compareTo(ArtifactData otherArtifactData) {
...
-                            pathToShow = getFileUniquePath(content);
-                        }
-                        attributeDataArray[2] = pathToShow;
                     }
                 }
+
+                attributeDataArray[2] = content.getUniquePath();
+                if (attributeDataArray[2].isEmpty()) {
+                    attributeDataArray[2] = getFileUniquePath(content);
+                }
```
rcordovano: getFileUniquePath calls Content.getUniquePath, which should always return something. I believe lines 912-914 can be removed.

Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
```diff
@@ -899,23 +899,32 @@ public int compareTo(ArtifactData otherArtifactData) {
...
+                try {
+                    List<ContentTag> contentTags = Case.getCurrentCase().getServices().getTagsManager().getContentTagsByContent(content);
+                    for (ContentTag ct : contentTags) {
+                        allTags.add(ct.getName().getDisplayName());
+                    }
+                } catch (TskCoreException ex) {
+                    errorList.add(NbBundle.getMessage(this.getClass(), "ReportGenerator.errList.failedGetContentTags"));
+                    logger.log(Level.SEVERE, "Failed to get content tags", ex); //NON-NLS
+                }
+                orderedRowData.add(makeCommaSeparatedList(allTags));
```
rcordovano: I wonder if it would be useful to distinguish content tags and artifact tags with two separate columns? Thoughts?

PR #2228, opened 2016-06-10
Made stix report properly dispose upon failure, add to db.

BasisOlivers:


Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -108,34 +109,20 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
         reportAllResults = configPanel.getShowAllResults();
 
         // Set up the output file
```
rcordovano: This comment about setting up the out[put file should be deleted.

Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -108,34 +109,20 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
...
-                    NbBundle.getMessage(this.getClass(), "STIXReportModule.progress.completedWithErrors"));
-            return;
-        }
-
         // Keep track of whether any errors occur during processing
         boolean hadErrors = false;
 
         // Process the file/directory name entry
         String stixFileName = configPanel.getStixFile();
+        File reportFile = new File(reportPath);
```
rcordovano: It would be better for "locality of reference" to move the creation of reportFile closer to where it is used.

Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -157,61 +145,76 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
...
-            idToResult = new HashMap<String, ObservableResult>();
-        }
-
-        // Close the output file
-        if (output != null) {
-            try {
-                output.close();
-            } catch (IOException ex) {
-                logger.log(Level.SEVERE, String.format("Error closing STIX report file %s", reportPath), ex); //NON-NLS
+            // Close the output file
```
rcordovano: This should be omitted. The try with resources will auto-close the output file.

Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -157,61 +145,76 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
...
+            Case.getCurrentCase().addReport(reportPath, Bundle.STIXReportModule_srcModuleName_text(), "");
+            if (hadErrors) {
+                progressPanel.complete(ReportStatus.ERROR);
+                progressPanel.updateStatusLabel(
+                        NbBundle.getMessage(this.getClass(), "STIXReportModule.progress.completedWithErrors"));
+            } else {
+                progressPanel.complete(ReportStatus.COMPLETE);
+            }
+        } catch (IOException ex) {
+            logger.log(Level.SEVERE, String.format("Unable to open STIX report file %s", reportPath), ex); //NON-NLS
```
rcordovano: With the IOException catch when the explicit close is removed, this will become a more general error, please make the log message more generic.

PR #2233, opened 2016-06-14
Keyword search cancellation policy update.

BasisOlivers:


Core/src/org/sleuthkit/autopsy/modules/photoreccarver/PhotoRecCarverFileIngestModule.java
```diff
@@ -256,6 +258,12 @@ public void startUp(IngestJobContext context) throws IngestModule.IngestModuleEx
             java.io.File newAuditFile = new java.io.File(Paths.get(outputDirPath.toString(), PHOTOREC_REPORT).toString()); //NON-NLS
```
rcordovano: Please insert as cancellation check after writing the file to disk but before executing PhotoRec.

Core/src/org/sleuthkit/autopsy/modules/photoreccarver/PhotoRecCarverOutputParser.java
```diff
@@ -154,4 +154,4 @@
         List<LayoutFile> empty = Collections.emptyList();
```
rcordovano: We need a cancellation check in PhotoRecCarverOutputParser.parse before processing each file object element.

Core/src/org/sleuthkit/autopsy/modules/photoreccarver/PhotoRecCarverOutputParser.java
```diff
@@ -154,4 +154,4 @@
...
         return empty;
     }
```
rcordovano: We need a cancellation check in PhotoRecCarverOutputParser.parse, before processing each file object element.

PR #2272, opened 2016-06-30
Fixed double open on double click for reports.

BasisOlivers:


rcordovano:
This is the right solution, but it does not go far enough - the MouseListener needs to be removed altogether. Closing this PR and implementing that solution.

PR #2277, opened 2016-07-14
Credit card account search

millmanorama:
filter credit card numbers through the luhn check and parse track 1 and 2 data.  Still posts them as keyword hit artifacts.


millmanorama:
the luhn check in this is broken

PR #2292, opened 2016-08-04
Added three-character suffix for Java in build.xml

hoyt-harness:
Latest java-8-oracle for Ubuntu is 1.8.0_101.

build.xml only has java-1.8.0 tests for versions with up to two-character suffixes - i.e. 1.8.0_99. Anything higher causes build to fail.

This change allows for up to three-character suffixes, i.e. 1.8.0_999.

Changes to Bundle.properties are a result of the build process and shouldn't have been included in this commit. I would remove them, but my Git kung fu isn't that strong.

### DIFF
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -1,6 +1,6 @@
-                  Last Updated: 04 August 2016
+                  Last Updated: 30 October 2015
 
-This file outlines what it takes to build Autopsy 4.1.0 from source.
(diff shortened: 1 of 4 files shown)


hoyt-harness:
This change removes my edits to BUILDING.txt. Edits that may need to be made can be done at a later time when UNIX installs are working. The edits to build.xml are still included.

PR #2314, opened 2016-09-01
AUT-2213 Added cancellation checks for ingest modules using ContentUtils.writeToFile

wishdasher:


Core/src/org/sleuthkit/autopsy/ingest/DataSourceIngestCancellationCheck.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
millmanorama: you need to fix this license header

Core/src/org/sleuthkit/autopsy/ingest/FileIngestCancellationCheck.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
```
millmanorama: fix this license header

millmanorama:
I have two overall comments, neither of which should hold up merging this PR:
1) there is  a lot of duplicated code across the different kind of analyzers in the Android module.  This is beyond the scope of this PR or the story, but it feels like we could do something to have a better abstraction  and reduce the duplicate code.
2) If we wanted to really be adventurous, I think we could do away with the two concrete classes FileIngestCancellationCheck and DataSourceCancellationCheck and just use method references to context::fileIngestIsCancelled or context::dataSourceIngestIsCancelled

PR #2315, opened 2016-09-02
AUT-2107 tag name deletion options panel

wishdasher:


rcordovano:
Please make the modifications we discussed and re-submit, thanks!

PR #2322, opened 2016-09-08
AUT-2105 keyboard shortcut for tagging content file

wishdasher:
Current shortcut is CTRL+b, this could be changed
This depends on a method added to TagsManager in AUT-2107 (getPredefinedTagNames)

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/actions/BookmarkFileAction.java
+++ b/Core/src/org/sleuthkit/autopsy/actions/BookmarkFileAction.java
@@ -19,10 +19,6 @@ import org.sleuthkit.datamodel.AbstractFile;
 import org.sleuthkit.datamodel.TagName;
 import org.sleuthkit.datamodel.TskCoreException;
 
-/**
- *
- * @author smori
- */
 public class BookmarkFileAction extends AbstractAction {
      
     private static final String NO_COMMENT = "";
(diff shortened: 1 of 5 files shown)


Core/src/org/sleuthkit/autopsy/corecomponents/DataResultTopComponent.java
```diff
@@ -18,18 +18,23 @@
...
+import javax.swing.JComponent;
+import javax.swing.KeyStroke;
 import org.openide.explorer.ExplorerManager;
 import org.openide.explorer.ExplorerUtils;
 import org.openide.util.NbBundle;
 import org.openide.windows.TopComponent;
 import org.openide.nodes.Node;
 import org.openide.windows.Mode;
 import org.openide.windows.WindowManager;
+import org.sleuthkit.autopsy.actions.BookmarkFileAction;
```
rcordovano: This does not compile. BookmarkFileAction is not public. If this is the only place the class is needed, we would leave it as package access and put it in this package rather than add a new public class.

Core/src/org/sleuthkit/autopsy/actions/BookmarkFileAction.java
```diff
@@ -0,0 +1,44 @@
...
+import org.sleuthkit.datamodel.TagName;
+
+class BookmarkFileAction extends AbstractAction {
+
+    private static final String NO_COMMENT = "";
+    private static final String BOOKMARK = NbBundle.getMessage(BookmarkFileAction.class, "BookmarkFileAction.bookmark.text");
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        List<TagName> tagNames = Case.getCurrentCase().getServices().getTagsManager().getPredefinedTagNames();
```
rcordovano: This method does not exist in this branch. Also, with the TagsManager code we have now, there is no guarantee that the Bookmark tag name has been added to the database.

Core/src/org/sleuthkit/autopsy/corecomponents/DataResultTopComponent.java
```diff
@@ -104,6 +109,8 @@ private void customizeComponent(boolean isMain, String title) {
 
         setTitle(title); // set the title
         setName(title);
+        getInputMap(JComponent.WHEN_ANCESTOR_OF_FOCUSED_COMPONENT).put(KeyStroke.getKeyStroke(KeyEvent.VK_B, InputEvent.CTRL_MASK), "addBookmarkTag");
```
rcordovano: The user has no way to discover that this short cut exists. It should be added to the Tag File menu item, e.g., "Tag File (Ctrl + B)."

Also, this implementation allows you to tag things that do not have the "Tag File" context menu action and should not be able to be tagged. For example, the $Unalloc and $CarvedFiles virtual directories.

rcordovano:
This does not auto-merge and has some issues to be resolved (see review comments). Please update and re-submit.

PR #2347, opened 2016-09-16
1985: update ingest module to use cancellable content utils

zhhl:


rcordovano:
Please make the small changes we discussed and resubmit.

PR #2353, opened 2016-09-21
844 generic account artifact

millmanorama:


Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -764,14 +786,17 @@ public long getHits() {
...
                     + "      GROUP_CONCAT(blackboard_artifacts.review_status_id) AS review_status_ids "
                     + " FROM blackboard_artifacts " //NON-NLS
-                    + " LEFT JOIN blackboard_attributes ON blackboard_artifacts.artifact_id = blackboard_attributes.artifact_id " //NON-NLS
-                    + "                                AND blackboard_attributes.attribute_type_id = " + BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SOLR_DOCUMENT_ID.getTypeID() //NON-NLS
-                    + " WHERE blackboard_artifacts.artifact_type_id = " + BlackboardArtifact.ARTIFACT_TYPE.TSK_CREDIT_CARD_ACCOUNT.getTypeID() //NON-NLS
+                    + " LEFT JOIN blackboard_attributes as solr_attribute ON blackboard_artifacts.artifact_id = solr_attribute.artifact_id " //NON-NLS
+                    + "                                AND solr_attribute.attribute_type_id = " + BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SOLR_DOCUMENT_ID.getTypeID() //NON-NLS
+                    + " LEFT JOIN blackboard_attributes as account_type ON blackboard_artifacts.artifact_id = account_type.artifact_id " //NON-NLS
```
rcordovano: Since there is some chance that we will be using Elastic instead of Solr in the future, I would prefer something like TSK_EXTRACTED_TXT_DOC_ID or TSK_KEYWORD_SEARCH_DOC_ID to TSK_SOLR_DOC_ID.

I am also wondering about TSK_PAYMENT_CARD_TYPE. There are several new TSK_CREDIT_CARD_ attributes, along with some more generic new attributes such as TSK_BRAND_(NAME?), TSK_BANK_NAME, TSK_COUNTRY, TSK_CITY. Is payment card type generic or part of the CREDIT_CARD attributes "family?

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -552,9 +574,9 @@ protected void addNotify() {
      *
      * NOTE: currently hard coded to work for Credit Card only
      */
```
rcordovano: This comment is no longer relevant.

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -77,11 +78,16 @@
 public class Accounts extends Observable implements AutopsyVisitableItem {
```
rcordovano: Is there anything stopping us from putting this class in a datamodel.accounts package that we can then exclude from public API export? Would be get a lot of warnings about exporting a non-public type from the Visitors code?

Core/src/org/sleuthkit/autopsy/report/ReportHTML.java
```diff
@@ -254,8 +254,8 @@ private String useDataTypeIcon(String dataType) {
                 case TSK_REMOTE_DRIVE:
                     in = getClass().getResourceAsStream("/org/sleuthkit/autopsy/report/images/drive_network.png"); //NON-NLS
                     break;
-                case TSK_CREDIT_CARD_ACCOUNT:
-                    in = getClass().getResourceAsStream("/org/sleuthkit/autopsy/report/images/credit-card.png"); //NON-NLS
+                case TSK_ACCOUNT:
+                    in = getClass().getResourceAsStream("/org/sleuthkit/autopsy/report/images/account_menu.png"); //NON-NLS
```
rcordovano: Why is "menu" in the name of this image file?

Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
```diff
@@ -152,52 +156,81 @@ private void makeBlackboardArtifactTables() {
...
-             */
-            Set<BlackboardAttribute.Type> attrTypeSet = new TreeSet<>((BlackboardAttribute.Type o1, BlackboardAttribute.Type o2) -> o1.getDisplayName().compareTo(o2.getDisplayName()));
-            for (ArtifactData data : artifactList) {
-                List<BlackboardAttribute> attributes = data.getAttributes();
-                for (BlackboardAttribute attribute : attributes) {
-                    attrTypeSet.add(attribute.getAttributeType());
+            /* TSK_ACCOUNT artifacts get grouped by their TSK_ACCOUNT_TYPE
+             * attribute, and then handed off the default method for writing
+             * tables. */
+            if (type.getTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT.getTypeID()) {
```
rcordovano: I think we need some additional documentation for this cleverness. Let's say something about the way the synthesized data type is of the form  [BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT.displayName()] :  [BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ACCOUNT.getValueString].

millmanorama:
these are some goof comments/questions.  I will address them tomorrow.

Core/src/org/sleuthkit/autopsy/report/ReportHTML.java
millmanorama: because that was the name of the file in the icon library it came from.  I will change it to something more generic.

Core/src/org/sleuthkit/autopsy/report/ReportHTML.java
```diff
@@ -264,7 +264,17 @@ private String useDataTypeIcon(String dataType) {
                     iconFilePath = path + File.separator + iconFileName;
                     break;
             }
-        } else {  // no defined artifact found for this dataType 
+        } else if (dataType.startsWith(ARTIFACT_TYPE.TSK_ACCOUNT.getDisplayName())) {
+            /* TSK_ACCOUNT artifacts get separated by their TSK_ACCOUNT_TYPE
+             * attribute, with a compund dataType name, so they are not caught
+             * by the switch statement above. For now we just give them all the
+             * general account icon, but we could do soemthing else in the
```
millmanorama: will fix

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
millmanorama: I will try this, I think some of the classes will still need to be public to be seen outside of the accounts package and so there will not be the warnings you suggest.  having a new package will also allow me to move some of the classes into their own files, which I think will make them easier to understand.

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -77,11 +78,16 @@
...
 
     private static final Logger LOGGER = Logger.getLogger(Accounts.class.getName());
-    private static final BlackboardArtifact.Type CREDIT_CARD_ACCOUNT_TYPE = new BlackboardArtifact.Type(TSK_CREDIT_CARD_ACCOUNT);
+    private static final BlackboardArtifact.Type ACCOUNT_TYPE = new BlackboardArtifact.Type(BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT);
```
millmanorama: maybe i forgot to open it... here it is https://github.com/sleuthkit/sleuthkit/pull/715

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -77,11 +78,16 @@
...
     @NbBundle.Messages("AccountsRootNode.name=Accounts")
     final public static String NAME = Bundle.AccountsRootNode_name();
 
     /**
+     * This is a secret handshake with
+     * org.sleuthkit.autopsy.keywordsearch.TermComponentQuery
+     */
+    private static final String CREDIT_CARD_NUMBER_ACCOUNT_TYPE = "Credit Card Number";
```
millmanorama: I had been using Credit Card Number as the main thing we are really looking for and finding is the account numbers, but I think it is fine to use just "Credit Card".  I will review the variable names too.

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
millmanorama: that is true.  I will remove it.

Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
millmanorama: sure.  Also, in regards to you comment in the email, there is already a coupling between ReportHTML and TableReportGenerator in that ReportHTML assumes the dataTypeNames it is given are going to be artifact types.  The warnings logged on unknown dataTypeNames even use the term "artifact".

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
millmanorama: yes, to work in the tree it has to implement AutopsyVisitableItem but that interface is not public.

millmanorama: since that interface is not public, I can make it public, and then move it into the same unexposed package!  It has more than just accounts, so i will name it _private.

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -518,15 +523,32 @@ public void propertyChange(PropertyChangeEvent evt) {
...
         }
 
         @Override
         protected Node createNodeForKey(String key) {
-            return new AccountTypeNode(key);
+            if (key.equals(CREDIT_CARD_NUMBER_ACCOUNT_TYPE)) {
+                return new CreditCardNumberAccountTypeNode(key);
+            } else {
+                //Flesh out what happens with other account types here.
+                return new AbstractNode(Children.LEAF);
```
millmanorama: done in PR

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
millmanorama: I made an Acocunt class in the TSK datamodel with a Type enum

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
```diff
@@ -1030,8 +1055,8 @@ protected boolean createKeys(List<BinResult> list) {
                     + "     COUNT(blackboard_artifacts.artifact_id) AS count " //NON-NLS
                     + " FROM blackboard_artifacts " //NON-NLS
                     + "      JOIN blackboard_attributes ON blackboard_artifacts.artifact_id = blackboard_attributes.artifact_id" //NON-NLS
-                    + " WHERE blackboard_artifacts.artifact_type_id = " + BlackboardArtifact.ARTIFACT_TYPE.TSK_CREDIT_CARD_ACCOUNT.getTypeID() //NON-NLS
-                    + "     AND blackboard_attributes.attribute_type_id = " + BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ACCOUNT_NUMBER.getTypeID() //NON-NLS
+                    + " WHERE blackboard_artifacts.artifact_type_id = " + BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT.getTypeID() //NON-NLS
+                    + "     AND blackboard_attributes.attribute_type_id = " + BlackboardAttribute.ATTRIBUTE_TYPE.TSK_CREDIT_CARD_NUMBER.getTypeID() //NON-NLS
```
millmanorama: did we reach a conclusion in our discussion.  I can't remember...

Core/src/org/sleuthkit/autopsy/datamodel/Accounts.java
millmanorama: did we reach a conclusion on TSK_PAYMENT_CARD_TYPE?

PR #2355, opened 2016-09-23
Added options panel for adding and deleting user tag names

wishdasher:


Core/src/org/sleuthkit/autopsy/casemodule/services/NewUserTagNameDialog.java
```diff
@@ -0,0 +1,241 @@
...
+        this.addWindowListener(new WindowAdapter() {
+            @Override
+            public void windowClosing(WindowEvent e) {
+                doButtonAction(false);
+            }
+        });
+
+        /*
+         * Add a listener to enable the save button when the text field changes.
+         */
```
rcordovano: Also, some test results

- The first time a new tag name is added through the context menu, it has to be selected twice before the Delete Tag Name button is enabled.
- If a user adds a new tag name to via the Options panel and the Apply button, the new tag name is selected (correct behavior), but the Delete Tag Name button is not enabled (incorrect behavior).
- The tag name already exists pop up seems a bit harsh. I'd give it an info icon instead of an error icon.
- The label on the "Add Tag Name" button should be "New Tag Name", unless you are waiting for that change until the options panel consistency story.

rcordovano: One other thing, I think it would be better if the menu presentatino code decide what tag names to use to populate the menu rather than the TagsManager, i.e., I think we should replace TagManager.getAllTagNamesForDisplay with a getUserTagNames metyhod (and other methods, if necessary) and make the menu presenter decide how to merge the various sets of tags names for the menu. In other words, we should have a clear separation of concerns between the service maintaining the tags, the "business layer" of the application, and the "presentation layer" of the application.

Core/src/org/sleuthkit/autopsy/casemodule/services/NewUserTagNameDialog.java
```diff
@@ -0,0 +1,246 @@
...
+import java.awt.Toolkit;
+import java.awt.event.WindowAdapter;
+import java.awt.event.WindowEvent;
+import javax.swing.JFrame;
+import javax.swing.JOptionPane;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
+import org.openide.util.NbBundle;
+
+public class NewUserTagNameDialog extends javax.swing.JDialog {
```
rcordovano: Please lock down access to this class to package access and final.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNamesSettingsPanel.java
```diff
@@ -0,0 +1,318 @@
...
+import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.autopsy.coreutils.ModuleSettings;
+
+/**
+ * A panel to allow the user to create new tag names or to delete tag names that
+ * user has created in the past. List of user tag names is maintained in a
+ * properties file, able to be used across cases. Potentially room to add other
+ * tag name options in the future.
+ */
+public class TagNamesSettingsPanel extends javax.swing.JPanel implements OptionsPanel {
```
rcordovano: Can we make this package access final instead of public?

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -166,21 +222,31 @@ public synchronized TagName addTagName(String displayName, String description, T
...
+                long count = getContentTagsCountByTagName(existingTagName) + getBlackboardArtifactTagsCountByTagName(existingTagName);
+                if (count == 0) {
+                    addNewTagNameToTagsSettings(existingTagName);
+                    return existingTagName;
+                } else {
+                    throw new TagNameAlreadyExistsException();
+                }
+            }
+        }
+
```
rcordovano: I don't think throwing the TagNameAlreadyExistsException is currently useful given that there is a race condition inherent in the creation of tag names for multi-user cases. I think we should check if we have a matching TagName object in memory. If not, we should try to add it to the database. If an exception is thrown by SleuthkitCase.addTagName, it may be due to a database issue, or it may be due to an attempt to duplicate a tag added by another user. To allow for the latter, we should "refresh" the tags in memory from the database and see if we can get a matching tag name TagName that way. If the call to SleuthkitCase.getAllTagNames for the "refresh" throws, the exception should propagate.

We will keep TagNameAlreadyExistsException in the exception specification because to change it would be a public API change, and we will probably need it anyway, if we get to the point where we use the description and color columns in the tag names table.

rcordovano:
I am going to accept this and make some changes.

PR #2387, opened 2016-10-26
Added tags column and coloring for tagged nodes in table view

wishdasher:


wishdasher:
This also contains the changes made in AUT-1849 (PR#2356)

PR #2392, opened 2016-10-28
2223 Add HTML Data Content Viewer

tjmonson:
These changes add an HTML data content viewer to Autopsy. The viewer uses a JavaFX WebView to show HTML content. Additionally, the viewer contains a button to enable and disable any loaded JavaScript (JavaScript is initially disabled). The WebView is able to load any resources (such as images, stylesheets, or .js files) that have full URLs in the HTML. It can also load local resources if the .html file being viewed is a local file that has a .html extension. Files without .html extensions, whether local or from a disk image, are detected and loadable with some limitations.

Limitations:
The WebEngine for the WebView cannot local disk image resources if the HTML file being viewed is from a disk image. Also, the WebEngine cannot load local resources if the HTML file being viewed does not contain a .html extension.


millmanorama:
Thank you for this contribution.  This a great start to this feature.  I have a few concerns, but nothing big.  

-  This doesn't merge because of conflicts in 2 branding files.  These are from auto-generated time stamps and can be ignored.  just remove these files from the pull request
.- The java script button is a little large for my taste and feels out of place to me. 
- Some other comments are inline.

Core/src/org/sleuthkit/autopsy/corecomponents/Bundle.properties
```diff
@@ -81,6 +81,11 @@ DataContentViewerHex.goToPageTextField.err=Invalid page number
 DataContentViewerHex.setDataView.errorText=(offset {0}-{1} could not be read)
 DataContentViewerHex.title=Hex
 DataContentViewerHex.toolTip=Displays the binary contents of a file as hexidecimal, with bytes that are displayable as ASCII characters on the right.
+DataContentViewerHTML.title=HTML
+DataContentViewerHTML.toolTip=Displays the rendered HTML from the file.
+DataContentViewerHTML.acceptedMIMEType=text/html
```
millmanorama: I don't think this needs to be in the bundle.  It is never going to be internationalized is it?

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.corecomponents;
+
+import org.sleuthkit.autopsy.modules.filetypeid.FileTypeDetector;
+import java.awt.BorderLayout;
+import java.awt.Component;
+import java.util.logging.Level;
+import com.sun.javafx.application.PlatformImpl;
```
millmanorama: This is a internal API and subject to change.  You should probably be using javafx.application.Platform

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+    
+    private static final Logger logger = Logger.getLogger(DataContentViewerHTML.class.getName());
+    
+  
+    public DataContentViewerHTML() {
+        
+        try {
+            fileTypeDetector = new FileTypeDetector();
+        } catch (FileTypeDetectorInitException ex) {
+            Exceptions.printStackTrace(ex);
```
millmanorama: Use the logger here instead of Exceptions.printStackTrace

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+       jfxPanel = new JFXPanel();
+       setLayout(new BorderLayout());
+       add(jfxPanel, BorderLayout.CENTER);
+       createWebViewScene();     
+    }
+    
+    private void resetComponents() {
+        
+        //Disable JavaScript 
+        PlatformImpl.runLater(()-> {
```
millmanorama: see above about PlatformImpl vs Platform

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+    
+    /**
+     * Loads an HTML file into the webEngine as long as the engine is not null and the
+     * content is not null.
+     * @param localFilePath Not null. The  local file path to the HTML file to be loaded.
+     */
+    private void loadWebEngineFileURL(String localFilePath) {
+        
+        if(localFilePath != null && webEngine != null) {
+            PlatformImpl.runLater(() -> {
```
millmanorama: see above about PlatformImpl vs Platform

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+                webEngine.load("file://" + localFilePath);
+            });
+        }
+    }
+    
+    private void createWebViewScene() {
+        PlatformImpl.startup(() -> {
```
millmanorama: see above about PlatformImpl vs Platform

Platform does not have a startup method, but I don't see why this couldn't be done with runLater

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+            try {            
+
+                long size = file.getSize();
+                byte[] dataBytes = new byte[(int)size];
+                dataSource.read(dataBytes, 0, (int)size);
+                String HTML = new String(dataBytes);
+                loadWebEngineContent(HTML);
+
+            } catch (TskCoreException ex) {
+                Exceptions.printStackTrace(ex);
```
millmanorama: use the logger

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+        
+        if(node.getLookup().lookupAll(BlackboardArtifact.class).size() > 0) {
+            return false;
+        }
+        
+        AbstractFile file = node.getLookup().lookup(AbstractFile.class);
+        
+        if (file != null && file.getSize() > 0 && file.isFile()) {
+            
+            String acceptedMIMEType = NbBundle.getMessage(this.getClass(),
```
millmanorama: why is this in the bundle?  Also we the file type detector is not 100% reliable and we fairly often get html files tagged at text/xml instead.  They are not going to be viewable.

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+            //If the MIMEType has not already been detected, detect it.
+            if(file.getMIMEType() == null) {
+                
+                if(fileTypeDetector == null)
+                    return false;
+                
+                try {
+                    return fileTypeDetector.detect(file).equals(acceptedMIMEType);
+                } catch (TskCoreException ex) {
+                     Exceptions.printStackTrace(ex);
```
millmanorama: use the logger

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+                "DataContentViewerHTML.enableJavaScriptBtnTxt");
+        disableJavaScriptTxt = NbBundle.getMessage(this.getClass(), 
+                "DataContentViewerHTML.disableJavaScriptBtnTxt");        
+        
+        
+        logger.log(Level.INFO, "Created HTMLView instance: " + this); //NON-NLS
+    }
+    
+    private void initComponents() {     
+
```
millmanorama: The precedent of using initComponents is a legacy of previous practices.  I would recommend in-lining this in the constructor, probably also in-lining the createWebScene method.  @rcordovano , what do you think?

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskData.TSK_DB_FILES_TYPE_ENUM;
+
+
+
+/**
+ * HTML view of file contents. This viewer uses a JavaFX WebView to display HTML data.
+ */
+@ServiceProvider(service = DataContentViewer.class, position = 6)
+public class DataContentViewerHTML extends javax.swing.JPanel implements DataContentViewer {
```
millmanorama: this class should probably be final.

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHTML.java
```diff
@@ -0,0 +1,293 @@
...
+            return;
+        }
+        
+        dataSource = file;
+        
+        resetComponent();
+
+        //Local files with .html extensions are loaded with a file URL
+        if(dataSource.getType() == TSK_DB_FILES_TYPE_ENUM.LOCAL &&
+                dataSource.getLocalAbsPath().endsWith(".html")) {
```
millmanorama: i would support .htm files too

tjmonson:
Thanks for the phenomenal feedback! I'm working applying the changes now. Do you have any suggestions for the style and placement of the JavaScript button?

tjmonson:
Sorry about the long wait on getting those changes in. I think I was able to clear up most of the concerns.

rcordovano:
@tjmonson, I apologize for keeping you hanging all this time. We have gone a different with this, integrating HTML rendering into a non-public file viewer class family within an Application content viewer. I am closing this PR with thanks for your efforts.

PR #2406, opened 2016-11-11
Add substring keyword option to keyword lists

wishdasher:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Keyword.java
```diff
@@ -31,6 +32,9 @@
  */
 class Keyword {
 
+    static final String EXACT = NbBundle.getMessage(NewKeywordPanel.class, "NewKeywordPanel.exactButton.text");
```
millmanorama: Should these be an enum?

wishdasher: Oops, all that was doing was letting the table model get a bundle String value for the keyword type column (changed from a checkmark for a regex column). I changed it to get that display value directly.

Richard and I also talked about switching the keyword's model to use enums instead of two booleans that map to three types, but we didn't want to change the xml config format too much. I think it is possible to use enums instead for all instances of Keyword though; we would just have to change more classes.

PR #2417, opened 2016-11-24
clean up kws in preparation for supporting regex with spaces via overlapping 33KB chunks

millmanorama:
This is experimental cleanup and refactoring of the KWS text extracting and indexing code.  It is in preparation for the chunking changes coming to support regex across white space.  I did not run it through the regression tests because the regression tests don't work from other timezones.


millmanorama:
I am closing this now, as it got out of hand.  Many of the changes will be incorporated in to subsequent, more focused, pull requests.

PR #2429, opened 2016-12-13
Port Android data source ingest module into Python

wishdasher:


rcordovano:
Moved to branch pending review and integration.

PR #2435, opened 2016-12-16
2131 refactor text extractors

millmanorama:
Refactored TextExtractors and Ingester to use a single chunking algorithm.  Implemented ArtifactExtractor to chunk artifact contents through that same algorithm


millmanorama:
replaced by https://github.com/sleuthkit/autopsy/pull/2451

PR #2438, opened 2016-12-20
2153: Restart message is needed while Auto Ingest options changes

zhhl:


zhhl:
close to check if the new changes can be pushed in.

PR #2451, opened 2017-01-02
2131 refactor text extractors - Search Improvements

millmanorama:
target chunking refactoring to the Search Improvements branch

PR #2455, opened 2017-01-04
2132 32k chunks - Search Improvements

millmanorama:
refactor the chunking algorithm some more and reduce chunk size to 32k

should come after https://github.com/sleuthkit/autopsy/pull/2451


rcordovano:
Does not auto-merge, I will take a shot at the merge.

PR #2477, opened 2017-01-23
2197 profile options panel

wschaeferB:


wschaeferB:
Bug with saving filters within profile creation needs to be fixed

PR #2503, opened 2017-02-03
2198 run ingest modules wizard

wschaeferB:


wschaeferB:
Bug with saving filters within profile creation needs to be fixed

PR #2637, opened 2017-03-27
Android module fires ModuleDataEvent event for Contacts, CallLogs and …

raman-bt:
…Text message artifacts.


eugene7646:
What JIRA issue is this for?

PR #2643, opened 2017-03-30
Fix for a key word only being used in no more than one KW list

eugene7646:


eugene7646:
The problem we were trying to solve is if we have an exact match key word "bomb" in one list and a substring key word "bomb" in another list, only one of those key words was actually getting used during ingest search (see story 2487). This fixes that problem - where the "term" is identical but the type of search is different. What you said is true but I think that's probably fine.  Same way that we remove identical duplicate key words from the same list - why run identical search twice? After all, the search will be performed and the user will get the results they were looking for. At least that was my thinking.

millmanorama:
Sounds good.  I just wanted to make sure somebody had thought about it.

eugene7646:
Some more on this - I think it actually wouldn't be difficult to allow the same identical key word (and search type) to be run as part of multiple lists. I think in SearchRunner we could just loop over KW lists and then all key words within each list, instead of looping over a collection of all "Keyword"s that we aggregate. I think in that scenario we can just get rid of the keywordToList map completely. 

@bcarrier and/or @rcordovano , what do you think?

PR #2644, opened 2017-03-30
2462 unable to delete cases

wschaeferB:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
```diff
@@ -379,4 +381,33 @@ public void closeCaseResources(CaseContext context) throws AutopsyServiceExcepti
...
+    public void deleteCore(String textIndexName, String caseDirectory) throws KeywordSearchServiceException {
+        String coreName = textIndexName;
+        if (!caseDirectory.isEmpty()) {
+            IndexMetadata indexMetadata;
+            try {
+                indexMetadata = new IndexMetadata(caseDirectory);
+                //When I was debugging and there were multiple cores they had the same name 
+                //so the first one is as good as any.
+                if (!indexMetadata.getIndexes().isEmpty()){
+                    coreName = indexMetadata.getIndexes().get(0).getIndexName();
```
eugene7646: Can't we make this call (ln 405) from Case class instead?
coreName = indexMetadata.getIndexes().get(0).getIndexName();
If we can, then I think we should be able to re-use the old API.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
```diff
@@ -379,4 +381,33 @@ public void closeCaseResources(CaseContext context) throws AutopsyServiceExcepti
...
+                }
+            } catch (IndexMetadata.TextIndexMetadataException ex) {
+                Exceptions.printStackTrace(ex);
```
eugene7646: We don't use Exceptions.printStackTrace(ex) for exception logging. Please use Logger class instead from org.sleuthkit.autopsy.coreutils package. See examples in other classes.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
eugene7646: Oh no, I guess we can't, because it is in KWS package and has package access only.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
wschaeferB: Thanks for the catch. I have fixed this one.  I see I did this a few other times previously and will fix the other instances in a new branch.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
wschaeferB: My previous solution which I decided was too convoluted that in Case before deleteTextIndex(metadata.getTextIndexName()) was called I checked if the arguement was equal to "" if it was I instead called searchservice.deleteTextIndex(metadata.getCaseDirectory()) instead and then inside deleteTextIndex I figured out if it was a path or just a core name if it was a path I got the IndexMetadata and checked to see if it had contents.   I however felt it was an inferior solution.

eugene7646: I agree. It does sound like a fairly convoluted alternative. However, given that there is a public API change I am going to defer the final decision to Richard and Brian.

rcordovano:
Please resubmit targeting rc-280 branch, thanks!

PR #2645, opened 2017-03-30
2505 profile name bug

wschaeferB:


Core/src/org/sleuthkit/autopsy/ingest/IngestProfiles.java
```diff
@@ -50,7 +51,7 @@
         List<IngestProfile> profileList = new ArrayList<>();
         if (directoryListing != null) {
             for (File child : directoryListing) {
-                String name = child.getName().split("\\.")[0];
+                String name = child.getName().replace(PROFILE_FILE_EXT, ""); //remove the extension
```
eugene7646: Can you please use FilenameUtils.removeExtension() from org.apache.commons.io.FilenameUtils package instead? It has a large collection of methods for manipulating file names. We use it in many places throughout the code.

wschaeferB:
I have adjusted it to use the apache removeExtension method, thank you for the suggestion I was ignorant to that methods existence.  And now people can even have profiles with the string .properties in the name, so definitely a better solution.

PR #2652, opened 2017-03-31
2371 - Experimental: remove jars from repo and download them with Ivy

millmanorama:


rcordovano:

PR #2662, opened 2017-04-04
2518 analyze data asap

wschaeferB:
This story also resolves issue 2280 as startIngest is no longer called in the storeSettings method of the config panel.


rcordovano:
Please target the rc-280 branch for this pull request, thanks!

PR #2675, opened 2017-04-10
Catch blocks added to silently catch and return.

dgrove727:


rcordovano:
Please re-submit after making discussed modifications.

PR #2677, opened 2017-04-11
2461 disable menu for auto ingest

wschaeferB:


wschaeferB:
Previously reviewed as PR #2672 for develop, remade for rc-280 base instead of develop

PR #2680, opened 2017-04-11
Ingest search produces same number of KW hits as ad-hoc search (2495)

eugene7646:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Keyword.java
```diff
@@ -68,6 +60,44 @@
...
+     *
+     * NOTE: The addition of keyword list name and original search term was
+     * added to facilitate proper de-duping of results of periodic keyword
+     * searches that does not lose any keyword hits. Without this addition when
+     * using substring search feature during ingest, if there are multiple
+     * searches on differnt keyword lists that produce the same keyword hit,
+     * that hit is only going to be displayed in results of one of the list. For
+     * example, two substring searches, such as "pass" and "enger", will be
+     * missing one copy of any shared entries (i.e., "passenger" will only show
+     * up on one list). See JIRA story 2495.
```
APriestman: Since this isn't a TODO and you've explained it well, I'm not sure we need the JIRA reference.

rcordovano: I actually like to have JIRA references for TODOs. TODOs without JIRAs are highly unlikely to get done, so the JIRA reference indicates that we have a story for it, and lets us know to delete the TODO comment when we are done. In other words, if we think something needs a TODO, we should not just comment it, we should write a story for it. Persaonlly, I have been either writing stories for TODOs or deleting them as I find them...

eugene7646: My thinking was that this problem is explained much better in the story. I think a year from now my explanation may not be sufficient to describe the full picture. But I don't have a strong preference either way, we can delete the JIRA reference.

APriestman: Either way is fine with us. But please do fix the typo "differnt"

eugene7646: Done!

PR #2685, opened 2017-04-12
2529 remove unused methods and dependency on solr 4.9.1

millmanorama:


rcordovano:
Does not auto-merge, please correct and re-submit, thanks!

PR #2692, opened 2017-04-12
2541 renaming keyword rc 280

wschaeferB:


rcordovano:
Regarding the comment from @millmanorama:

1. If we are going to make the button bigger, we should just do it now. 
2.  I think that changing the use of the shouldAdd and replace flags would make GlobalListSettingsPanel.copyAction cleaner. In my opinion, the use of flags to implement the logic of a method is generally harder to read and maintain - our human brains are not that good at keeping track of flags when we are writing code, and the fewer branches there are, the better. In this case, this method uses two flags, shouldAdd (8 references) and replace (2 references).  The first three references to shouldAdd could be replaced by returning false. The reference on line 140 could be replaced by  returning true. Then the code on lines 133 through 137 and 143 through 147 could be combined into something like this pseudo code, which eliminates the replace flag also (which could also be eliminated by assigning directly to if shouldAdd was retained):

if  list exists
    if confirm dialog
        add list
        do dialog
        return true
   else
      return false
else
    return true
    
Personally, I find that easier to understand, but this is definitely subjective.

millmanorama:
Thank you for the more detailed explanation, it is exactly what I meant by "inline shouldAdd"

wschaeferB:
I simplified the logic in a manner close to what was suggested in the first commit, I then made the buttons larger using the button size on the interesting file sets as a guide, after I had resized them to be similar I went through and made the buttons on interesting file defs and profiles the same size.  And swapped the copy and new icons on the keyword list panel for consistantcy as well while I was in there.

This leaves Hash databases, file extension mismatch, file types, tags, and external viewer all with their own button sizes but I think that is best left for another story because when I took a look at resizing those too I noticed that the text on the Import Databases button will either require the buttons be larger or be cramped if changed to the size I have just made the keyword search buttons.

PR #2703, opened 2017-04-13
2483 add new line before pasting into keywords list jTextArea

wschaeferB:


wschaeferB:
made PR on wrong base

PR #2704, opened 2017-04-13
2483 add new line before pasting into keywords list jTextArea

wschaeferB:


wschaeferB:
I have come around to the idea that paste will always append in this keyword list box, because the number of conditionals necessary to make the text play nicely when pasted in the center of lines was getting out of hand.  So it should work as your originally expected now, hopefully it still makes sense to work that way.

The new line is now conditional, on the text area having contents and not ending in a new line. 

I have removed the keywordTextArea from the auto generated code so that it can be easily modified without code duplication.

PR #2705, opened 2017-04-13
New catch blocks added to silently catch/return on errors.

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/ImageExtractor.java
```diff
@@ -211,6 +213,13 @@ void extractImage(AbstractFile abstractFile) {
         HWPFDocument doc = null;
         try {
             doc = new HWPFDocument(new ReadContentInputStream(af));
+        } catch (OldFileFormatException ex) {
```
eugene7646: I think we should combine the two "catch" blocks into a single catch block and move the comments there. The same applies to all other changes in the pull request.

eugene7646: Also, we have a lot of exception catch-all blocks when we are actually retrieving the data from the documents. All those catch blocks log a warning upon catching an exception. @rcordovano do we want to keep those warnings?

rcordovano: No, I don't think we should log those, either. This is a best effort endeavor, and the WARNINGS you mentioned are in the same category as the others, i.e., not really helping anybody, just inflating the logs.

However, the TskCoreException catches where it is a database issue should log SEVEREs, not WARNINGs. For example, the TskCoreException catch in isImageExtractionSupported and the AbstractFile.hasChildren call in extractImage.

eugene7646: I still see all the WARNING logs that we have talked about removing. See Ln 250 in the new code. There are a lot of these " catch (Exception ex)" blocks remaining in the code. I believe @rcordovano has decided to no longer log warnings inside those catch blocks.

rcordovano: The intent is to have no warnings from the calls into the POI library code.

dgrove727: I did another commit to remove the rest of these, and then I did a second commit to remove an unused error from the Bundle.properties file.

rcordovano:
Having all of the POI calls in a try with a catch that returns null is good, and making the TskCoreExceptions logged as SEVERE rather than as WARNING is good. However, I'm thinking that since we kept the firewalls for possible runtime exceptions, we should make those log SEVERE error messages. If we only do WARNINGs, nobody will notice - they won't be reported by automated regression testing, etc. @dgrove727  and @eugene7646, what are your thoughts?

eugene7646:
Makes sense to me.

dgrove727:
Updated to switch some of the 'WARNING' log levels to 'SEVERE'.

PR #2713, opened 2017-04-18
2536: Review ChildFactory uses for long-running tasks

raman-bt:


Core/src/org/sleuthkit/autopsy/datamodel/InterestingHits.java
```diff
@@ -345,23 +347,28 @@ private HitFactory(String setName) {
...
+               return true;
             }
+            
+            interestingResults.getArtifactIds(setName).forEach((id) -> {
+                try {
+                    BlackboardArtifact art = skCase.getBlackboardArtifact(id);
+                    
+                    artifactHits.put(id, art);
+                    list.add(id);
+                } catch (TskCoreException ex) {
```
rcordovano: Since this is a case database error, will you please change the log level to SEVERE instead of WARNING, while we are in this code?

Core/src/org/sleuthkit/autopsy/datamodel/HashsetHits.java
```diff
@@ -370,23 +372,28 @@ protected void removeNotify() {
...
+               return true;
+            }
+            
+            hashsetResults.getArtifactIds(hashsetName).forEach((id) -> {
+                try {
+                    BlackboardArtifact art = skCase.getBlackboardArtifact(id);
+                    
+                    artifactHits.put(id, art);
+                    list.add(id);
+                } catch (TskException ex) {
```
rcordovano: Since this is a case database error, will you please change the log level to SEVERE instead of WARNING, while we are in this code?

Core/src/org/sleuthkit/autopsy/datamodel/InterestingHits.java
raman-bt: Done

Core/src/org/sleuthkit/autopsy/datamodel/HashsetHits.java
raman-bt: Done

rcordovano: Thanks!

PR #2714, opened 2017-04-18
Revert to Solr 4 schema + content_str and chunk_size.

esaunders:


KeywordSearch/ivy.xml
```diff
@@ -20,7 +20,7 @@
     <dependencies>
         <!-- Solr -->
         <dependency conf="solr-libs->default" org="org.apache.solr" name="solr-cell" rev="4.9.1"/>
-        <dependency conf="solr-war->default" org="org.apache.solr" name="solr" rev="4.9.1"/> <!-- the war file -->
+        <dependency conf="solr-war->default" org="org.apache.solr" name="solr" rev="4.9.1" transitive="false" /> <!-- the war file -->
```
millmanorama: I am really surprised this was needed.  Shouldn't setting transitive to false, result in Ivy pulling down less stuff?

esaunders: Without it I was ending up with a war file that was not functional. The documentation for transitive isn't particularly enlightening ("true to resolve this dependency transitively, false otherwise").

PR #2721, opened 2017-04-19
Remove python semicolons

rcordovano:


rcordovano:
Re-targeting to develop.

PR #2723, opened 2017-04-19
Allow user to add agency logo

rcordovano:


rcordovano:
Targeted to wrong branch, misleading branch name

PR #2726, opened 2017-04-19
Updated docs link and uncommented 4.0-only code in Python tutorials

rcordovano:


dgrove727:
Everything looks good, except that there are some old URLs that should reference the latest documentation (the associated JIRA ticket suggests 4.4 since that will be around the corner). All eight files have outdated URLs. A search for "4." and "3." will reveal these.

millmanorama:
Should the version numbers in the URLs be generated as part of the build, so they don't get out of sync?

rcordovano:
Replaced by PR# 2573.

PR #2733, opened 2017-04-20
2572 agency logo

wschaeferB:


wschaeferB:
Logo changes not being saved between runs

wschaeferB:
1. Has been fixed and now you can use any image the browser will show you (png, jpeg, jpg, or gif)  the logo will now also retain its original file name in your report.
2.This is no longer an issue because you can no longer type in that field, and must use the Browse button. This was done because we were not validating the path entered was to an image file.

Core/src/org/sleuthkit/autopsy/casemodule/GeneralFilter.java
```diff
@@ -37,14 +37,17 @@
...
     public static final List<String> VIRTUAL_MACHINE_EXTS = Arrays.asList(new String[]{".vmdk", ".vhd"}); //NON-NLS
     public static final String VIRTUAL_MACHINE_DESC = NbBundle.getMessage(GeneralFilter.class,
-            "GeneralFilter.virtualMachineImageDesc.text");    
+            "GeneralFilter.virtualMachineImageDesc.text");
 
     public static final List<String> EXECUTABLE_EXTS = Arrays.asList(new String[]{".exe"}); //NON-NLS
     public static final String EXECUTABLE_DESC = NbBundle.getMessage(GeneralFilter.class, "GeneralFilter.executableDesc.text");
-    
+
+    public static final List<String> GRAPHIC_IMAGE_EXTS = Arrays.asList(new String[]{".png", ".jpeg", ".jpg", ".gif"}); //NON-NLS
```
millmanorama: I haven't looked at this in details or tested it, but in most places (in Java especially) that png, jpeg, and gif are supported, bmp is also

wschaeferB: It does indeed look like bmp would be supported by all 4 major browsers, chrome, firefox, IE, and safari.  So I can include it do a test to ensure it works, I assume based on this comment that is something we would like to have.

PR #2739, opened 2017-04-24
2578 fix text retrieval on "upgraded cases"

millmanorama:


rcordovano:
Does not auto-merge, please address and resubmit, thanks!

PR #2757, opened 2017-04-27
Change to manifest file parsing (2600)

eugene7646:
AutopsyManifestFileParser looks for 'AutopsyManifest' root element instead of 'Manifest' to resolve a customer's issue

IMPORTANT!!!! All existing Autopsy Manifest.xml files MUST be updated to use 'AutopsyManifest' root element!


rcordovano:
We are going to go with a somewhat different solution, so I am going to close this PR.

PR #2761, opened 2017-04-28
2119 issues with developer test

zhhl:


zhhl:
There is not much result from Google search for error "No suitable relative path from", here is the only one I can find a solution: http://stackoverflow.com/questions/19514228/no-suitable-relative-path-in-ant-os-windows. 5 for maxParentLevels is a suggestion from this link. The default value 2 is not working for Raman's setting and 5 is just one level up than double of the default value 2.  So I think it's not too small like 3 just one level higher or too large. So I chose 5.

rcordovano:
Does not auto-merge, please update and re-submit, thanks!

PR #2778, opened 2017-05-04
persist multi column sorting in DataResultViewerTable

millmanorama:


rcordovano:
Please see comment in JIRA.

PR #2793, opened 2017-05-09
2210 persist sort

millmanorama:
persist the column sorting info plus some minor cleanup

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
+++ b/Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
@@ -26,15 +26,18 @@ import java.awt.Graphics;
 import java.awt.dnd.DnDConstants;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.beans.FeatureDescriptor;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
```diff
@@ -26,15 +26,19 @@
 import java.awt.dnd.DnDConstants;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.beans.FeatureDescriptor;
```
rcordovano: The FeatureDescriptor import is not used.

Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
```diff
@@ -88,6 +94,7 @@
     // the column started and where it ended up.
     private int startColumnIndex = -1;
     private int endColumnIndex = -1;
+    private OutlineView ov;
```
rcordovano: Since we are touching this file, I would prefer to rename "ov" to "outlineView", or even just "view." Highly abbreviated variable names are discouraged by the coding guidelines.

PR #2813, opened 2017-05-30
1659 sort thumbnailviewer

millmanorama:
- use the persisted sort information to sort the Thumbnail result viewer
- add a GUI to choose sort criteria from the Thumbnail result viewer


millmanorama:
Latest commit fixes the bug @raman-bt  reported where the label wasn't resetting.

raman-bt:
Verified that "Sorted by" labels are now getting reset.

PR #2815, opened 2017-05-31
2660: Reorder columns in Email Messages results view table

raman-bt:


wschaeferB:
If any of our gold images have email messages artifacts those gold images will need to be rebuilt after this.

PR #2825, opened 2017-06-05
Potential NPE bug fix (2629)

eugene7646:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/Manifest.java
```diff
@@ -46,7 +46,11 @@ public Manifest(Path manifestFilePath, String caseName, String deviceId, Path da
         this.dateFileCreated = new Date(attrs.creationTime().toMillis());
         this.caseName = caseName;
         this.deviceId = deviceId;
-        this.dataSourcePath = dataSourcePath.toString();
+        if (dataSourcePath != null) {
```
rcordovano: While we are in this class, we should take a few minutes to document it, including documenting that getDataSourcePath might return an empty path. In fact, I'm wondering if the expectation is that it is valid to pass a null path to the constructor, why not just return null from getDataSourcePath? And does this apply to all of the fields and getters?

rcordovano: While we are in this class, we should take a few minutes to document it, including documenting that getDataSourcePath might return an empty path. In fact, I'm wondering if the expectation is that it is valid to pass a null path to the constructor, why not just return null from getDataSourcePath? And does this apply to all of the fields and getters, i.e., null signifies no data?

PR #2830, opened 2017-06-07
Limit the results displayed from istat in the Metadata tab

APriestman:


Core/src/org/sleuthkit/autopsy/contentviewers/Bundle.properties
```diff
@@ -16,4 +16,5 @@ Metadata.title=File Metadata
 Metadata.toolTip=Displays metadata about the file.
 Metadata.nodeText.nonFilePassedIn=Non-file passed in
 Metadata.nodeText.text=From The Sleuth Kit istat Tool\:
-Metadata.nodeText.exceptionNotice.text=Error getting file metadata\:
\ No newline at end of file
+Metadata.nodeText.exceptionNotice.text=Error getting file metadata\:
+Metadata.nodeText.truncated=(results truncated)
```
rcordovano: Please add the new message using a message annotation instead of writing it directly to bundle.properties, thanks!

PR #2843, opened 2017-06-13
made Artifact(s) plural in package and module name

mrtizmo:
renamed package to have plural 'Artifact(s)' in the name. updated resource bundle keys, message, and other fields to use the plural form.


rcordovano:
Has merge conflicts, please address and resubmit when time permits, thanks!

PR #2860, opened 2017-06-16
2732 only store the sort info for visible columns

millmanorama:


millmanorama:
closing to retarget to release branch

PR #2867, opened 2017-06-19
2676 GIFs now included in thumbnails on HTML reports

wschaeferB:


wschaeferB:
This PR has been rendered unnecessary by PR #2876 https://github.com/sleuthkit/autopsy/pull/2876 which makes ReportHTML save the buffered image directly to its Report Directory instead of copying a thumbnail.

PR #2869, opened 2017-06-20
2706 regex hit count

millmanorama:
fix the hit counts so that the tree and the table match.

There is some other minor cleanup and refactoring.  I will make comments to call out the changes specific to this bug fix


Core/src/org/sleuthkit/autopsy/datamodel/KeywordHits.java
```diff
@@ -635,7 +644,7 @@ protected Sheet createSheet() {
             ss.put(new NodeProperty<>(NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.name"),
                     NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.displayName"),
                     NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.desc"),
-                    keywordResults.getKeywordInstances(setName, keyword).size()));
+                    countTotalDescendants()));
```
millmanorama: show the correct total using the instance value

Core/src/org/sleuthkit/autopsy/datamodel/KeywordHits.java
```diff
@@ -803,7 +821,7 @@ protected Sheet createSheet() {
             ss.put(new NodeProperty<>(NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.name"),
                     NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.displayName"),
                     NbBundle.getMessage(this.getClass(), "KeywordHits.createSheet.filesWithHits.desc"),
-                    keywordResults.getKeywordInstances(setName, keyword).size()));
+                    keywordResults.getArtifactIds(setName, keyword, instance).size()));
```
millmanorama: again, use the instance value here

Issue #176, opened 2013-03-30
Forensic image hashing


aoighost:
Currently, Autopsy will hash files inside of the image, but it fails to hash the image itself. Implementing the ability to hash the image would help with validating the image itself.

bcarrier:
This was added in 4.10.0

Issue #175, opened 2013-03-30
SHA1/256/512 hashes in autopsy browser


peterclemenko:
Currently, Autopsy only displays MD5 hashes. This can be a problem in some situations where a SHA hash is required as well (I.E. forensics competitions, court cases, et. al). Please implement SHA hashing for files in the autopsy browser.

PR #2871, opened 2017-06-20
eam db configuration and options panel UI revision

mrtizmo:
This resolves CE Issues #174, #175, and #176.

- new Options panel and Dialog UI
- main options panel has read-only db settings and extra buttons to add data
- configure button opens single dialog that can be used to configure postgres/sqlite or disable the module.
- sqlite db file name will always end in .db
- during configuration of a db platform, the user is now forced to enter configuration settings, then click Test, optionally click Create, and finally click Ok.
- clicking Cancel will ignore any changes made to values in the settings dialog. It will not delete a newly created Sqlite Db though.
- password field is now a proper JPasswordField box. The password is now stored as an encrypted value in the properties file.
- extracted the TextConverter class from the UserPreferences class and made it publicly accessible. Moved it into the coreutils package.
- created a new EamDbUtil class to store common db platform testing and setup code.


APriestman:
Some comments from my initial testing:

Main options panel
* After I disabled the database, a red message appeared in the middle of the main options panel saying "You must configure the database." Disabling EAM isn't an error so I don't think we should have bold red text like this. Possibly have a label down near the buttons saying they're disabled because the database must be configured.
* There's a lot of empty space between the database config section and the buttons

Database config panel

* I personally don't like the red text giving instructions since we tend to use red for errors. I'd like it better just for the errors like "Database initialization failed"
* I hit the test button using the default postgres settings (no username or password), and I got a red pop-up error on the main Autopsy window. If postgres always needs a user name and password (I believe it does), we shouldn't even attempt the test. Perhaps the button should be disabled.
* I also got a red pop-up error when I used an illegal character in the database name. We should do the same validation we use elsewhere before trying to create the database
* In general, it'd be nice to catch the database errors and display them in the config panel to the user instead of doing the pop-ups (I also got one for an incorrect password). But it might be hard to do with our postgres code.
* The following sequence caused invalid settings to get saved:
** Set up SQLite, do test/create, click OK
*** Database config will show the SQLite database
** Click configure again, select the Postgres radio button, click the X in the upper right of the dialog
*** The default postgres settings appear on the main options panel

mrtizmo:
@APriestman I'm doing revisions based on your feedback. One question for you, how do I stop those red pop-ups? I'm catching the exception, but my LOGGER message is not displayed and the exception is thrown as if I didn't catch it. Do you see what I'm doing wrong?

mrtizmo:
Also, regarding input validation. I'm doing the same input validation that is happening on the Multi-User case panel.
1. Not empty fields
2. Port > 0.

There is no regex checking or anything like that on input values.

mrtizmo:
Pushed some changes:
1. added input validation to db settings. 
2. changed red text to be red if it's an error, else black. 
3. changed postgres connection to use a properties object to store the user/pass, so it can do it's own quoting. 
4. reduced the large area between the db config button and the lower buttons on the main options panel.

mrtizmo:
This should address all of @APriestman 's concerns and a few other fixes, except for the red pop-up issue.

* send database settings exception messages back to config window and display them in red. 
* surround guidance message panel with a scrollpane, so main window does not change size with added error messages.
* added red/green check boxes to the Create button. 
* correctly clear the red/green check boxes by the Test and Create buttons when a change is made.
* Do not save state until OK button is clicked, so either the X or the Cancel button will cause no change (except the possible creation of a db).

APriestman:
It doesn't seem like I can disable the EAM - when I select "Disable" the Ok button is greyed out.

It looks like the red pop-up issue is related to the weirdness you already commented on around line 185 of PostgresEamDbSettings where it looks kind of like an exception escapes. There's something weird going on there that we don't understand either, so don't worry about that for now.

mrtizmo:
Fixed the Ok button, so the EAM can be disabled.

PR #2876, opened 2017-06-21
2675 ReportHTML saves bufferedImages directly to report directory 

wschaeferB:
instead of copying from thumbnail cache


rcordovano:
Taking this work into release-4.4.1 branch instead.

PR #2880, opened 2017-06-22
Image Gallery opens un docked

millmanorama:


rcordovano:
Taking PR#2885 instead.

PR #2892, opened 2017-06-26
2780 multi line support for Results Content Viewer table

wschaeferB:
Fixes display of Value fields with multi-line content in the Results Content Viewer.  Value fields with extremely long single line content will still need the user to adjust column widths and possibly copy out the cell contents and paste them into a text editor.


wschaeferB:
Addressed issue with extremely long lines, they now wrap and height of cell updates to account for wrapped lines in most cases.

PR #2896, opened 2017-06-27
stop showing counts in file type tree once there are more than 200k files…

millmanorama:
… in the db.


Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByMimeType.java
```diff
@@ -170,12 +146,59 @@ private void populateHashMap() {
...
+    /**
+     * Should the nodes show counts?
+     *
+     *
+     * @return True, unless the DB has more than 200k rows.
+     */
+    private boolean shouldShowCounts() {
+        if (showCounts) {
+            try {
+                if (skCase.countFilesWhere("1") > 200000) {
```
esaunders: This results in an exception in PostgreSQL...argument of where must be type boolean, not integer type. I believe that changing this to "1 = 1" works for both SQLite and PostgreSQL.

millmanorama:
@esaunders I just pushed the change you suggested

PR #2899, opened 2017-06-27
Hashdb import error handling and code cleanup

mrtizmo:
- Removed global list from AbstractSqlEamDb. Also removed preparebulkglobalinsert().
- Manage the list of imported global files in the ingest thread and pass the completed list to bulk global insert. 
- Added a Type field to the bulk global insert, so in the future, we can have global lists of different Types. 
- Converted text messages to use @Messages. 
- Removed pop-up windows and now use dialog label for dialog content errors.
- For ingest errors, use an Inbox Message.


rcordovano:
Please resolve merge conflicts and resubmit.

PR #2901, opened 2017-06-28
Inform user that Correlation Properties are global

mrtizmo:
- Added sentence to correlation properties dialog to indicate that those settings are global and impact all users of the central database.


mrtizmo:
Corrected wording per @bcarrier's feedback.

PR #2910, opened 2017-06-29
Added documentation for central repository

APriestman:


rcordovano:
I suggest we call this "Central Repository User Guide" or something like that, thoughts?

rcordovano:
Thanks, looks great!

PR #2912, opened 2017-06-29
Issue192

mjkoo:
Resolves domain/phone number normalization issues as discussed in correlation engine issue 192.

Note that the phone number normalization only removes punctuation as discussed, there are other potential issues when attempting to normalize numbers but this should cut down on variations that prevent correlation.


bcarrier:
Can we do this normalization in the DB class so that all data going into the DB is normalized (regardless of it if it is coming from the listener or ingest module)?  Seems like we should be doing the lowercase for all 'values' in all tables and path.

millmanorama:
Do we care that this nomralization strips the +  that indicates this is an
international number and the first digit(s) are to be interpreted as the
country code and not the beginning of a local number?

bcarrier:
Oh yea. We walked about that before, but it's true that this will remove those. @mjkoo, can you add that by either a fancier pattern or by checking if it starts with a '+' before the cleanup and then adding it back on after.

PR #2936, opened 2017-07-10
2804 bg updatedisplay name

millmanorama:


Core/src/org/sleuthkit/autopsy/datamodel/FileTypes.java
```diff
@@ -149,4 +180,63 @@ protected AbstractNode defaultVisit(Content di) {
...
+        }
+
+        @Override
+        public void update(Observable o, Object arg) {
+            updateDisplayName();
+        }
+
+        abstract String getDisplayNameBase();
+
+        abstract String geQuery();
```
esaunders: What's a geQuery?

PR #2941, opened 2017-07-12
2075 single query for mime counts

millmanorama:


Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
```diff
@@ -325,10 +325,9 @@ String getDisplayNameBase() {
         }
 
         @Override
-        String geQuery() {
-            return createQuery(filter);
+        long calculateItems() throws Exception {
```
esaunders: What is it in calculateItems that throws Exception? Could it be more specific?

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByMimeType.java
```diff
@@ -61,7 +61,7 @@
      * which exist in the database. This hashmap will store them with the media
      * type as the key and a list of media subtypes as the value.
```
esaunders: We should update the comment to describe the change in the existingMimeTypes data structure.

millmanorama:
@esaunders I updated this pull request to address your comments and do some other cleanup

PR #2942, opened 2017-07-12
Adding local directory type

APriestman:


Core/src/org/sleuthkit/autopsy/casemodule/services/FileManager.java
```diff
@@ -510,26 +511,26 @@ private static synchronized String generateFilesDataSourceName(SleuthkitCase cas
      * @throws TskCoreException If there is a problem completing a database
      *                          operation.
      */
-    private AbstractFile addLocalFile(CaseDbTransaction trans, VirtualDirectory parentDirectory, java.io.File localFile,
+    private AbstractFile addLocalFile(CaseDbTransaction trans, AbstractFile parentDirectory, java.io.File localFile,
```
rcordovano: Perhaps use SpecialDirectory instead of AbstractFile as teh type for the parentDirectory parameter?

Core/src/org/sleuthkit/autopsy/datamodel/SpecialDirectoryNode.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public boolean isLeafTypeNode() {
+        return false;
+    }
+
+    @Override
+    public String getItemType() {
+        // use content.isDataSource if different column settings are desired
+        return DisplayableItemNode.FILE_PARENT_NODE_KEY;
+    }
```
rcordovano: Let's be sure to test that the column orders and sorts apply cleanly, given that VirtualDirectory there are additional columns in the property sheet for virtual directories when they are data sources. If this is broken, it is probably an existing bug in the way we handle persisting the sorts and column orders for virtual directories.

PR #2947, opened 2017-07-13
2811 - Warning not to use SQLite databases for multi-user purposes added

wschaeferB:


bcarrier:
This should go towards 4.4.1

PR #2949, opened 2017-07-15
first pass at content cache

millmanorama:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -522,26 +530,20 @@ else if (attribute.getAttributeType().getValueType() == BlackboardAttribute.TSK_
...
-        } catch (TskCoreException ex) {
-            LOGGER.log(Level.WARNING, "Getting file failed", ex); //NON-NLS
+            Content content = contentCache.get(artifact.getObjectID(), () -> artifact.getSleuthkitCase().getContentById(artifact.getObjectID()));
+            if (content == null) {
+                return Lookups.fixed(artifact);
+            } else {
+                return Lookups.fixed(artifact, content);
+            }
+        } catch (ExecutionException ex) {
+            //JMTODO: Clean this up.  Either swallow it with appropriate logging/notification, or propogate it, not both.
```
rcordovano: Let's resolve this TODO, please. I would not throw an unchecked exception here, and there is nothing for the calling code to do, so I'm inclined to  log the error and continue to construct the BlackboardArtifactNode with an incomplete lookup. The consequence would be evident ion the content viewers, right? Thoughts?

millmanorama: done in https://github.com/sleuthkit/autopsy/pull/2961

PR #2951, opened 2017-07-17
Introduced specialized PostgreSQL query for file types by extension q…

esaunders:
…ueries.


Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
```diff
@@ -330,24 +331,39 @@ long calculateChildCount() throws TskCoreException {
...
+                + " AND (NULL "; //NON-NLS
+
+        if (skCase.getDatabaseType().equals(TskData.DbType.POSTGRESQL)) {
+            // For PostgreSQL we get a more efficient query by using builtin
+            // regular expression support and or'ing all extensions. We also
+            // escape the dot at the beginning of the extension.
+            // We will end up with a query that looks something like this:
+            // OR LOWER(name) ~ '(\.zip|\.rar|\.7zip|\.cab|\.jar|\.cpio|\.ar|\.gz|\.tgz|\.bz2)$')
+            query += "OR LOWER(name) ~ '(\\";
+            query += StringUtils.join(filter.getFilter(), "|\\");
```
millmanorama: do we need to use toLowerCase here like we do in the SQLite version?

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
```diff
@@ -330,24 +331,39 @@ long calculateChildCount() throws TskCoreException {
...
-        query.append("(dir_type = ").append(TskData.TSK_FS_NAME_TYPE_ENUM.REG.getValue()).append(")"); //NON-NLS
-        if (UserPreferences.hideKnownFilesInViewsTree()) {
-            query.append(" AND (known IS NULL OR known != ").append(TskData.FileKnown.KNOWN.getFileKnownValue()).append(")"); //NON-NLS
-        }
-        query.append(" AND (NULL"); //NON-NLS
-        for (String s : filter.getFilter()) {
-            query.append(" OR LOWER(name) LIKE LOWER('%").append(s).append("')"); //NON-NLS
+    private String createQuery(FileTypesByExtension.SearchFilterInterface filter) {
+        if (filter.getFilter().isEmpty()) {
+            return "";
```
millmanorama: I think returning an empty string is wrong here.  In the intended use case, this gets passed to SleuthkitCase.countFilesWhere() or findAllFilesWhere() where it is used as the predicate expression in the Where clause.  If this is empty, ie  "select count(*) as count from tsk_files where ", it is a SQL syntax error.  I would rather throw an exception or return something that evaluates to true (in both Postgres and SQLite); i think you recommended "1=1" as such an expression previously.

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
esaunders: Yes. I could use the PostgreSQL ~* operator but I think toLowerCase is more obvious. I'll make the change.

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
esaunders: I'm leaning towards throwing an IllegalArgumentException if the filter list is empty since this is clearly a programming error that we want to be aware of. Thoughts?

millmanorama:
yes, that makes sense to me.

esaunders:
@millmanorama I've addressed your review feedback. Let me know if there is anything else.

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
```diff
@@ -358,7 +362,8 @@ private String createQuery(FileTypesByExtension.SearchFilterInterface filter) {
             // We will end up with a query that looks something like this:
             // OR LOWER(name) ~ '(\.zip|\.rar|\.7zip|\.cab|\.jar|\.cpio|\.ar|\.gz|\.tgz|\.bz2)$')
             query += "OR LOWER(name) ~ '(\\";
-            query += StringUtils.join(filter.getFilter(), "|\\");
+            query += StringUtils.join(filter.getFilter().stream()
+                    .map(String::toLowerCase).collect(Collectors.toList()), "|\\");
```
millmanorama: This is pretty nit picky, but I would use Collectors.joining() instead of StringUtils.join, since there is already a stream involved.

Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByExtension.java
```diff
@@ -340,24 +342,43 @@ long calculateChildCount() throws TskCoreException {
...
+
+        if (skCase.getDatabaseType().equals(TskData.DbType.POSTGRESQL)) {
+            // For PostgreSQL we get a more efficient query by using builtin
+            // regular expression support and or'ing all extensions. We also
+            // escape the dot at the beginning of the extension.
+            // We will end up with a query that looks something like this:
+            // OR LOWER(name) ~ '(\.zip|\.rar|\.7zip|\.cab|\.jar|\.cpio|\.ar|\.gz|\.tgz|\.bz2)$')
+            query += "OR LOWER(name) ~ '(\\";
+            query += StringUtils.join(filter.getFilter().stream()
+                    .map(String::toLowerCase).collect(Collectors.toList()), "|\\");
```
millmanorama: This is pretty nit picky, but I would use collectors.joining instead of stringutils.join, since there is already a stream involved.

PR #2955, opened 2017-07-20
2689: Artifacts are Content

raman-bt:


Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -456,12 +457,13 @@ public boolean isSupported(Node node) {
...
-            try {
-                return content.getAllArtifactsCount() > 0;
-            } catch (TskException ex) {
-                logger.log(Level.WARNING, "Couldn't get count of BlackboardArtifacts for content", ex); //NON-NLS
+        for (Content content : node.getLookup().lookupAll(Content.class)) {
+            if ( (content != null)  && (!(content instanceof BlackboardArtifact)) ){
+                try {
+                    return content.getAllArtifactsCount() > 0;
+                } catch (TskException ex) {
+                    logger.log(Level.WARNING, "Couldn't get count of BlackboardArtifacts for content", ex); //NON-NLS
```
rcordovano: While we are in this code, please change this warning to SEVERE, thanks!

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -693,22 +695,25 @@ protected ViewUpdate doInBackground() {
...
-                logger.log(Level.WARNING, "Couldn't get artifacts", ex); //NON-NLS
-                return new ViewUpdate(getArtifactContents().size(), currentPage, ERROR_TEXT);
+            for (Content content : contents) {
+                if ( (content != null)  && (!(content instanceof BlackboardArtifact)) ) {
+                    // Get all of the blackboard artifacts associated with the content. These are what this
+                    // viewer displays.
+                    try {
+                        artifacts = content.getAllArtifacts();
+                    } catch (TskException ex) {
+                        logger.log(Level.WARNING, "Couldn't get artifacts", ex); //NON-NLS
```
rcordovano: While we are in this code, please change this warning to SEVERE, thanks!

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -124,8 +123,20 @@ public BlackboardArtifactNode(BlackboardArtifact artifact, String iconPath) {
...
+                if ( (content != null)  && (!(content instanceof BlackboardArtifact)) ){
+                    this.associated = content;
+                    break;
+                }
+            }
+        }
+        if (null == this.associated ) {
+            this.associated = this.getLookup().lookup(Content.class);
+        }
+
```
rcordovano: I am a bit confused by this code. I don't understand why we need to call lookupAll twice - won't the for loop just not execute if the collection returned is empty? Also, if we do not find associated via lookupAll(Content.class), why would we expect to find it via lookup(Content.class)? 

If we are going to have several lines of duplicated code here. it would be nice to have a littlr private helper method for this, called by both constructors.

raman-bt: I'm fixing this as part of "Show messages in Data source tree" work,  Your point about not needing to call lookupAll() twice is well taken. The way I had it, we were looping through the contents returned by lookupAll only if we knew there was more than 1.  Hence the second call to lookup.

The two constructors for BlackboardArtifactNode are nearly identical, one should simply call the other passing the assumed iconpath, that will eliminate the duplication.

rcordovano: Somds good.

rcordovano:
Closing thsi PR, because we are going to  pull iunto the art0fact-are-content branch first, then integrate into develop later.

PR #2962, opened 2017-07-25
remove nodecount threshold

millmanorama:


rcordovano:
Per @bcarrier, we will leacve this threshold in place until we have better verified that it is no longer needed. I am closing this PR for now, but please retain the branch for a future integration, thanks!

PR #2966, opened 2017-07-26
2672 more check on failed build

zhhl:


test/script/regression.py
```diff
@@ -1391,6 +1391,12 @@ def _generate_common_log(test_data):
                 except UnicodeDecodeError as e:
                     pass
                 log.close()
+            # some build errors are stored in antlog.txt not the autopsy logs
+            antlog = 'antlog.txt'
+            for ant_line in codecs.open(os.path.join(logs_path, os.pardir, antlog)):
+                ant_ignoreCase = ant_line.lower()
+                if ant_line.startswith("BUILD FAILED") or "fatal error" in ant_ignoreCase or "crashed" in ant_ignoreCase:
+                    common_log.write(antlog +": " +  ant_line)
```
rcordovano: I think we need to do this check immediately after the TestRunner._run_ant call and mark the test a failure because of the error; we don't need to add the ant log erors to the common autopsy log for diffing, because we already know something very bad has happened in the build and run process.

PR #2972, opened 2017-07-28
use extension column in query

millmanorama:
requires a case created/upgraded with https://github.com/sleuthkit/sleuthkit/pull/885


millmanorama:
closing this while I address @bcarrier's comments.

millmanorama:
reopening this as nothing has changed in autopsy only the corresponding TSK PR (#885)

PR #2986, opened 2017-08-03
Changing the status message based on text index size (2909)

eugene7646:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
```diff
@@ -211,6 +213,8 @@ public String getServiceName() {
         "SolrSearch.checkingForLatestIndex.msg=Looking for text index with latest Solr and schema version",
         "SolrSearch.indentifyingIndex.msg=Identifying text index to use",
         "SolrSearch.openCore.msg=Opening text index",
+        "SolrSearch.openLargeCore.msg=Opening text index. This may take several minutes.",
+        "SolrSearch.openGiantCore.msg=Opening text index. Text index for this case is really large and may take long time to load.",
```
millmanorama: I would say very large, instead of really large as I think the tone is better.

eugene7646: Done

PR #2989, opened 2017-08-03
2897 view file counts

dgrove727:


Core/src/org/sleuthkit/autopsy/datamodel/DeletedContent.java
```diff
@@ -427,8 +428,14 @@ static private String makeQuery(DeletedContent.DeletedContentFilter filter) {
                         logger.log(Level.SEVERE, "Unsupported filter type to get deleted content: {0}", filter); //NON-NLS
 
                 }
+                
+                if(UserPreferences.hideKnownFilesInViewsTree()) {
+                    query += " AND (known != " + TskData.FileKnown.KNOWN.getFileKnownValue() //NON-NLS
+                            + " OR known IS NULL)"; //NON-NLS
+                }
```
wschaeferB: Would I be correct in assuming Slack files are not possible in the deleted files tree and so their visibility doesn't need to be accounted for in query construction?

dgrove727: That's correct. It has a limited scope of types it looks at, and SLACK is not part of that scope.

PR #2995, opened 2017-08-09
2918 update script and gold

zhhl:


rcordovano:
Closing because we have worked out a different solution.

PR #2998, opened 2017-08-10
2914 new aid

dgrove727:


millmanorama:
Is this supposed to go into 2.8.3 or develop?  Also there are merge conflicts as it is now.

CentralRepository/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -540,18 +540,20 @@ public void addArtifact(EamArtifact eamArtifact) throws EamDbException {
         try {
```
rcordovano: This should be going into the develop branch.

PR #3004, opened 2017-08-15
Use counts of files + message type artifacts to determine node counts…

APriestman:
…/whether a node is a leaf node


Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeFilterChildren.java
```diff
@@ -120,7 +121,7 @@ private static boolean isLeafDirectory(DirectoryNode node) {
                         && !((Directory) c).getName().equals(".."))) {
                     ret = false;
                     break;
-                } else if (c.hasChildren()) {
+                } else if (0 < Case.getCurrentCase().getSleuthkitCase().getVisibleContentChildrenCount(c)) {
```
esaunders: Can c.hasVisibleChildren() be used here?

APriestman: Not easily I don't think - I also wanted to make that change, but hasVisibleChildren is on AbstractContentNode not Content

rcordovano: @APriestman and @esaunders , I was actually looking for this query to be made via a companion method to AbstractContentNode.getContentChildren, rather than adding a new, specialized API to SleuthkitCase.  This is for the UI, not for the data model, e.g., the method header in SleuthkitCase refers to lazy (node) loading, which has no meaning in the SleuthKit layer.  So I think we should call SleuthkitCase.executeQuery from a new AbstractContentNode.getVisibleContentChildrenCount method.

Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeFilterChildren.java
```diff
@@ -204,11 +205,11 @@ private Boolean visitDeep(AbstractAbstractFileNode<? extends AbstractFile> node)
                 if ((childContent instanceof AbstractFile) && ((AbstractFile) childContent).isDir()) {
                     return false;
                 } else {
-                    try {
-                        if (childContent.hasChildren()) {
+                    try{
+                        if(0 < Case.getCurrentCase().getSleuthkitCase().getVisibleContentChildrenCount(childContent)){
```
rcordovano: @APriestman and @esaunders , I was actually looking for this query to be made via a companion method to AbstractContentNode.getContentChildren, rather than adding a new, specialized API to SleuthkitCase.  This is for the UI, not for the data model, e.g., the method header in SleuthkitCase refers to lazy (node) loading, which has no meaning in the SleuthKit layer.  So I think we should call SleuthkitCase.executeQuery from a new AbstractContentNode.getVisibleContentChildrenCount method instead.

PR #3011, opened 2017-08-18
2935 changes obj_id to par_obj_id in Blackboard Artifacts table

wschaeferB:


wschaeferB:
Requires sleuthkit PR https://github.com/sleuthkit/sleuthkit/pull/898

PR #3013, opened 2017-08-22
Pass weak listener references to event publisher to allow blackboard …

esaunders:
…artifact nodes to be garbage collected.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
+++ b/Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
@@ -25,9 +25,11 @@ import java.beans.PropertyChangeListener;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -37,6 +39,7 @@ import org.openide.nodes.Children;
 import org.openide.nodes.Sheet;
 import org.openide.util.Lookup;
 import org.openide.util.NbBundle;
+import org.openide.util.WeakListeners;
 import org.openide.util.lookup.Lookups;
 import org.sleuthkit.autopsy.casemodule.Case;
 import org.sleuthkit.autopsy.casemodule.events.BlackBoardArtifactTagAddedEvent;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -155,11 +165,11 @@ public BlackboardArtifactNode(BlackboardArtifact artifact) {
         this.setName(Long.toString(artifact.getArtifactID()));
         this.setDisplayName();
         this.setIconBaseWithExtension(ExtractedContent.getIconFilePath(artifact.getArtifactTypeID())); //NON-NLS
-        Case.addPropertyChangeListener(pcl);
+        Case.addEventSubscriber(CASE_EVENTS_OF_INTEREST, WeakListeners.propertyChange(pcl, null));
```
millmanorama: Could we modify Case.addEventSubscriber to wrap the listenerers in the WeakListener "proxy" automatically?  Then this wouldn't crop up again/in other places.

millmanorama: It is alluded to that this was discused offline and rejected.

esaunders: Yeah, my initial attempt involved modifying LocalEventPublisher to wrap everything in a weak reference but that resulted in some events no longer being processed. Presumably the listener(s) in question were not being held onto by a strong reference elsewhere so they were getting garbage collected. I suspect using lambdas as a listener might have a role in this somehow but don't have the time at this point to continue investigation.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -64,6 +67,13 @@
 public class BlackboardArtifactNode extends DisplayableItemNode {
 
     private static final Logger LOGGER = Logger.getLogger(BlackboardArtifactNode.class.getName());
+    private static final Set<String> CASE_EVENTS_OF_INTEREST = new HashSet<>(Arrays.asList(new String[]{
```
esaunders: In the next story I'm working on I'm adding a new API to case that will take these events as a Set<Event> instead of Strings so the line referenced here will become:
    private static final Set<Case.Events> CASE_EVENTS_OF_INTEREST = EnumSet.of(Case.Events.BLACKBOARD_ARTIFACT_TAG_ADDED,
        Case.Events.BLACKBOARD_ARTIFACT_TAG_DELETED,
        Case.Events.CONTENT_TAG_ADDED,
        Case.Events.CONTENT_TAG_DELETED);

PR #3019, opened 2017-08-24
Fix for cuncurrent.ExecutionException when opening Solr core (2941)

eugene7646:
* Upgraded to latest apache common-io which fixes a race condition in FileUtils.sizeOfDirectory()
* Reverted back the Solr core open retries


CoreLibs/nbproject/project.properties
```diff
@@ -10,6 +10,7 @@ file.reference.commons-codec-1.10.jar=release/modules/ext/commons-codec-1.10.jar
 file.reference.commons-collections4-4.1.jar=release/modules/ext/commons-collections4-4.1.jar
 file.reference.commons-csv-1.4.jar=release/modules/ext/commons-csv-1.4.jar
 file.reference.commons-io-2.4.jar=release/modules/ext/commons-io-2.4.jar
```
esaunders: Can the 2.4 jar be removed from project.properties?

PR #3035, opened 2017-09-01
2943 artifacts as content cleanup

wschaeferB:


Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerThumbnail.java
```diff
@@ -384,7 +384,9 @@ public void setNode(Node givenNode) {
         }
         try {
             if (givenNode != null) {
-                tfn = (TableFilterNode) givenNode;
+                if (givenNode instanceof TableFilterNode){
+                    tfn = (TableFilterNode) givenNode;
+                }
```
rcordovano: I'm not sure how this relates to artifacts as content cleanup. All nodes that are pushed from the results view to the tree view are expected to be wrapped with TableFilterNode decorators, so a bad cast exception is probably a good thing to draw our attention to a programming error.  Did you encounter an error here?

Core/src/org/sleuthkit/autopsy/datamodel/ContentUtils.java
```diff
@@ -404,20 +404,22 @@ public Void visitDir(AbstractFile dir) {
...
-                    // If this is the source directory of an extract it
-                    // will have a progress and worker, and will keep track
-                    // of the progress bar's progress
-                    if (worker != null && worker.isCancelled()) {
-                        break;
+                    if (child instanceof AbstractFile){
+                        java.io.File childFile = getFsContentDest(child);
+                        ExtractFscContentVisitor<T, V> childVisitor
+                                = new ExtractFscContentVisitor<>(childFile, progress, worker, false);
+                        // If this is the source directory of an extract it
```
rcordovano: There are AbstractFiles that are not FsContent objects - LayoutFiles, LocalFiles, DerivedFiles, and SpecialDirectories (VirtualDirectory, LocalDirectory). The code here seems to be focused on FsContent. Is the AbstractFile type check correct here?

Core/src/org/sleuthkit/autopsy/directorytree/ExtractUnallocAction.java
```diff
@@ -170,7 +171,9 @@ public void approveSelection() {
         try {
             List<Content> unallocFiles = c.getChildren();
             if (null != unallocFiles && unallocFiles.isEmpty() == false) {
-                return unallocFiles.get(0).accept(uv); //Launching it on the root directory
+                if (unallocFiles.get(0) instanceof AbstractContent) {
+                    return unallocFiles.get(0).accept(uv); //Launching it on the root directory
+                }
```
rcordovano: I find the name "unallocFiles" for the List<Content> and the comment about launching on the root directory confusing here. We are not getting a collection of unallocated files, we are getting the children of an Image or Volume, and making the assumption that the first child is  a root directory to be searched for unallocated space files. Let's document and button this down better. We need a more accurate name for unallocFiles (Maybe just contentChildren?  I'm open to a better idea.)  Then are we sure that the first child of an Image or Volume will always be the "root directory"? Are there overrides of getChildren for Image and Volume that ensure this? It looks like AbstractFile has an isRoot method, is that applicable here? Or, looking at the UnallocVisitor, is the comment just wrong?

Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerThumbnail.java
wschaeferB: Sorry this change got accidentally included as it was part of me trouble shooting another bug before I wrote it up, regarding the thumbnail view in timeline mode throwing errors regarding being unable to cast to a TableFilterNode.

Core/src/org/sleuthkit/autopsy/datamodel/ContentUtils.java
wschaeferB: With it now possible for Directory.getChildren to return an artifact where as it previously only returned AbstractFiles it seemed like checking for instanceof AbstractFile was would maintain the current functionality.  However the existing and correct functionality are not necessarily the same thing, and checking for FsContent does seem like the the correct check given the context.

Core/src/org/sleuthkit/autopsy/directorytree/ExtractUnallocAction.java
wschaeferB: The getChildren methods for both Image and Volume rely on getChildrenInfo which performs a query whose results will ordered by tsk_objects.obj_id so the first result will be the child which was added to the database first.  Whether or not that will always be a "root directory" I need to investigate the creation order of the children a bit, but it seems reasonable that if there is a root directory being created as a child it will be the first child created.

Core/src/org/sleuthkit/autopsy/datamodel/ContentUtils.java
```diff
@@ -404,22 +407,24 @@ public Void visitDir(AbstractFile dir) {
...
                 for (Content child : dir.getChildren()) {
-                    java.io.File childFile = getFsContentDest(child);
-                    ExtractFscContentVisitor<T, V> childVisitor
-                            = new ExtractFscContentVisitor<>(childFile, progress, worker, false);
-                    // If this is the source directory of an extract it
-                    // will have a progress and worker, and will keep track
-                    // of the progress bar's progress
-                    if (worker != null && worker.isCancelled()) {
-                        break;
+                    if (child instanceof FsContent){
```
rcordovano: You were right the first time, and I misled you. This instanceof check is not quite correct. The visitor is supposed to be able to "extract"  LayoutFile,  DerrivedFile, LocalFile, LocalFile and recurse through VirtualDirectory and LocalDirectory. All of these classes extend AbstractFile, not FsContent.

PR #3042, opened 2017-09-06
Create a central repo case entry if one does not exist when the user …

APriestman:
…opens the properties panel.


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -254,6 +255,31 @@ public void newCase(EamCase eamCase) throws EamDbException {
         }
     }
 
+     /**
+     * Creates new Case in the database from the currently open case
+     */
+    @Override    
+    public EamCase newCaseFromCurrentCase() throws EamDbException{
+        Case curCase = Case.getCurrentCase();
```
rcordovano: I think we should keep getting the current case in the case details dialog, and use whatever API we already have for adding a case to the CR instead. By doing so, we would keep the lower level CR services more flexible and less likely to need modification, e.g., when/if we move to having multiple cases open.

PR #3045, opened 2017-09-06
2831 event listener cr disables

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
```diff
@@ -52,7 +52,7 @@
     private static final Logger LOGGER = Logger.getLogger(EamArtifact.class.getName());
 
     final Collection<String> addedCeArtifactTrackerSet = new LinkedHashSet<>();
-
+    private static long ingestJobCounter = 0;
```
APriestman: Should ingestJobCounter be volatile?

wschaeferB: I was under the impression that since all access to it was in synchronized methods, so that they can be safely accessed from different threads, that making it volatile would not accomplish any additional safeguarding.  But perhaps I am missing something.

APriestman: Yeah I think you're right. Just wanted to be sure.

Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
```diff
@@ -195,6 +197,7 @@ public void shutDown() {
     })
     @Override
     public void startUp(IngestJobContext context) throws IngestModuleException {
+        IngestEventsListener.enableCentralRepositoryModule();
```
rcordovano: I believe that this method should be called at the very end of startUp, not the beginning. There are a lot of places in the code that follows where an IngestModuleException is thrown, meaning that the module will not run and the ingest job will not be started.

Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
```diff
@@ -155,6 +156,7 @@ public ProcessResult process(AbstractFile af) {
 
     @Override
     public void shutDown() {
+        IngestEventsListener.disableCentralRepositoryModule();
```
rcordovano: I believe that this method should be called after the  is enabled check, since the module may never have run and corresponding counter increment may never have happened. 

Note that this code in the startUp and shutDown methods has been modified by both Ann and myself, so this code no longer auto merges. You will need to merge in the latest code in develop.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
rcordovano: This not really an ingest job counter, is it? I think it would be less confusing for future maintainers if we called it an ingest module instance count instead.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
```diff
@@ -72,65 +72,104 @@ public void uninstallListeners() {
...
+    /**
+     * Wether or not the Central Repository Module is enabled for any of the
+     * currently running ingest jobs.
+     *
+     * @return boolean True for Central Repo enabled, False for disabled
+     */
+    private synchronized static boolean isCentralRepositoryModuleEnabled() {
+        return ingestJobCounter > 0;
+    }
+
```
rcordovano: As above, I think it would be more clear for future maintainers if we named these method to reflect that this is a counter for ingest module instances, not ingest jobs.

Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
wschaeferB: I was assuming shutdown would still be called if startup threw an exception but based on your comment I am gathering that is incorrect, and I will adjust it to reflect that since my placement of the calls as the first thing in both methods was intended to ensure both were always executed when the module was selected.

wschaeferB: I tested and shutdown is still called if startup threw an exception, so in order to keep the count accurate if I move the method calls to the end of startup and shutdown I need to first ensure that shutdown will also detect the same issues that startup detects prior to the method call.

wschaeferB: I feel like given the current error try catch structure of shutDown and startUp that the farthest down I can safely move the call in startUp is Line 236  of startUp after the dbManager instance is gotten but before the filesType variable is set.   With the call in startUp at that location the call in shutDown would need to occur after line 170.

PR #3068, opened 2017-09-12
Hold onto a reference to the weak listener so that it can be successf…

esaunders:
…ully removed. Use finalize() to take care of automatic listener removal when the node is garbage collected.


Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -131,6 +147,14 @@ private void removeListeners() {
         }
     };
 
+    // We pass a weak reference wrapper around the listener to the event publisher.
+    // This allows Netbeans to delete the node when the user navigates to another
+    // part of the tree (previously, nodes were not being deleted because the event
```
millmanorama: I would use a javadoc comment here.  Also weakPcl should be further up to avoid codacy complaining,

PR #3069, opened 2017-09-12
Known bad can be automatically added when setting a tag to BAD

APriestman:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1150,6 +1153,41 @@ public void setArtifactInstanceKnownStatus(EamArtifact eamArtifact, TskData.File
...
+            for(ContentTag contentTag:fileTags){
+                final EamArtifact eamArtifact = EamArtifactUtil.getEamArtifactFromContent(contentTag.getContent(), 
+                            TskData.FileKnown.BAD, "");
+                setArtifactInstanceKnownStatus(eamArtifact, TskData.FileKnown.BAD);
+            }
+        } catch (TskCoreException ex){
+            throw new EamDbException("Error updating artifacts", ex);
+        }
+        
+    }
```
rcordovano: I don't think this is the best place to put this method. My thinking is that adding methods to SleuthkitCase for lots of "special case" options is what has lead to a very large and therefore difficult to use SleuthkitCase API.  I would prefer to see a low-level service class such as the CR database provide primitive operations that can then be combined in various ways by clients to get the desired result. In my opinion, that this is not a primitive operation on the CR database is evident in the call to SleuthkitCase, which couples the CR database implementation to the case database implementation. I propose that we put this code in the ManageTagsDialog class instead. Then it will transfer nicely to the Tags options panel, which will then act as a client of the low level case database and CR database classes, linking them together to get create an operation specific to the UI.

rcordovano:
Please re-target the PR to the release-4.5.0 branch when the review comments a re resolved, thanks!

PR #3065, opened 2017-09-12
2834 manage tags panel info

wschaeferB:

PR #3072, opened 2017-09-13
2834 manage tags panel info

wschaeferB:
re-targeting of PR #3065 still pending review when created


rcordovano:
I'd like us to do a few very minor tweaks to the actual text as follows:

- On the options panel, let's not capitalize "central repository." 
- In the first label on the options panel, let's change "The central repository..." to "A central repository..." to be consistent with "Use a central repository..."
- To be consistent with the rest of the UI, please replace the references to "forensic artifacts" and "artifacts" with "results."
- For the *Manage Tags* dialog, please reword the instruction as follows: "Select the tags that cause files and results to be recorded in the central repository."

Thanks!

PR #3077, opened 2017-09-15
2978 db version compatibility

millmanorama:
depends on https://github.com/sleuthkit/sleuthkit/pull/961


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1787,7 +1791,15 @@ private void createCaseData(ProgressIndicator progressIndicator) throws CaseActi
...
-        "Case.exceptionMessage.couldNotOpenCaseDatabase=Failed to open case database."
+        "Case.exceptionMessage.couldNotOpenCaseDatabase=Failed to open case database.",
+        "# {0} - unsupported  version",
+        "# {1} - app name",
+        "# {2} - maximum supported major scheme version",
+        "Case.unsupportedSchemaVersionMessage=Unsupported DB schema version: {0}\n"
+        + "The highest supported DB schema version for this release of {1} is {2}.X",
+        "Case.databaseConnectionInfo.error.msg=Error accessing database server connection info. See Tools -> Options -> Multi-user.",
+        "Case.open.exception.multiUserCaseNotEnabled=Cannot open a multi-user case if multi-user cases are not enabled. "
+        + "See Tools, Options, Multi-user."
```
rcordovano: I we are going to use arrows instead of commas to show a sequence of menu selections, we should be consistent. 

Also, the strings here are ever so slightly out-of-date: the options panel is now "Multi-User", not "Multi-user."

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1798,16 +1810,19 @@ private void openCaseData(ProgressIndicator progressIndicator) throws CaseAction
...
-                        "Case.open.exception.multiUserCaseNotEnabled"));
+                throw new CaseActionException(Case_open_exception_multiUserCaseNotEnabled());
             }
+        } catch (TskUnsupportedSchemaVersionException ex) {
+            throw new CaseActionException(
+                    Bundle.Case_unsupportedSchemaVersionMessage(
+                            ex.getUnsupportedVersion().toString(),
+                            Version.getName(),
+                            ex.getCurrentVersion().getMajor()),
+                    ex);
```
rcordovano: Please see the comment in the corresponding sleuthkit pull request regarding the design of the TskUnsupportedSchemaVersion class, as it would affect the construction here of the displayable "user-friendly" message to satisfy the current convention of the usage of CaseActionException. Note that as soon as time permits, I want to do away with that convention. We would be much better served by a hierarchy of CaseActionException subtypes that allow clients to assemble meaningful error messages based on exception type, and could make such a change without breaking backwards compatibility of the Case public API.

PR #3083, opened 2017-09-18
2740 file system details dialog

wschaeferB:


Core/src/org/sleuthkit/autopsy/datamodel/FileNode.java
```diff
@@ -147,6 +148,7 @@ private void setIcon(AbstractFile file) {
     public Action[] getActions(boolean context) {
         List<Action> actionsList = new ArrayList<>();
         actionsList.addAll(Arrays.asList(super.getActions(true)));
+        actionsList.add(new FileSystemDetailsAction(content));
         if (!this.getDirectoryBrowseMode()) {
```
millmanorama: should this move into some common place such as AbstractFsContentNode, which is the supe type of all three of these node classes?

wschaeferB: I thought about that but currently we don't have an implementation of the getActions method there so we would need to do the following and I wasn't confident creating yet another array list of actions was preferable to having one like repeated three times.  If having the additional list created is indeed preferable then I will make those changes.
 
List<Action> actionsList = new ArrayList<>();
actionsList.addAll(Arrays.asList(super.getActions(true)));
actionsList.add(new FileSystemDetailsAction(content));
return actionsList;

millmanorama:
There is also the question of order of actions...

wschaeferB:
Assuming the order I first put them in with the File System Details inserted after Properties and before  View File in Directory is the place we want it then it will still be where we want it with the adding of the action pushed up a level.   I have gone ahead and made that change, there are probably additional actions we could move up also with a little work to reduce code duplication, but since all the ViewContextActions are using node specific bundle messages currently I have left that for now.

PR #3091, opened 2017-09-21
2963 validate ccn format

millmanorama:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/RegexQuery.java
```diff
@@ -273,44 +273,44 @@ public QueryResults performQuery() throws NoOpenCoreException {
...
+                        case TSK_EMAIL:
+                            /*
+                             * Reduce false positives by eliminating email
+                             * address hits that are either too short or are not
+                             * for valid top level domains.
+                             */
+                            if (hit.length() >= MIN_EMAIL_ADDR_LENGTH
+                                    && DomainValidator.getInstance(true).isValidTld(hit.substring(hit.lastIndexOf('.')))) {
+                                addHit(content, snippet, hitMatcher, hit, hits, docId);
+                            }
```
rcordovano: @millmanorama , in conversation you wondered whether the switch is better than the if-else blocks. FWIW, I like the switch. I think it is an improvement to code clarity.

PR #3096, opened 2017-09-22
Added the 'Prioritize Case' button.

dgrove727:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/Bundle.properties
```diff
@@ -240,8 +240,10 @@ AutoIngestCasePanel.bnOpen.text=&Open
 AutoIngestCasePanel.bnShowLog.toolTipText=Display case log file for selected case
 AutoIngestCasePanel.bnShowLog.text=&Show Log
 AutoIngestCasePanel.rbGroupLabel.text=Show cases accessed in the last 10:
-AutoIngestDashboard.prioritizeButton.toolTipText=Prioritizes the selected job
-AutoIngestDashboard.prioritizeButton.text=&Prioritize
 AutoIngestDashboard.refreshButton.toolTipText=Refresh displayed tables
 AutoIngestDashboard.refreshButton.text=&Refresh
 AutoIngestDashboard.jButton1.text=jButton1
+AutoIngestDashboard.prioritizeJobButton.toolTipText=Prioritizes the selected job
```
esaunders: Is there a reason why we didn't use the same tooltip text that is used in AID 1.0?

esaunders:
Can you resubmit this with the latest release-4.5.0 code? Thanks.

PR #3102, opened 2017-09-25
3061 changed runIngestmodules from callableSystemAction 

wschaeferB:
to abstractAction to avoid exeption

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/ingest/runIngestModuleWizard/RunIngestModulesAction.java
+++ b/Core/src/org/sleuthkit/autopsy/ingest/runIngestModuleWizard/RunIngestModulesAction.java
@@ -23,13 +23,12 @@ import java.awt.event.ActionEvent;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/ingest/runIngestModuleWizard/RunIngestModulesAction.java
```diff
@@ -103,35 +117,9 @@ public void actionPerformed(ActionEvent e) {
...
-    @Override
-    public String getName() {
-        return Bundle.RunIngestModulesAction_name();
-    }
 
     @Override
-    public HelpCtx getHelpCtx() {
-        return HelpCtx.DEFAULT_HELP;
+    public Object clone() throws CloneNotSupportedException {
+        return super.clone(); //To change body of generated methods, choose Tools | Templates.
```
rcordovano: Please change this implementation to throw CloneNotSupportedException. This implementation would result in sharing references to the  List field and the items in the list among clones. A deep copy approach  like that used by the superclass would work, but it's not worth the effort. Thanks!

PR #3123, opened 2017-09-28
3087 cr schema changes

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationDataSource.java
```diff
@@ -36,14 +36,17 @@
...
     private final String deviceID;  //< Unique to its associated case (not necessarily globally unique)
     private final String name;
+    private final int caseID;
 
 
     CorrelationDataSource(int dataSourceId,
             String deviceID,
-            String name) {
+            String name,
+            int caseID) {
```
bcarrier: We should specify if this is the UUID or the database ID. If UUID, then rename it.  Otherwise, just make it clear in the docs.

Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
```diff
@@ -226,53 +218,64 @@ public void startUp(IngestJobContext context) throws IngestModuleException {
...
         try {
             filesType = dbManager.getCorrelationTypeById(CorrelationAttribute.FILES_TYPE_ID);
         } catch (EamDbException ex) {
             LOGGER.log(Level.SEVERE, "Error getting correlation type FILES in ingest module start up.", ex); // NON-NLS
             throw new IngestModuleException("Error getting correlation type FILES in ingest module start up.", ex); // NON-NLS
         }
+        // ensure we have this case defined in the EAM DB
+        CorrelationCase existingCase;
+        Case curCase = Case.getCurrentCase();
+        CorrelationCase curCeCase = new CorrelationCase(
```
bcarrier: Seems like we could be doing this only if the case doesn't already exist. i.e. first do the check based on Case.getCurrentCase() on line 254 and then if it doesn't exist, then create one based on the current case.  We could probably add some locking around that too to reduce errors if there are multiple ingest threads that do it at the same time.  Though, it really should have been created when the case was opened.

wschaeferB: Should optimizing this pre-existing code be part of this pull request? (it showed up as changed because I needed to re-order elements of the startUp method.  The reason it may not of been created when the case was opened would be if the Central Repository wasn't enabled when the case was opened.

rcordovano: For future reference, this code is in the startUp method which is called sequentially for each ingest module instance for an ingest job by a single-threaded executor in either IngestManager or AutoIngestManager, so no locking is required.

We should do as Brian suggests, and I'm fine with doing it now. However, note that the constructor for CorrelationCase that takes an integer id as the first argument should not be called by client code; that constructor should be reserved for use by AbstractSQLEamDb. I know this is existing code and not very important, but we should still fix this while we are in here.

PR #3129, opened 2017-09-28
Cancel kws between chunks if ingest is cancelled 2888

eugene7646:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Ingester.java
```diff
@@ -176,13 +180,15 @@ void indexMetaDataOnly(BlackboardArtifact artifact) throws IngesterException {
...
-            indexChunk(null, sourceName, fields);
+            if (!context.fileIngestIsCancelled()) {
+                //after all chunks, index just the meta data, including the  numChunks, of the parent file
+                fields.put(Server.Schema.NUM_CHUNKS.toString(), Integer.toString(numChunks));
+                //reset id field to base document id
+                fields.put(Server.Schema.ID.toString(), Long.toString(sourceID));
+                //"parent" docs don't have chunk_size
+                fields.remove(Server.Schema.CHUNK_SIZE.toString());
+                indexChunk(null, sourceName, fields);
+            }
```
rcordovano: Shouldn't the finally block be returning false if ingest was cancelled, as in the loop? 

Looking at the actual use of this method, I notice that it isn't really designed to handle ingest cancellation:
- The call to this method in SolrSearchService.indexArtifact ignores the return value, so its sole caller,  Blackboard.indexArtifact does too, and the latter is called in all of the "core" ingest modules. This works out okay in the end,  but it exposes a lot of places where modules could be much more proactive about checking for ingest cancellation.
- In KeywordSearchIngestModule.indexFile, a return value of false from this method causes the module to fall back to string extraction. However, there is an ingest job cancellation check in KeywordSearchIngestModule.extractStringsAndIndex.
- In  KeywordSearchIngestModule.extractStringsAndIndex, the return value of false is passed through, and the three calls to extractStringsAndIndex in KeywordSearchIngestModule.indexFile return quickly, and there will be a quick return from  KeywordSearchIngestModule.process, which will allow an ingest job cancellation check by FileIngestPipeline.process.

For the release branch, let's fix the return value. I will do that, add a story to JIRA, and put a refernce to it in place of JM/s TODO.

PR #3137, opened 2017-10-06
Replaced folder scan with coordination service call.

dgrove727:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestCaseManager.java
```diff
@@ -49,21 +59,81 @@ synchronized static AutoIngestCaseManager getInstance() {
...
+    /**
+     * Initialize the coordination service.
+     * 
+     * @throws org.sleuthkit.autopsy.experimental.autoingest.AutoIngestCaseManager.AutoIngestCaseManagerException 
+     */
+    private void init() throws AutoIngestCaseManager.AutoIngestCaseManagerException {
+        try {
+            coordinationService = CoordinationService.getInstance();
+        } catch (CoordinationServiceException ex) {
+            throw new AutoIngestCaseManager.AutoIngestCaseManagerException(ex.getMessage(), ex);
```
rcordovano: The AutoIngestCaseManager should add some context to the Coordination service message, the message of which will already be logged in the stack trace, since the Autopsy Logger recurses through nested exceptions. Please replace the ex.getMessage with a suitable message created via an @Messages annotation, e.g., "Failed to get coordination service." The client can then add additional context when it logs the exception.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestCaseManager.java
```diff
@@ -49,21 +59,81 @@ synchronized static AutoIngestCaseManager getInstance() {
...
-        for (Path caseFolderPath : caseFolders) {
-            cases.add(new AutoIngestCase(caseFolderPath));
+        List<Path> casePathList;
+        try {
+            casePathList = getCasePaths();
+            for (Path casePath : casePathList) {
+                cases.add(new AutoIngestCase(casePath));
+            }
+        } catch (AutoIngestCaseManagerException ex) {
+            String errorMessage = NbBundle.getMessage(AutoIngestCaseManager.class, "AutoIngestCaseManager.CasePathsError");
```
rcordovano: I don't see any AutoIngestCaseManager messages in the Bundle.properties file. However, per our coding standard, we don't internationalize/localize exception messages, so just use some suitable English text here, given that we will not be displaying the message for the user (see next comment), thanks!

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestCaseManager.java
```diff
@@ -49,21 +59,81 @@ synchronized static AutoIngestCaseManager getInstance() {
...
+            for (String node : nodeList) {
+                if(node.indexOf('\\') >= 0 || node.indexOf('/') >= 0) {
+                    /*
+                     * This is not a case name lock (name specifies a path).
+                     */
+                    String nodeUpperCase = node.toUpperCase();
+                    if(!nodeUpperCase.endsWith("_RESOURCES") && !nodeUpperCase.endsWith("AUTO_INGEST_LOG.TXT")) {
+                        /*
+                         * This is not a case resource lock. Collect the path.
+                         */
```
rcordovano: For completeness, the comment should probably also mention that it is not a case auto ingest log lock node, either.

PR #3142, opened 2017-10-16
2774 central repo details gui

wschaeferB:


wschaeferB:
Currently it is hanging for a bout a minute after Autopsy is closed before finalizing the closing,  I will resubmit it when I have eliminated this issue.

wschaeferB:
Issue appears to be an existing one when a case is open and CR enabled.  Made a jira story to reflect it.

Core/src/org/sleuthkit/autopsy/casemodule/CaseDetails.java
```diff
@@ -0,0 +1,94 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.casemodule;
+
+/**
+ * Wrapper to contain the modifiable details of a case, such as case display
+ * name, case number, and examiner related fields.
+ */
+final class CaseDetails {
```
rcordovano: We should make this class public (see comment on new Case.createAsCurrentCase method overload).

Core/src/org/sleuthkit/autopsy/casemodule/CaseDetails.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private final String caseDisplayName;
+    private final String caseNumber;
+    private final String examinerName;
+    private final String examinerPhone;
+    private final String examinerEmail;
+    private final String caseNotes;
+
+    CaseDetails(String displayName, String number, String exName, String exPhone, String exEmail, String notes) {
```
rcordovano: All of the fields in a CaseDetails object except the case display name are optional. For clarity, I would like to see us model this by providing a constructor that takes only a display name argument, setting all other fields to the default values of the empty string.

We can then use this in AutoIngestManager to call the new Case.createAsCurrentCase method overload (see prior comments) cleanly and compactly; this will also be nice for the large number of Case.createAsCurrentCase calls we will soon be seeing in unit test code.

Core/src/org/sleuthkit/autopsy/casemodule/NewCaseVisualPanel2.java
```diff
@@ -61,75 +67,61 @@ public String getName() {
...
+    }
+
+    void setOrganization(String organizationName) {
+        propertiesPanel.setCurrentlySelectedOrganization(organizationName);
     }
 
-    String getExaminer() {
-        return examinerTextField.getText();
+    void refreshCentralRepoFields() {
+        propertiesPanel.setUpCentralRepoFields();
```
rcordovano: I think this name (across both classes) is a bit misleading now, given that we are storing several of the values gathered in the case metadata file as well as the CR. Perhaps refresh/setUpCaseDetailsFields?

Core/src/org/sleuthkit/autopsy/casemodule/NewCaseWizardAction.java
```diff
@@ -75,11 +79,36 @@ private void runNewCaseWizard() {
...
+                                    for (EamOrganization org : dbManager.getOrganizations()) {
+                                        if (org.getName().equals(organizationName)) {
+                                            cRCase.setOrg(org);
+                                            dbManager.updateCase(cRCase);
+                                        }
+                                    }
+                                }
+                            }
+                        } catch (EamDbException ex) {
+
```
rcordovano: We should only rarely swallow exceptions.  Here, we should allow the exception to propagate and be handled in the catch block for the get() call in the done() method, with the reasoning that if the user has decided to use the central repository, they should be notified if something goes wrong creating the case in the CR.

Core/src/org/sleuthkit/autopsy/casemodule/NewCaseWizardPanel2.java
```diff
@@ -46,7 +56,10 @@
     public NewCaseVisualPanel2 getComponent() {
         if (component == null) {
             component = new NewCaseVisualPanel2();
+        } else {
+            component.refreshCentralRepoFields();
```
rcordovano: See previous comment about this method name.

Core/src/org/sleuthkit/autopsy/casemodule/CaseMetadata.java
```diff
@@ -452,13 +463,28 @@ private void readFromFile() throws CaseMetadataException {
...
+            String examinerEmail;
+            String caseNotes;
+            try {
+                examinerPhone = getElementTextContent(caseElement, EXAMINER_ELEMENT_PHONE, false);
+                examinerEmail = getElementTextContent(caseElement, EXAMINER_ELEMENT_EMAIL, false);
+                caseNotes = getElementTextContent(caseElement, CASE_ELEMENT_NOTES, false);
+            } catch (CaseMetadataException ex) {
+                examinerPhone = "";  //case had metadata file written before additional examiner details were included 
+                examinerEmail = "";
+                caseNotes = "";
```
rcordovano: We should do a version number conditional here, as in the existing code, keeping this code consistent, and reserving exceptions for error conditions.

PR #3144, opened 2017-10-18
3133 remove examiner mode

dgrove727:


Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCaseManager.java
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param caseMetadataFilePath Path to the case metadata file.
+     *
+     * @throws CaseActionException
+     */
+    synchronized void openCase(Path caseMetadataFilePath) throws CaseActionException {
+        /*
+         * Open the case.
+         */
+        Case.openAsCurrentCase(caseMetadataFilePath.toString());
```
rcordovano: When a case is opened via this path, it does not appear as a **Recent Cases** menu item.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/PathUtils.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Collections;
 import java.util.List;
 import org.sleuthkit.autopsy.casemodule.CaseMetadata;
+import org.sleuthkit.autopsy.coreutils.TimeStampUtils;
```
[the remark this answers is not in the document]
dgrove727: It won't compile without it. It doesn't share the same package or NBM. I believe TimeStampUtils was one of those classes that was moved out of Experimental.

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
```diff
@@ -110,34 +121,60 @@ Date getLastAccessedDate() {
...
+     */
+    static CaseMetadata getCaseMetadataFromCaseDirectoryPath(Path caseDirectoryPath) throws CaseMetadata.CaseMetadataException {
+        CaseMetadata caseMetadata = null;
+        
+        File directory = new File(caseDirectoryPath.toString());
+        if (directory.isDirectory()) {
+            String fileNamePrefix = directory.getName();
+            if (TimeStampUtils.endsWithTimeStamp(fileNamePrefix)) {
+                fileNamePrefix = fileNamePrefix.substring(0, fileNamePrefix.length() - TimeStampUtils.getTimeStampLength());
+            }
```
dgrove727: Although I revised this code, I'm rethinking it. There's no guarantee that if the path has a time stamp, the AUT file won't. A full directory scan for an AUT file would avert this odd scenario, but it may be slower too.

dgrove727: New commit addresses this.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/PathUtils.java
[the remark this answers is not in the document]
rcordovano: Got it, thanks!

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
```diff
@@ -46,10 +48,19 @@
...
+        } catch (CaseMetadata.CaseMetadataException ex) {
+            logger.log(Level.SEVERE, String.format("Error reading the case metadata for %s.", caseDirectoryPath), ex);
+            throw ex;
+        }
+        
         this.caseDirectoryPath = caseDirectoryPath;
-        caseName = PathUtils.caseNameFromCaseDirectoryPath(caseDirectoryPath);
-        metadataFilePath = caseDirectoryPath.resolve(caseName + CaseMetadata.getFileExtension());
+        caseName = caseMetadata.getCaseDisplayName();
+        metadataFilePath = caseDirectoryPath.resolve(caseMetadata.getCaseName() + CaseMetadata.getFileExtension());
```
rcordovano: CaseMetadata has a getFilePath method, you can just use that here.

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
```diff
@@ -110,34 +121,81 @@ Date getLastAccessedDate() {
...
+            
+            /*
+             * If the AUT file has been found, grab the case metadata.
+             */
+            if(autFile.isFile()) {
+                caseMetadata = new CaseMetadata(Paths.get(autFile.getAbsolutePath()));
+            }
+        }
+        
+        return caseMetadata;
```
rcordovano: This method can return null (not documented), and your use of it does not check the return value.  Personally, I would define a MultiUserCaseException class, which keeps the use of CaseMetadata encapsulated instead of coupling all clients of this class to CaseMetadata  via CaseMetadataException. Then, if this method throws, the error can just propagate and there is no use of the error prone use of a special return value (null) to indicate an error condition. After all, if the case metadata file cannot be found, the case cannot be opened.

Also, a method like this should be private. Private access should be the default for everything.

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
rcordovano: CaseMetadata has a getFilePath method, you can just use that here.

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
rcordovano: Have you profiled the isDirectory call, plus the regular expression pattern matching, plus the string concatenation, plus File object construction, plus the File.isFile calls to see if there is a time savings? This seems like premature optimization at the cost of greater code complexity, anmd the complexity of the code has already led to a possible NPE bug (See below). Let's just keep it simple and do the file listing. Then you could even consider inlining this code. If we are having a performance problem, then we can profile and

Core/src/org/sleuthkit/autopsy/casemodule/CaseMetadata.java
```diff
@@ -172,6 +172,17 @@ Path getFilePath() {
...
+    /**
+     * Gets the full case directory path.
+     * 
+     * @return The case directory path.
+     */
+    public String getCaseDirectoryPath() {
+        String fileName = metadataFilePath.getFileName().toString();
+        String caseDirectoryPath = metadataFilePath.toString();
+        return caseDirectoryPath.substring(0, caseDirectoryPath.lastIndexOf(fileName));
+    }
```
rcordovano: This class already has a getCaseDirectory method. Why does it need another one?

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCase.java
rcordovano: I am confused. Why is this code still doing anything other than a simple file listing to find the case metadata file?

Core/src/org/sleuthkit/autopsy/casemodule/MultiUserCasesPanel.java
```diff
@@ -320,103 +251,77 @@ private void setSelectedCase(Path path) {
...
     private void setButtons() {
-        boolean enabled = casesTable.getSelectedRow() >= 0 && casesTable.getSelectedRow() < casesTable.getRowCount();
-        bnOpen.setEnabled(enabled);
-        bnShowLog.setEnabled(enabled);
+        boolean openEnabled = casesTable.getSelectedRow() >= 0 && casesTable.getSelectedRow() < casesTable.getRowCount();
+        bnOpen.setEnabled(openEnabled);
+
+        Path pathToLog = getSelectedCaseLogFilePath();
+        boolean showLogEnabled = openEnabled && pathToLog != null && pathToLog.toFile().exists();
+        bnShowLog.setEnabled(showLogEnabled);
```
rcordovano: This code does not work. The paths are all to the case metadata file, with the log file name tacked on. It appears that this code was not tested?

PR #3145, opened 2017-10-18
3080 Moved all actions as a result of CaseEventLister to background

wschaeferB:


Core/src/org/sleuthkit/autopsy/coreutils/ExecUtil.java
```diff
@@ -213,6 +214,40 @@ public static void killProcess(Process process) {
...
 
     /**
+     * Shuts down a task executor service, waiting until all tasks are
+     * terminated. The current policy is to wait for the tasks to finish so that
+     * the case for which the executor is running can be left in a consistent
+     * state.
+     *
+     * @param executor The executor.
+     */
+    public static void shutDownTaskExecutor(ExecutorService executor) {
```
rcordovano: @wschaeferB and @APriestman, good job factoring this code out of the Case class for general use! However, ExecUtil is documented as "Executes a command line using an operating system process with a configurable timeout and pluggable logic to kill or continue the process on timeout."  We should make a new utility class for this code, ThreadUtils or something. We could perhaps even add another method right away for the intuitive ThreadFactoryBuilder stuff we do all over the place, what do you think?

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -53,9 +57,19 @@
  * accordingly
  */
 @Messages({"caseeventlistener.evidencetag=Evidence"})
-public class CaseEventListener implements PropertyChangeListener {
+final class CaseEventListener implements PropertyChangeListener {
```
rcordovano: Nice!

Core/src/org/sleuthkit/autopsy/coreutils/ExecUtil.java
```diff
@@ -213,6 +214,40 @@ public static void killProcess(Process process) {
...
+                 * experienced an error condition when opening case resources
+                 * that left the service blocked uninterruptibly on a socket
+                 * read. This eventually led to a mysterious "freeze" as the
+                 * user-cancelled service task continued to run holdiong a lock
+                 * that a UI thread soon tried to acquire. Thus it has been
+                 * deemed better to make the "freeze" happen in a more
+                 * informative way, i.e., with the progress indicator for the
+                 * unfinished task on the screen, if a similar error condition
+                 * arises again.
+                 */
```
rcordovano: This comment seems pretty specific to the Case class use case. now that we are "going public", I think it would be a good idea to remove this comment and change the method documentation to leave it at "Shuts down a task executor service, waiting until all tasks are terminated."

PR #3146, opened 2017-10-19
3116: Update the credit card account HMTL report title from Accounts:…

zhhl:
… CREDIT_CARD to Accounts: Credit Card

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
+++ b/Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
@@ -176,8 +176,11 @@ class TableReportGenerator {
                      * compund name by appending a ":" and the account type.
                      */
                     String accountDisplayname = accountType;
-                    if (accountType != null && accountType.equals(Account.Type.CREDIT_CARD.name())) {
-                        accountDisplayname = Account.Type.CREDIT_CARD.getDisplayName();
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/report/TableReportGenerator.java
```diff
@@ -174,7 +175,11 @@ private void makeBlackboardArtifactTables() {
                      * does not require a artifact name, so we make a synthetic
                      * compund name by appending a ":" and the account type.
                      */
-                    final String compundDataTypeName = BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT.getDisplayName() + ": " + accountType;
+                    String accountDisplayname = accountType;
+                    if (accountType != null && accountType.equals(Account.Type.CREDIT_CARD.name())) {
+                        accountDisplayname = Account.Type.CREDIT_CARD.getDisplayName();
+                    }
+                    final String compundDataTypeName = BlackboardArtifact.ARTIFACT_TYPE.TSK_ACCOUNT.getDisplayName() + ": " + accountDisplayname;
```
rcordovano: I see how this works, but this is not extensible, which I think you may have already noticed yourself. I think a better generic solution is to map the accountType string to the correct enum member using Account.Type.valueOf, then call getDisplayName on the result.

PR #3151, opened 2017-10-24
Build out cvt gui

millmanorama:
initial PR for CVT GUI:
TopComponent with button/tool menu item to open it.
Table that shows accounts.
no filters, lots of missing things and rough edges.


Core/src/org/sleuthkit/autopsy/communications/OpenCVTAction.java
```diff
@@ -0,0 +1,107 @@
...
+@Messages("CTL_OpenCVTAction=Visualize Communications")
+public final class OpenCVTAction extends CallableSystemAction implements Presenter.Toolbar {
+
+    private static final long serialVersionUID = 1L;
+
+    private final JButton toolbarButton = new JButton(getName(),
+            new ImageIcon(getClass().getResource("images/email_link.png"))); //NON-NLS
+
+    public OpenCVTAction() {
+        toolbarButton.addActionListener(actionEvent -> performAction());
```
rcordovano: I think it would be better to spell out the name more rather use an acronym. Perhaps OpenCommVisualizationToolAction.

Core/src/org/sleuthkit/autopsy/communications/AccountsNodeFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.List;
+import org.openide.nodes.ChildFactory;
+import org.openide.nodes.Node;
+import org.sleuthkit.datamodel.Account;
+
+
+
+class AccountsNodeFactory extends ChildFactory<Account> {
+
```
rcordovano: It looks like we are plowing a green field in terms of the view that will display the AccountNodes, am I right? If so, I still think we should go with lazy child node creation here.

millmanorama:
I addressed both your comments richard in the latest commit

PR #3154, opened 2017-10-24
865 build out filters host container

millmanorama:
I built out the filters host panel, and populated the filter widgets, and plugged in the Filters API as it is so far..  I also polished various parts of the UI including adding icons for various things, and making the table size its columns to the data


millmanorama:
builds on PR #3151

millmanorama:
Is there a reason you approved but didn't merge this @rcordovano ?

PR #3155, opened 2017-10-24
3153 aid column sorting

wschaeferB:


eugene7646:
I have tested this and everything seems to work well. One thing we need to change though. When I click on any column to sort by that criteria, the results are always displayed in ascending order which is the reverse of the more intuitive chronological order. We should set the default on all columns/sorts to be in descending order.

wschaeferB:
I agree it is not intuitive that dates would be sorted ascending order first however it is consistent with the rest of our software where everything is sorted ascending order first.

I can make these few date columns sort in descending order first while leaving the Name related columns to sort in Ascending order first as would be most intuitive (Basically taking Windows Explorer behavior as the expected standard)  However that would leave the date columns here behaving differently than all the other columns in Autopsy and I feel like changing all columns containing date, size, or other numerical information where the largest should be first by default in Autopsy will be a larger task, warranting its own pull request.

eugene7646:
I guess @rcordovano can provide his opinion on this. It makes sense that you wanted to make things consistent with Autopsy case navigation. We certainly shouldn't try to change the behavior of all Autopsy columns in this story. AID being a completely separate mode/panel, my preference is to have the dates sorted in the more intuitive "latest first" order. After all, that's the original purpose of the story.

rcordovano:
@eugene7646 and @wschaeferB, you both make good points.  I think we should make the date columns sort in descending order, since that is the most useful sort in this particular use case. Thanks, guys!

wschaeferB:
Date Columns in the AID should now sort Descending order prior to sorting in Ascending order.

eugene7646:
Looks good. The only change we should make is to also display "Prioritized" column sort results in descending order. It is not intuitive to sort by "priority" and have the highest priority jobs at the bottom.

wschaeferB:
Alright change made to sort the Priority column in descending order also

PR #3157, opened 2017-10-25
Milestone 1 UI/CommuncationsManager integration

raman-bt:


Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
```diff
@@ -47,25 +52,25 @@
...
+        } else if (type.equals(Account.Type.PHONE)) {
             return "phone.png";
-        } else if (type == Account.Type.TWITTER) {
+        } else if (type.equals(Account.Type.TWITTER)) {
             return "twitter.png";
-        } else if (type == Account.Type.WEBSITE) {
+        } else if (type.equals(Account.Type.WEBSITE)) {
             return "web-file.png";
-        } else if (type == Account.Type.WHATSAPP) {
+        } else if (type.equals(Account.Type.WHATSAPP)) {
```
rcordovano: Why not use a switch statement here? I think it would improve readability significantly.

Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
```diff
@@ -88,20 +93,32 @@ protected Sheet createSheet() {
...
+        // RAMAN TBD: need to figure out how to get the right filters here
+        // We talked about either creating a wrapper class around AccountDeviceInstance to push the selected filters
+        // Or some kind of static access to pull the selected filters
+        long msgCount = 0;
+        try {
+            CommunicationsFilter filter = null;
+            msgCount = Case.getCurrentCase().getSleuthkitCase().getCommunicationsManager().getRelationshipsCount(filter, accountDeviceInstance);  
+        }
+        catch (TskCoreException ex) {
+            Exceptions.printStackTrace(ex);
```
rcordovano: Please propagate exceptions or log them instead of writing stack traces to stdout. Don't forget to remove the import of Exceptions when you change this.

Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
raman-bt: Account.Type is not an enum or a primitive type.  

To allow for creation of new account types, and not having to separate out numeric ranges for predefined vs dynamically created account type,s we defined Account.Type as class.

Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
```diff
@@ -88,20 +97,32 @@ protected Sheet createSheet() {
             s.put(properties);
         }
 
+        // RAMAN TBD: need to figure out how to get the right filters here
+        // We talked about either creating a wrapper class around AccountDeviceInstance to push the selected filters
+        // Or some kind of static access to pull the selected filters
```
millmanorama: What if we passed the filter into the node when it was created, then it could use the same filter to do query for the count

Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
rcordovano: Got it, thanks.

Core/src/org/sleuthkit/autopsy/communications/AccountNode.java
rcordovano: @millmanorama, @bcarrier has been advocating pushing the filters through as far as needed. Passing the filter through the nodes directly is a nice alternative to having a key class (@Raman's wrapper class in his comment), since it reduces the total number of abstractions. The idea behind the key class was that we would need to pair a AccountDeviceInstance with either a count of relationships or a filter, depending on where we do the relationships count query.  I suppose we could pass the counts through the node as well. Personally, I'm inclined to think we will end up passing the filter, whether in a wrapper or as a node property. The idea of the key cl;ass seems to be running out of steam.

PR #3158, opened 2017-10-25
3166 aid two column sorting

wschaeferB:
Branched off the changes changes from 3153 in PR #3155 so it should be taken first.


wschaeferB:
Adjusted to sort Priority column in descending order also.

PR #3159, opened 2017-10-26
2969 more validation on file search by attributes

zhhl:


Core/src/org/sleuthkit/autopsy/filesearch/DateSearchFilter.java
```diff
@@ -207,41 +204,26 @@ public void addActionListener(ActionListener l) {
 
     @Override
     @Messages ({
-        "End_Date_Before_Start_Date=The end date should be after the start date."
+        "End.date.before.start.date=The end date should be after the start date.",
+        "No.checkbox.selected=Non of the date search checkbox is selected."
```
APriestman: For the messages, we most commonly name them in the form <class name>.<optional subtype>.<name> As in Java class names, dots should only separate the main sections, not the words. So for example, "End.date.before.start" would be better as "DateSearchFilter.endDateBeforeStart"

rcordovano: @zhhl and @APriestman, Ann is correct. I would go one step further and do names like DateSearchFilter.errMsg.endDateBeforeStart or DateSearchFilter.errorMessage.endDateBeforeStart

Core/src/org/sleuthkit/autopsy/filesearch/HashSearchFilter.java
```diff
@@ -60,7 +61,25 @@ public void addActionListener(ActionListener l) {
     }
 
     @Override
+    @Messages({
+        "HashSearchFilter.errorMessage.emptyHash=Hash data is empty.",
+        "# {0} - hash data length", "HashSearchFilter.errorMessage.wrongLength=Input lenght({0}), doesn''t match the MD5 length(32).",
```
APriestman: Please fix this typo on length - "Input lenght({0})"

PR #3170, opened 2017-11-01
Integrate central repository hash sets into HashDbManager

APriestman:


APriestman:
Thanks I'll take another pass at the comments. I'll try to get it down to
one of those and globalSetID (for the calls into EamDb). All of the Eam
stuff needs renaming but that's another story.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -301,12 +344,12 @@ public void propertyChange(PropertyChangeEvent event) {
      *
      * @throws HashDbManagerException
      */
-    public synchronized void removeHashDatabase(HashDb hashDb) throws HashDbManagerException {
+    public synchronized void removeHashDatabase(HashDatabase hashDb) throws HashDbManagerException {
         this.removeHashDatabaseNoSave(hashDb);
         this.save();
     }
     
-    public synchronized void removeHashDatabaseNoSave(HashDb hashDb) throws HashDbManagerException {
```
wschaeferB: Are these new enough public methods that we can change them? Or do we need to add deprecated methods that take a HashDb argument?

APriestman: That one got replaced with:
public synchronized void removeHashDatabaseNoSave(HashDatabase hashDatabase) throws HashDbManagerException {
I think that will be invisible to anyone using it - HashDb implements HashDatabase, so I can't make a deprecated version anyway.

wschaeferB: HashDb implementing HashDatabase was detail I missed.  All makes sense now.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresEamDbSettings.java
```diff
@@ -370,6 +370,8 @@ public boolean initializeDatabaseSchema() {
         createReferenceSetsTable.append("org_id integer NOT NULL,");
         createReferenceSetsTable.append("set_name text NOT NULL,");
         createReferenceSetsTable.append("version text NOT NULL,");
+        createReferenceSetsTable.append("known_status integer NOT NULL,");
+        createReferenceSetsTable.append("read_only boolean NOT NULL,");
```
rcordovano: Since we have publicly released the CR, I think we will need to add some upgrade code for this schema change. We can use org.sleuthkit.datamodel.CaseDbSchemaVersionNumber (CaseDbSchemaVersionNumber => DbSchemaVersionNumber). I suggest that we make a separate story for the upgrade code.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -494,14 +586,37 @@ private void configureSettings(HashLookupSettings settings) {
...
+            for(HashDbInfo hashDbInfo : crHashDbInfoList) {
+                if(hashDbInfoIsNew(hashDbInfo)){
+                    addExistingCentralRepoHashSet(hashDbInfo.getHashSetName(), hashDbInfo.getVersion(), 
+                                hashDbInfo.getReferenceSetID(), 
+                                hashDbInfo.getSearchDuringIngest(), hashDbInfo.getSendIngestMessages(), hashDbInfo.getKnownFilesType(),
+                                hashDbInfo.isReadOnly());   
+                }
+            }
+        }
+    }
```
rcordovano: I noticed that this method is being called from just one of the UI classes, and whenever I see one class controlling the behavior of another class, rather than asking for some service, I tend to wonder if implementation details have leaked into the API. Would it be better to have the HashDbManager refresh its view of the CR hash sets when it is queried for hash sets? It think that would be less error-prone for the clients of HashDbManager (they don't need to remember to call this method) and would reduce the scope of the overall class API (always shooting for minimal but sufficient). Thoughts?

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -582,16 +693,91 @@ public String getDisplayName() {
...
+        
+        abstract String getDisplayName();
+
+        public abstract String getDatabasePath() throws TskCoreException;
+
+        public abstract SleuthkitHashSet.KnownFilesType getKnownFilesType();
+
+        public abstract boolean getSearchDuringIngest();
+        
+        abstract boolean getDefaultSearchDuringIngest();
```
rcordovano: I'm thinking the default is a property of the type of hash set, and the flag should be initialized for each concrete hash set type in the constructors of the corresponding class, eliminating the need for this method. Am I missing something?

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -786,48 +1031,340 @@ public boolean equals(Object obj) {
...
+            this.referenceSetID = referenceSetID;
+            this.searchDuringIngest = useForIngest;
+            this.sendIngestMessages = sendHitMessages;
+            this.knownFilesType = knownFilesType;
+            this.readOnly = readOnly;
+            
+            try{
+                orgName = EamDb.getInstance().getReferenceSetOrganization(referenceSetID).getName();
+            } catch (EamDbException ex){
+                Logger.getLogger(SleuthkitHashSet.class.getName()).log(Level.SEVERE, "Error looking up central repository organization", ex); //NON-NLS
```
rcordovano: I think it could help in debugging if we identify the database that experienced the error.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -786,48 +1031,340 @@ public boolean equals(Object obj) {
...
+         *
+         * @return True if the database accepts updates, false otherwise.
+         *
+         * @throws org.sleuthkit.datamodel.TskCoreException
+         */
+        @Override
+        public boolean isUpdateable() throws TskCoreException {
+            return false;
+            // TEMP this will change as soon as adding to the database is supported 
+            // return (! readOnly);
```
rcordovano: Coming in one of the other PRs?

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -786,48 +1031,340 @@ public boolean equals(Object obj) {
...
+        public boolean lookupMD5Quick(Content content) throws TskCoreException {
+            // This only works for AbstractFiles and MD5 hashes 
+            assert content instanceof AbstractFile;
+            if (content instanceof AbstractFile) {
+                AbstractFile file = (AbstractFile) content;
+                if (null != file.getMd5Hash()) {
+                    try{
+                        return EamDb.getInstance().isFileHashInReferenceSet(file.getMd5Hash(), this.referenceSetID);
+                    } catch (EamDbException ex){
+                        Logger.getLogger(SleuthkitHashSet.class.getName()).log(Level.SEVERE, "Error performing central reposiotry hash lookup", ex); //NON-NLS
```
rcordovano: I think it could help in debugging if we identify the database that experienced the error.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -786,48 +1031,340 @@ public boolean equals(Object obj) {
...
+            if (content instanceof AbstractFile) {
+                AbstractFile file = (AbstractFile) content;
+                if (null != file.getMd5Hash()) {
+                    try{
+                        if(EamDb.getInstance().isFileHashInReferenceSet(file.getMd5Hash(), this.referenceSetID)){
+                            // Make a bare-bones HashHitInfo for now
+                            result = new HashHitInfo(file.getMd5Hash(), "", "");
+                        }
+                    } catch (EamDbException ex){
+                        Logger.getLogger(SleuthkitHashSet.class.getName()).log(Level.SEVERE, "Error performing central reposiotry hash lookup", ex); //NON-NLS
```
rcordovano: I think it could help in debugging if we identify the database that experienced the error.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
```diff
@@ -786,48 +1031,340 @@ public boolean equals(Object obj) {
...
-                    NbBundle.getMessage(this.getClass(), "HashDbManager.progress.indexingHashSet", hashDb.hashSetName));
+                    NbBundle.getMessage(this.getClass(), "HashDbManager.progress.indexingHashSet", hashDb.getHashSetName()));
             progress.start();
             progress.switchToIndeterminate();
             try {
-                SleuthkitJNI.createLookupIndexForHashDatabase(hashDb.handle);
+                SleuthkitJNI.createLookupIndexForHashDatabase(hashDb.getHandle());
             } catch (TskCoreException ex) {
-                Logger.getLogger(HashDb.class.getName()).log(Level.SEVERE, "Error indexing hash database", ex); //NON-NLS
+                Logger.getLogger(SleuthkitHashSet.class.getName()).log(Level.SEVERE, "Error indexing hash database", ex); //NON-NLS
```
rcordovano: I think it could help in debugging if we identify the database that experienced the error.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamDb.java
```diff
@@ -435,6 +500,15 @@ static boolean isEnabled() {
...
     EamGlobalSet getReferenceSetByID(int globalSetID) throws EamDbException;
+    
+    /**
+     * Get all reference sets
+     *
+     * @return List of all reference sets in the central repository
+     *
+     * @throws EamDbException
+     */
+    List<EamGlobalSet> getAllReferenceSets() throws EamDbException;
```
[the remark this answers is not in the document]
APriestman: Reference sets have no type and can contain any combination of hashes, emails, etc (or they could if we had anything but files implemented). If this is not what we want, that's a conversation with Brian (and likely outside the scope of this PR)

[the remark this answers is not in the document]
rcordovano: Got it. Let's add the type stuff per discussion with Brian; let's do it later, though, so please make a story for it.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
APriestman: Assuming these close parentheses come from updateHashSetsFromCentralRepository, it is called from three places:
HashDbManager.getAllHashSets
HashDbManager.configureSettings
HashLookupModuleSettingsPanel.initializeHashSetModels

getAllHashSets covers most cases. We could add it in getKnownFileHashSets/getKnownBadFileHashSets and then the Panel wouldn't need to call it and it could be private.  It might make sense on getUpdateableHashSets, too.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
APriestman: I added this in preparation for checking whether the database is locally created. In a coming PR, getDefaultSearchDuringIngest for central repo queries the properties file for the list of names of locally created databases and returns true or false based on whether the database is found. That settings file I believe will not be updated when the HashDb object is created (it waits until Ok/Apply).

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbManager.java
APriestman: Yes the one that allows the user to add hashes to a central repo db

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresEamDbSettings.java
APriestman: JIRA-3245

PR #3171, opened 2017-11-01
3185 Reduce size of KeywordHit instances

esaunders:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/RegexQuery.java
```diff
@@ -101,6 +101,15 @@
     private String field = Server.Schema.CONTENT_STR.toString();
 
     /**
+     * The following map is an optimization to ensure that we reuse
+     * the same keyword hit String object across all hits. Even though we 
+     * benefit from G1GC String deduplication, the overhead associated with
+     * creating a new String object for every KeywordHit can be significant 
+     * when the number of hits gets large.
+     */
+    private final HashMap<String, String> keywordsFoundAcrossAllDocuments;
```
millmanorama: I assume this was explored, but I will just mention it incase:  Was string interning looked at as an alternative to this map.

esaunders:
That's an interesting idea. I had looked into string interning in the
context of the earlier string de-duplication work but not for this
particular issue.
Though from what I've read it sounds like interning may still result in
String instances (just like the G1GC string deduplication).
I'll run some tests to see if that is the case and if not I'll replace the
2 maps I've created (which I don't like anyway).

PR #3173, opened 2017-11-02
update and apply the filters when the CVT window is (re)opened

millmanorama:
plus comments and some minor cleanup


Core/src/org/sleuthkit/autopsy/communications/FiltersPanel.java
```diff
@@ -58,17 +56,27 @@
...
+        }
     }
 
     @Override
     public void addNotify() {
         super.addNotify();
+        /*
+         * Since we get the exploreremanager from the parent JComponenet, wait
+         * till this FiltersPanel is actaully added to a parent.
+         */
```
rcordovano: Several typos in this comment.

PR #3180, opened 2017-11-06
3148 alert files

dgrove727:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -1254,29 +1265,32 @@ private void doRecoveryIfCrashed(Manifest manifest, AutoIngestJobNodeData nodeDa
...
+                         * Update the case node data and do the logging.
                          */
                         if (null != caseDirectoryPath) {
                             try {
-                                AutoIngestAlertFile.create(nodeData.getCaseDirectoryPath());
-                            } catch (AutoIngestAlertFileException ex) {
-                                SYS_LOGGER.log(Level.SEVERE, String.format("Error creating alert file for crashed job for %s", manifestPath), ex);
+                                CaseNodeData caseNodeData = new CaseNodeData(coordinationService.getNodeData(CoordinationService.CategoryNode.CASES, caseDirectoryPath.toString()));
+                                caseNodeData.setErrorsOccurred(true);
+                                updateCoordinationServiceCaseNode(caseNodeData, caseDirectoryPath);
```
rcordovano: These three lines of code are duplicated many times in this class. Please create a method for them. Perhaps we don't need  updateCoordinationServiceCaseNode.

Core/src/org/sleuthkit/autopsy/casemodule/CaseNodeData.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * An object that converts case data for a case directory coordination service
+ * node to and from byte arrays.
+ */
+public final class CaseNodeData {
+
+    private static final int CURRENT_VERSION = 0;
+    
+    private int version;
+    private boolean errorsOccurred;
```
rcordovano: Now that this data is available to the MultiUserCaseManager, it should be be read in MultiUserCaseManager.getCases and stored in MultiUserCaseManager.MultiUserCase.  As stated in the DoD of the story for this work in JIRA:

"The Status column of the open multi-user case dialog is populated using the flag in the node data instead of the presence or absence of an "alert" file in the case directory."

Looking at the code in MultiUserCase.

`   
     CaseStatus getStatus() {
            if(caseDirectoryPath.resolve("autoingest.alert").toFile().exists()) {
                return CaseStatus.ALERT;
            } else {
                return CaseStatus.OK;
            }
        }
`

this code, as is, will fail to report errors except for cases created when alert files were still written.

Also, considering our concerns about how the multi-user case panel code is performing, we should check for legacy alert files in MultiUserCaseManager.getCases, where we have already done a file listing.

PR #3194, opened 2017-11-09
3198 tag properties update

wschaeferB:
This PR changes how the Tags are stored.  The STANDARD_NOTABLE_TAG_DISPLAY_NAMES list is changed in VIK-3199 to be populated in a more appropriate fashion.  This PR also causes the setting a tag as Notable to be broken until the GUI changes in VIK-3201.


Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
```diff
@@ -35,9 +37,12 @@
 
     private static final String TAGS_SETTINGS_NAME = "Tags"; //NON-NLS
     private static final String TAG_NAMES_SETTING_KEY = "TagNames"; //NON-NLS    
+    private static final String STANDARD_NOTABLE_TAG_DISPLAY_NAMES = "Evidence,Notable Item,"
+            + "CAT-1: Child Exploitation (Illegal),CAT-2: Child Exploitation (Non-Illegal/Age Difficult),CAT-3: Child Exploitive"; // NON-NLS
```
rcordovano: I believe Notable Item is supposed to replace Evidence. Also, why are only three of the CAT tags included here, instead of all five?

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
```diff
@@ -35,9 +37,12 @@
...
     private final String displayName;
     private final String description;
     private final TagName.HTML_COLOR color;
+    private final TskData.FileKnown knownStatusDenoted;
```
rcordovano: I believe we agreed tom just call this knownStatus, notKnownStatusDenoted.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
```diff
@@ -152,15 +164,37 @@ private String toSettingsFormat() {
...
+                    }
+                    if (badTags.contains(tagNameAttributes[0])) {
+                        tagNames.add(new TagNameDefiniton(tagNameAttributes[0], tagNameAttributes[1], TagName.HTML_COLOR.valueOf(tagNameAttributes[2]), TskData.FileKnown.BAD));
+                    } else {
+                        tagNames.add(new TagNameDefiniton(tagNameAttributes[0], tagNameAttributes[1], TagName.HTML_COLOR.valueOf(tagNameAttributes[2]), TskData.FileKnown.UNKNOWN)); //add the default value for that tag 
+                    }
+                } else if (tagNameAttributes.length == 4) {
+                    tagNames.add(new TagNameDefiniton(tagNameAttributes[0], tagNameAttributes[1], TagName.HTML_COLOR.valueOf(tagNameAttributes[2]), TskData.FileKnown.valueOf(tagNameAttributes[3])));
+                }
+
```
rcordovano: I do not understand what is going on in this code block. Let's discuss it, then comment it.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagOptionsPanel.java
```diff
@@ -190,7 +191,7 @@ private void newTagNameButtonActionPerformed(java.awt.event.ActionEvent evt) {//
         NewTagNameDialog.BUTTON_PRESSED result = dialog.getResult();
         if (result == NewTagNameDialog.BUTTON_PRESSED.OK) {
             String newTagDisplayName = dialog.getTagName();
-            TagNameDefiniton newTagType = new TagNameDefiniton(newTagDisplayName, DEFAULT_DESCRIPTION, DEFAULT_COLOR);
+            TagNameDefiniton newTagType = new TagNameDefiniton(newTagDisplayName, DEFAULT_DESCRIPTION, DEFAULT_COLOR, TskData.FileKnown.UNKNOWN);
```
rcordovano: Was the old code here using place holders for non-editable properties? Is editing not included in this PR?

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -97,7 +99,17 @@ public static boolean containsIllegalCharacters(String tagDisplayName) {
             /*
              * No current case, nothing more to add to the set.
              */
-        } 
+        }
+        return tagDisplayNames;
+    }
+
+    public static List<String> getNotableTagDisplayNames() {
```
rcordovano: Who is the client for this method...does it need to be public?

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -224,11 +237,36 @@ public synchronized TagName addTagName(String displayName, String description) t
      *                                       name to the case database.
      */
     public synchronized TagName addTagName(String displayName, String description, TagName.HTML_COLOR color) throws TagNameAlreadyExistsException, TskCoreException {
+        TskData.FileKnown knownStatus = TskData.FileKnown.UNKNOWN;
+        if (getNotableTagDisplayNames().contains(displayName)) {
+            knownStatus = TskData.FileKnown.BAD;
+        }
```
rcordovano: I do not understand this logic.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
wschaeferB: At the point of this pull request there is no Notable Item tag being made, the evidence tag is still being made, and for static String definitions the Category class has not been moved to the Core module so those names are not accessible and I haven't moved the static text for Follow Up or created it for Notable Item.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
rcordovano: Please add some comments, thanks!

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -97,7 +99,17 @@ public static boolean containsIllegalCharacters(String tagDisplayName) {
...
+        return tagDisplayNames;
+    }
+
+    public static List<String> getNotableTagDisplayNames() {
+        List<String> tagDisplayNames = new ArrayList<>();
+        for (TagNameDefiniton tagDef : TagNameDefiniton.getTagNameDefinitions()) {
+            if (tagDef.isNotable()) {
+                tagDisplayNames.add(tagDef.getDisplayName());
+            }
+        }
```
rcordovano: Let's make this a utility method in the centralrepository package rather than expanding the public API here.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -224,11 +237,36 @@ public synchronized TagName addTagName(String displayName, String description) t
      *                                       name to the case database.
      */
     public synchronized TagName addTagName(String displayName, String description, TagName.HTML_COLOR color) throws TagNameAlreadyExistsException, TskCoreException {
+        TskData.FileKnown knownStatus = TskData.FileKnown.UNKNOWN;
+        if (getNotableTagDisplayNames().contains(displayName)) {
+            knownStatus = TskData.FileKnown.BAD;
+        }
+        return addTagName(displayName, description, color, knownStatus);
```
rcordovano: Let's just go ahead and pass unknown to the four args method, I think it will be a tad simpler.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
rcordovano: Ah, changing it in next PR.

Issue #1010, opened 2014-12-30
scalpel - included (partial?) but apparently nonfunctional & no ref in KNOWN_ISSUES.txt 


johnmccash:
I'm guessing, based on the response to issue #403 last March, that the scalpel plugin doesn't really work yet. That being the case, this 'bug' really relates to the lack of documentation in the release regarding this issue. Please update KNOWN_ISSUES.txt, and possibly the integral help. It might be worthwhile, as well, to either remove the plugin until it can actually be used, or else do something to make it obvious to people looking through their plugins that it doesn't work, so they don't waste time on it. On the other hand, if it can be used in some fashion, it would be nice to have some sort of documentation as to how exactly to employ it.
Thanks

bcarrier:
Fair enough.  The next release has a PhotoRec plug-in, so that Scalpel effort is dead for quite a while until some memory leaks can be fixed in it. We should make it invisible to the user though.

bcarrier:
Just removed it from the Autopsy project so that it is not visible to user in plug-ins area. 

Checkin: 57811bb

PR #3200, opened 2017-11-11
868 attachments

millmanorama:
adds an embedded DataResultPanel to the MessageContentViewer allowing the user to view attachments for messages

depends on Sleuthkit [PR #1010](https://github.com/sleuthkit/sleuthkit/pull/1010)


Core/src/org/sleuthkit/autopsy/corecomponentinterfaces/DataResultViewer.java
```diff
@@ -90,5 +90,6 @@
      * @param contentViewer content viewer to respond to selection events from
      *                      this viewer
      */
+    @Deprecated
```
rcordovano: Why deprecate this method? 

Also, assuming we proceed to do so, the @deprecated directive should be added to the method doc.

raman-bt:
@millmanorama: I'm not seeing the attachment Thumbnails. Also something goofy is going on with the Message tab in the Data Result viewer section. I sent you an email with screenshot.

millmanorama:
I'll take look

PR #3203, opened 2017-11-14
3172 size metrics collection

dgrove727:
Built on top of 3177.


dgrove727:
Replaced with https://github.com/sleuthkit/autopsy/pull/3204

PR #3209, opened 2017-11-15
869 daterange filter GUI

millmanorama:


Core/src/org/sleuthkit/autopsy/communications/Bundle.properties
```diff
@@ -8,3 +8,6 @@ FiltersPanel.unCheckAllAccountTypesButton.text=Uncheck All
 FiltersPanel.checkAllAccountTypesButton.text=Check All
 FiltersPanel.unCheckAllDevicesButton.text=Uncheck All
 FiltersPanel.checkAllDevicesButton.text=Check All
+FiltersPanel.dateRangeLabel.text=Date Range:
+FiltersPanel.startCheckBox.text=start:
+FiltersPanel.endCheckBox.text=end:
```
raman-bt: Do these start/end label need to be capitalized

PR #3210, opened 2017-11-15
3207: Add a collection of attributes to artifact instead of add one …

zhhl:
…attribute at a time.


zhhl:
Close now to wait for the python code be updated.

PR #3212, opened 2017-11-16
put check lists in scroll panes, some other UI tweaks

millmanorama:


rcordovano:
This does not auto-merge, please address and resubmit.

PR #3214, opened 2017-11-17
881 cvt refresh

millmanorama:
add listeners to refresh device filter and change apply button to refresh.


rcordovano:
This does not auto-merge, please address and resubmit.

PR #3218, opened 2017-11-17
Crypto module created for detecting encrypted files.

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/crypto/CryptoDetectionFileIngestModule.java
```diff
@@ -0,0 +1,242 @@
...
+                !file.getType().equals(TskData.TSK_DB_FILES_TYPE_ENUM.LOCAL_DIR)) {
+            /*
+             * Qualify the file against hash databases.
+             */
+            if (!file.getKnown().equals(TskData.FileKnown.KNOWN)) {
+                /*
+                 * Qualify the size.
+                 */
+                long contentSize = file.getSize();
+                if (contentSize >= 0x500000 && (contentSize & 511) == 0) {
```
bcarrier: For consistency, we should use 5242880 and (contentSize % 512) == 0.

PR #3220, opened 2017-11-17
3199 consolidate default tags

wschaeferB:
Requires PR https://github.com/sleuthkit/sleuthkit/pull/1015 for Sleuthkit

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
+++ b/Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
@@ -24,13 +24,14 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.logging.Level;
 import javax.annotation.concurrent.Immutable;
-import org.openide.util.Exceptions;
 import org.openide.util.NbBundle;
 import org.sleuthkit.autopsy.casemodule.Case;
+import org.sleuthkit.autopsy.coreutils.Logger;
 import org.sleuthkit.autopsy.coreutils.ModuleSettings;
 import org.sleuthkit.datamodel.TagName;
-import org.sleuthkit.autopsy.datamodel.tags.Category;
+import org.sleuthkit.autopsy.datamodel.DhsImageCategory;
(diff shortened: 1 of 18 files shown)


Core/src/org/sleuthkit/autopsy/casemodule/services/TagNameDefiniton.java
```diff
@@ -156,40 +177,55 @@ private String toSettingsFormat() {
...
+            if (badTagsStr == null || badTagsStr.isEmpty()) {  //if there were no bad tags in the central repo properties file use the default list
+                notableTags.addAll(STANDARD_NOTABLE_TAG_DISPLAY_NAMES);
+            } else {  //otherwise use the list that was in the central repository properties file
+                notableTags.addAll(Arrays.asList(badTagsStr.split(",")));
+            }
+            for (String tagNameTuple : tagNameTuples) { //for each tag listed in the tags properties file
+                String[] tagNameAttributes = tagNameTuple.split(","); //get the attributes
+                if (tagNameAttributes.length == 3) {  //if there are only 3 attributes so Tags.properties does not contain any tag definitions with knownStatus
+                    standardTags.remove(tagNameAttributes[0]);  //remove tag from default tags we need to create still
+                    if (notableTags.contains(tagNameAttributes[0])) { //if tag should be notable mark create it as such
```
rcordovano: Perhaps we can split this code into two methods? I know there could be duplication, but I'm thinking the logic would be easier to follow if we checked the tuple length and then branched into either a method to do the upgrade from the Central Repo properties, or just parse the tag properties file. It would also mean that we would not need to read the Central Repo properties file more than once.

Core/src/org/sleuthkit/autopsy/datamodel/tags/Category.java
```diff
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.sleuthkit.autopsy.imagegallery.datamodel;
+package org.sleuthkit.autopsy.datamodel.tags;
```
rcordovano: I think this class is more general purpose. Rather than put it in a new package, I think it can go into the datamodel package, just with a more specific name: DhsImageCategory.

PR #3227, opened 2017-11-20
894 polish cvt opening behaviour

millmanorama:
do DB queries off the EDT
use ChildFactory and asynchronous node creation


Core/src/org/sleuthkit/autopsy/communications/AccountDetailsNode.java
```diff
@@ -39,37 +38,43 @@
 
     private final static Logger logger = Logger.getLogger(AccountDetailsNode.class.getName());
 
-    AccountDetailsNode(Set<AccountDeviceInstance> accountDeviceInstances, CommunicationsFilter filter, CommunicationsManager commsManager) {
-        super(new AccountRelationshipChildren(accountDeviceInstances, commsManager, filter));
+    AccountDetailsNode(Set<BlackboardArtifact> accountDeviceInstances, CommunicationsFilter filter, CommunicationsManager commsManager) {
```
rcordovano: Artifacts are account file instances, so I am finding this a bit confusing. Are we moving away from the AccountDeviceInstance class, which has no corresponding table in the case database?

rcordovano:
Does not auto-merge.

Core/src/org/sleuthkit/autopsy/communications/AccountDetailsNode.java
millmanorama: the change here was from passing the accountDeviceInstances to passing the blackboard artifacts that represent the relationships.  I was focussing on how to the get the queries off the EDT, but in that process I kind of distorted the purpose of AccountDetailsNode.  I think in the future we will want it to be more account focussed, and less relationship focussed.  I will review this and see if there is something clearer to do.

PR #3228, opened 2017-11-20
3202 tags messaging on status change

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -294,6 +298,97 @@ public void run() {
...
+                                    continue;
+                                }
+                                //if any other tags on this artifact are Notable in status then this artifact can not have its status changed 
+                                if (notableTags.contains(t.getName().getDisplayName())) {
+                                    hasOtherBadTags = true;
+                                    break;
+                                }
+                            }
+                        }
+                        if (!hasOtherBadTags) {
```
rcordovano: I have not been able to figure out what this method is doing. It  seems possible that this code belongs outside of the loop, given that the flag is being set and used in the loop and the loop breaks if the flag is set. I'm just not sure though. Please add some comments to clarify the algorithm.

wschaeferB: This code is already outside of inner most for loop which the flag is set in and breaks out of upon being set.  More comments are incoming though

PR #3229, opened 2017-11-20
 Enable any central repo has sets the user personally creates by default.

APriestman:
Also clear out the list of new central repo hashsets when settings are being saved.


rcordovano:
I believe the newReferenceSetIDs.clear() call you were looking for is in this PR. 

This PR does not auto-merge though; please address and re-submit, thanks!

PR #3261, opened 2017-12-05
Handle build error after merging develop to accounts_relationships br…

raman-bt:
…anch


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/RegexQuery.java
```diff
@@ -518,10 +519,15 @@ private void createCCNAccount(Content content, Keyword foundKeyword, KeywordHit
...
+           
             if (hit.isArtifactHit()) {
                 LOGGER.log(Level.SEVERE, String.format("Failed to parse credit card account number for artifact keyword hit: term = %s, snippet = '%s', artifact id = %d", foundKeyword.getSearchTerm(), hit.getSnippet(), hit.getArtifactID().get())); //NON-NLS
             } else {
-                LOGGER.log(Level.SEVERE, String.format("Failed to parse credit card account number for content keyword hit: term = %s, snippet = '%s', object id = %d", foundKeyword.getSearchTerm(), hit.getSnippet(), hit.getContentID())); //NON-NLS
+                try {
+                    LOGGER.log(Level.SEVERE, String.format("Failed to parse credit card account number for content keyword hit: term = %s, snippet = '%s', object id = %d", foundKeyword.getSearchTerm(), hit.getSnippet(), hit.getContentID())); //NON-NLS
+                } catch (TskCoreException ex) {
+                    LOGGER.log(Level.SEVERE, "There was a error getting contentID for keyword hit.", ex); //NON-NLS
+                }
```
rcordovano: I would rather not lose the exception about the parse failure. Please work this so that if the try fails, we log a less specific parsing error message in addition to the message about failing to get the content id, thanks!

PR #3271, opened 2017-12-06
Added TSK_CORE_ERROR to exception checks.

dgrove727:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -2883,7 +2904,8 @@ boolean isStale(Instant lastUpdateTime) {
         FILE_EXPORT_ERROR("File export error"),
         JOB_LOGGER_ERROR("Job logger error"),
         DATA_SOURCE_PROCESSOR_ERROR("Data source processor error"),
-        UNEXPECTED_EXCEPTION("Unknown error");
+        UNEXPECTED_EXCEPTION("Unknown error"),
+        TSK_CORE_ERROR("TSK core error");
```
rcordovano: Propagating a "raw" TskCoreException is inconsistent with the current design for exception handling in AutoIngestManager. Exceptions in the job processing task are propagated, processing is paused, waiting for human intervention, and an error state is set so that the ErrorState.toString method can be used to put a fairly specific error message on the auto ingest control panel to indicate what went wrong. So what we need here is a wrapper, e.g., the existing DatabaseServerDownException. I would call a new wrapper something like JobMetricsCollectionException.

Frankly, I have debated whether we want to stop processing for this, but it would indicate a problem with the case database. I don't expect it to come up, but I have decided I am fine with propagating the error as described above.

PR #3287, opened 2017-12-08
3287 make actions for addDataSource and RunIngestModules wait cursor

wschaeferB:


wschaeferB:
Found issue with clicking while wait cursor was up.

PR #3289, opened 2017-12-08
3278 unassign agency logo

dgrove727:


Core/src/org/sleuthkit/autopsy/corecomponents/Bundle.properties
```diff
@@ -201,3 +200,5 @@ DataResultViewerThumbnail.sortButton.text=Sort
 CriterionChooser.ascendingRadio.text=\u25b2 Ascending\n
 CriterionChooser.removeButton.text=Remove
 CriterionChooser.descendingRadio.text=\u25bc Descending
+AutopsyOptionsPanel.defaultLogoRB.text=Use default
+AutopsyOptionsPanel.specifyLogoRB.text=Specify a logo
```
wschaeferB: New messages should be added as @Messages instead of directly to the Bundle.properties files

rcordovano:
Taking this, with additional work to follow.

Issue #3239, opened 2017-11-25
deadlock.netbeans.org is down


tukusejssirs:
I tried to install/build `autopsy` on Ubuntu 17.10 AMD64 (aka x86_64).

I’ve got this:

```
$ ant build
Buildfile: /media/ubuntu/DVD/ubuntu_vanilla_changes/autopsy/autopsy-autopsy-4.5.0/build.xml

download:
     [echo] Downloading clusters harness|java|platform
      [get] Getting: http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar
      [get] To: /tmp/tasks.jar
      [get] Error opening connection java.io.IOException: Server returned HTTP response code: 503 for URL: http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar
      [get] Error opening connection java.io.IOException: Server returned HTTP response code: 503 for URL: http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar
      [get] Error opening connection java.io.IOException: Server returned HTTP response code: 503 for URL: http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar
      [get] Can't get http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar to /tmp/tasks.jar

BUILD FAILED
/media/ubuntu/DVD/ubuntu_vanilla_changes/autopsy/autopsy-autopsy-4.5.0/build.xml:8: The following error occurred while executing this line:
/media/ubuntu/DVD/ubuntu_vanilla_changes/autopsy/autopsy-autopsy-4.5.0/nbproject/build-impl.xml:41: The following error occurred while executing this line:
/media/ubuntu/DVD/ubuntu_vanilla_changes/autopsy/autopsy-autopsy-4.5.0/nbproject/platform.xml:27: Can't get http://deadlock.netbeans.org/hudson/job/nbms-and-javadoc/lastStableBuild/artifact/nbbuild/netbeans/harness/tasks.jar to /tmp/tasks.jar

Total time: 0 seconds
```

[...]

PR #3292, opened 2017-12-10
changed defunct netbeans url (deadlock.netbeans.*) to the current url

seannicholls:
fixes issue #3239


seannicholls:
@estuans Thanks for spotting that. Bizarrely when I tried it locally it seemed to work.

Issue #2938, opened 2017-07-11
Make a Linux Package of Autopsy


bcarrier:
I'm making this issue to track progress on making a package for Linux.  Last time this was discussed, there were some Java build issues that needed to be resolved. But, the details were not fully  understood at the time.

kcgthb:
Not quite complete packaging yet, but those instructions at least give some solid foundation for building Autopsy on Linux: https://periciacomputacional.com/linux-install-autopsy/

marshalla99:
It'll compile, but won't run cleanly.

Case setup works fine, but attempting to add a data source, or alter the Keyword Search options throws a null pointer exception - similar to https://github.com/sleuthkit/autopsy/issues/2210 - but manifesting as
```
[exec] SEVERE [global]
     [exec] java.lang.NullPointerException
     [exec] 	at org.sleuthkit.autopsy.keywordsearch.KeywordSearchGlobalSettingsPanel.addPropertyChangeListener(KeywordSearchGlobalSettingsPanel.java:56)
```

I also think you need to add a
            `  apt-get install solr-tomcat`
to your script to get the solr & lucene server installed for keyword searching.

oylenshpeegul:
For what it's worth, [I got Autopsy 4.5 to compile and "run" in Xubuntu 17.04 today](https://oylenshpeegul.wordpress.com/2017/10/18/autopsy-4-5-works-in-linux/).

marshalla99:
Ditto, on Debian. I've made a little progress by disabling the Keyword Search module for now.

rcordovano:
@marshalla99, a quick analysis suggests to me that your conjecture about #2210 is correct and wishdasher's fix on Sept 12, 2016 missed a place where super class addPropertyChangeListener/removePropertyChangeListener methods need to be invoked, i.e., in KeywordSearchGlobalSettingsPanel and any child components that override addPropertyChangeListener/removePropertyChangeListener.

marshalla99:
Thanks @rcordovano Unfortunately, I'm not a Java programmer so I'm kinda hoping someone might have time to patch that. FWIW, the rest of the modules seem to be working pretty well. I have successfully ingested a 2Gb dd image and carried out some simple viewing tasks.

My goal is to get a reasonably stable version running on a live distro., so I can use it when I teach my Masters forensic computing module in March.

rcordovano:
@marshalla99, I have made a fix to the add/remove property change listener code of these two classes:

[...]

Issue #3291, opened 2017-12-10
Make package / install candidate for Mac OS X


seannicholls:
Similar to #2938 and #2863

I believe much of the work done in getting Autopsy to work for Linux can be used to also get it working satisfactorily on Mac OS X systems. That said, there will doubtless be system specific concerns to overcome, hence the need for this feature request.

I have successfully built Autopsy on Ubuntu 16.04 LTS and will be contributing to efforts to make a more streamlined 'official' means than is presently available. I have yet to be able to produce something which works on a Mac OS X system, however. As such, I think some effort should be put into getting Autopsy working on Mac Operating Systems.

marshalla99:
Given the issues with some of the functionality, even after a "successful" build, I think we have bigger problems than "just" porting to deal with. 

We need to get a relatively stable version which works on both Windows and Linux, and which handles installation of alternative supporting software (e.g. PhotoRec) before we introduce a third platform, IMNSHO.

seannicholls:
@marshalla99 I don't disagree, however I would like to keep this issue open as something to aim for.

I think that once Linux issues are resolved, I think porting to Mac should be "relatively easy" - at least by comparison.

In terms of priority, I would certainly rate it lower than Linux by a wide margin.

marshalla99:
I agree.

Right now - we seem to have some java code which is not particularly portable. If that is fixed for Linux it *might* be OK for Mac as well. That would leave the issue of the external support programs and their installation. Given the *nix nature of Mac, that side of porting would, I hope, be simpler - especially if we can use Macports, fink or Homebrew to assist with those.

marshalla99:
I take back my comment about the java code as it seems it was more of a missing package problem. The Linux build is actually pretty stable and, I think, gives a pretty good indication of the sort of dependencies we may have on Mac.

bcarrier:
Closign this since you can now install on OS X.  Instructions are [here](https://github.com/sleuthkit/autopsy/blob/develop/Running_Linux_OSX.txt)

PR #3293, opened 2017-12-11
Add Linux build instructions to README

seannicholls:
* Deleted stub file README_LINUX_OSX.md as I believe each platform deserves its own file given the complexity of setting up the build process (at least a present).

* Added a README file for each Linux and OSX, with the Linux README file containing instructions on how to build Autopsy for Ubuntu 16.04

This pull request should serve as an initial work to begin to resolve issues #2938 #2863 and ultimately #3291 

As I felt it was inappropriate to discuss as part of a README, and comments are not available in the file format, I would like to take the opportunity to acknowledge the great work by other community members in contributing to the information provided here, with particular thanks to:

Angus Marshall (angus@n-gate.net)
Petter Lopes (https://periciacomputacional.com/linux-install-autopsy/)
Oylenshpeegul (https://oylenshpeegul.wordpress.com/2017/10/18/autopsy-4-5-works-in-linux/)


README_LINUX.md
```diff
@@ -0,0 +1,139 @@
...
+wget -c http://central.maven.org/maven2/org/postgresql/postgresql/9.4.1211.jre7/postgresql-9.4.1211.jre7.jar 
+wget -c http://central.maven.org/maven2/com/mchange/c3p0/0.9.5/c3p0-0.9.5.jar
+```
+
+# Building Autopsy
+
+#### Get a local copy of autopsy (if you haven't already)
+```bash
+cd ${WORKDIR}
+git clone git@github.com:seannicholls/autopsy.git
```
bcarrier: Is there anything special about your version of the repo or should this point people to the sleuthkit one?

README_LINUX.md
```diff
@@ -0,0 +1,139 @@
...
+cp ./bindings/java/dist/Tsk_DataModel.jar ${TSK_HOME}/bindings/java/dist/Tsk_DataModel_PostgreSQL.jar
+cp ./bindings/java/lib/mchange-commons-java-0.2.9.jar ${TSK_HOME}/bindings/java/lib/
+cd ${WORKDIR}
+```
+
+### Java & Python bindings
+```bash
+mkdir -p ${TSK_HOME}/tsk/bindings/java/lib
+cd ${TSK_HOME}/tsk/bindings/java/lib
+wget -c https://repo1.maven.org/maven2/org/xerial/sqlite-jdbc/3.8.11/sqlite-jdbc-3.8.11.jar
```
bcarrier: This is surprising that you need to do this manually.  It should be done automatically by 'ivy' when you type in 'make' in the root TSK folder.  It works on my Mac. 

Or, is this because you have TSK_SRC set differently from TSK_HOME?   If the later, you can probably simplify this and copy bindings/java/lib from TSK_SRC to TSK_HOME and leave the list of specific libraries only in the ivy file.

README_LINUX.md
```diff
@@ -0,0 +1,139 @@
...
+```bash
+export JAVA_HOME=/usr/lib/jvm/java-8-oracle
+export CLASSPATH=$JAVA_HOME/lib/:$CLASSPATH
+export PATH=$JAVA_HOME/bin/:$PATH
+export JDK_HOME="/usr/lib/jvm/java-8-oracle/"
+export JRE_HOME="/usr/lib/jvm/java-8-oracle/jre/"
+export TSK_HOME="/opt/tsk"
+```
+
+### Set up TSK_HOME
```
bcarrier: I don't quite understand this step.  Autopsy should be copying the datamodel JAR into its directory structure and therefore it shouldn't care if it is 'opt' or a user folder.  Or, is the design here that TSK_HOME is always in a predefined place and you can clone the TSK repo into wherever you want w/out needing to change the TSK_HOME environment variable?

README_LINUX.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+### The Slueth Kit (TSK)
+```bash
+git clone https://github.com/sleuthkit/sleuthkit.git
+cd sleuthkit
+export TSK_SRC=`pwd`
+./bootstrap
+./configure --prefix=${TSK_HOME} --with-libewf=${LIBEWF_SRC}
+sed -i '67s/libewf_handle_read_random/libewf_handle_read_buffer_at_offset/' ${TSK_SRC}/tsk/img/ewf.c
```
bcarrier: Is this because the libewf you get will be the latest version that is still experimental and has a new method name?

bcarrier:
Thanks. This was useful for me to see what people are doing to get Linux up and running.  I added some clarifying questions, but we can certainly pull it in as is.

README_LINUX.md
seannicholls: No, that was  a mistake. I'll fix that

README_LINUX.md
seannicholls: TSK_Home is required to get Netbeans to work, the particular path is not important, but I chose /opt/tsk because its a fair assumption to think that whoever is going to build Autopsy from source may need to do so again in the future, so rather than picking a random location, I picked this so it can be re-used.

I'll add a comment to point out that the particular location is not important.

README_LINUX.md
seannicholls: issue #642 has background information on this

Essentially, the function needs to be renamed or else it will cause build failure. I cant speak to the originating cause, whether non-experimental versions of LIBEWF don't need this. But I've been unable to get a working build without it.

README_LINUX.md
seannicholls: There are quite a few missing dependencies, this is just one example.

It's not a result of TSK_SRC, but its something that has become necessary going through the various comments/scripts to get Autopsy to build. 

Not doing this manually on Ubuntu causes a build failure on my system. Perhaps it is something that should be investigated along with other missing dependencies?

seannicholls:
@bcarrier Thanks for the feedback. I've committed some changes to clarify based on your feedback.

README_LINUX.md
seannicholls: Given @bcarrier comments on another issue, I think its safe to say that these are the result of 'ivy' not fully satisfying dependencies on Linux

seannicholls:
I also should point out, these instructions currently are missing instructions to fix the deadlock servers as resolved by #3292 so if this is to be pulled, then that should be merged first

README_LINUX.md
h3xx: sleuthkit/sleuthkit#1052 should clear this issue up if it's merged. Better to have support for the library fixed rather than using `sed` to patch the source code.

bcarrier: The problem is that the head of the libewf github repo is the non-stable release and I don't see a tag to get it. I'd suggest either:
- Use the repo we have that is a fork of libewf from the stable release: https://github.com/sleuthkit/libewf_64bit
- Use a package. For our Debian work, we've just been using the libewf-dev package.

If you do either of these, no TSK code changes are required.

README_LINUX.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+#### Build Autopsy
+At this point you should be able to use Netbeans IDE to build the project, or
+alternatively, you can build directly from the commandline using ant:
+
+##### Bug-fix for Ubuntu Java tmp folder writing permissions
+```bash
+mkdir -p ./tmp
+export _JAVA_OPTIONS=-Djava.io.tmpdir=`pwd`/tmp
```
bcarrier: Is this related to the DEAD JAR error messages people were getting?  Do you effectively want to have each user have their own temp folder?

seannicholls: This is only for building the application, not for running.

I cant find the report off-hand, but bulding without this causes the java compiler to bail.

obilodeau:
What is missing to push this PR forward? I'm interested in fixing Arch Linux's [autopsy AUR package](https://aur.archlinux.org/packages/autopsy/) so I might as well push this PR forward. Don't know where to start though.

bcarrier:
@obilodeau , I had been thinking about closing this because we have since gone through the work to make this easier to run on Linux. There seems to be a need for two docs.  One is how to run the packaged version on Linux, we we have [here](https://github.com/sleuthkit/autopsy/blob/develop/Running_Linux_OSX.txt).  The other is how to build it from source on Linux for those who want to develop (or want to make their own package).     The best place for building instructions is the travis config files, such as [here](https://github.com/sleuthkit/autopsy/blob/develop/.travis.yml).  

Regarding the error you are seeing from that ticket though, it looks like it can't find the JAR libraries that should be installed by sleuthkit-java.   Are you sure that sleuthkit-java is installing the sleuthkit-postgresql-4.6.1.jar file?

obilodeau:
You are right! The packaged version pulled from ArchLinux's AUR is broken but the one downloaded from your website works after commenting out the `$JAVA_HOME` test in `unix_setup.sh` ([arch doesn't use JAVA_HOME](https://wiki.archlinux.org/index.php/Java_package_guidelines#Java_packaging_on_Arch_Linux)). Thank to your comment I saved a lot of time.

I haven't tried building but for running, I agree that this PR can be closed. I'll try to fix AUR package on my own, its built in a very weird way anyways (extracting the whole zip in `/usr`, not following unix conventions), I'll try to put everything in `/opt` and see what fails from there.

> Regarding the error you are seeing from that ticket though, it looks like it can't find the JAR libraries that should be installed by sleuthkit-java. Are you sure that sleuthkit-java is installing the sleuthkit-postgresql-4.6.1.jar file?

I was having a different error but since this was about running it on Linux I was willing to give it a try.

PR #3295, opened 2017-12-11
updated build.xml to download dependencies if they do not exist

seannicholls:
Added checks in build.xml for Autopsy Core to check if required dependencies exist and to download them if they are missing.

Particularly useful for building on Linux.


bcarrier:
Hello,

The current plan has been for 'ivy' to download the needed files and The TSK datamodel uses ivy to download its dependencies.   it looks like you are adding additional statements to download files that should have already been downloaded by the TSK datamodel ivy run. Are they not there in your build process?  

LGoodDatePicker should be pulled in by Autopsy's ivy: 
https://github.com/sleuthkit/autopsy/blob/develop/CoreLibs/ivy.xml#L16

seannicholls:
@bcarrier ah, I see.

It looks like 'ivy' is not working well on my Ubuntu 16.04 LTS test system. Perhaps that should be resolved instead of using this PR.

seannicholls:
Am closing this in favour of fixing 'ivy' dependency management system.

PR #3296, opened 2017-12-11
disable Timeline and image gallery if there are more than 6M files in…

millmanorama:
… the DB


Core/src/org/sleuthkit/autopsy/guiutils/JavaFXUtils.java
```diff
@@ -0,0 +1,84 @@
...
+import javafx.stage.Stage;
+import org.openide.util.NbBundle;
+import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.autopsy.coreutils.ThreadConfined;
+
+/**
+ * Utilities for dealing with JavaFX gui components.
+ *
+ */
+public class JavaFXUtils {
```
rcordovano: This class does not feel like it is very cohesive, and I'm not enthusiastic about adding adding a showTooManyFiles method to the public API, which I personally view as a stop gap measure. I think I would rather have some code duplication here, consistent with the decomposition of the code into NBMs, then expand the API this way.

Core/src/org/sleuthkit/autopsy/timeline/OpenTimelineAction.java
```diff
@@ -78,15 +82,26 @@ public boolean isEnabled() {
...
+        return super.isEnabled() && Case.isCaseOpen() && Installer.isJavaFxInited();
     }
 
     @Override
     @ThreadConfined(type = ThreadConfined.ThreadType.AWT)
     public void performAction() {
-        showTimeline();
+        if (tooManyFiles()) {
+            Platform.runLater(() ->
+                    JavaFXUtils.showTooManyFiles(Bundle.Timeline_dialogs_title()));
```
rcordovano: See prior comment.

Core/src/org/sleuthkit/autopsy/guiutils/JavaFXUtils.java
millmanorama: ok I will split it up and keep it out of the API

PR #3324, opened 2017-12-15
 922: Put DataSource name in AccountDeviceInstanceKey 

raman-bt:


raman-bt:
Review comment addressed.

PR #3344, opened 2017-12-21
937: Accounts sub-trees do not refresh their contents after being ini…

raman-bt:
…tial viewed


Core/src/org/sleuthkit/autopsy/datamodel/accounts/Accounts.java
```diff
@@ -201,234 +196,289 @@ protected void addNotify() {
...
+                        accountTypeDisplayname = accountType;
+                    }
+
+                    return new Node[]{new DefaultAccountTypeNode(key, accountTypeDisplayname)};
+                }
+            } catch (IllegalArgumentException ex) {
+                LOGGER.log(Level.WARNING, "Unknown account type: {0}", key);
+                //Flesh out what happens with other account types here.
+                return new Node[]{new DefaultAccountTypeNode(key, key)};
+            }
```
rcordovano: What is throwing an unchecked IllegalArgumentException that requires us to do this catch?

Core/src/org/sleuthkit/autopsy/datamodel/accounts/Accounts.java
```diff
@@ -1027,74 +1275,85 @@ protected Sheet createSheet() {
...
-                    return new Node[]{new AccountArtifactNode(art)};
-                } catch (TskCoreException ex) {
-                    LOGGER.log(Level.WARNING, "Error creating BlackboardArtifactNode for artifact with ID " + artifactID, ex);   //NON-NLS
-                    return new Node[0];
-                }
+            try {
+                BlackboardArtifact art = skCase.getBlackboardArtifact(artifactID);
+                return new Node[]{new AccountArtifactNode(art)};
+            } catch (TskCoreException ex) {
+                LOGGER.log(Level.WARNING, "Error creating BlackboardArtifactNode for artifact with ID " + artifactID, ex);   //NON-NLS
```
rcordovano: An exception here would be a database query failure and should be logged as SEVERE.

Core/src/org/sleuthkit/autopsy/datamodel/accounts/Accounts.java
raman-bt: I'm not really sure what if anything is throwing the IllegalArgumentException. 

@millmanorama: Jonathan do you recall?

millmanorama: I am not sure.  It looks like nothing throws it now.  Maybe when Account.Type was an enum and if we were looking it up by name or something we threw an exception if we got an unexpected type.  I *think* we can remove the catch block.

raman-bt: That was my thought too Nothing in the current code looked like it could throw it. I took out the catch block and tested it, and of course it worked. But since its a unchecked exception, I wanted to make sure.

PR #3345, opened 2017-12-21
3412: rename index.html to report.html and move html report files to …

zhhl:
…a subdirectory

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
+++ b/Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
@@ -49,6 +49,7 @@ import org.mitre.cybox.objects.WindowsRegistryKey;
 import org.mitre.stix.common_1.IndicatorBaseType;
 import org.mitre.stix.indicator_2.Indicator;
 import org.mitre.stix.stix_1.STIXPackage;
+import org.openide.filesystems.FileUtil;
 import org.openide.util.NbBundle;
 import org.openide.util.NbBundle.Messages;
 import org.sleuthkit.autopsy.casemodule.Case;
@@ -66,6 +67,7 @@ import org.sleuthkit.datamodel.TskCoreException;
 public class STIXReportModule implements GeneralReportModule {
 
     private static final Logger logger = Logger.getLogger(STIXReportModule.class.getName());
+    private static final String STIX_REPORT = "STIX Report";
     private STIXReportModuleConfigPanel configPanel;
     private static STIXReportModule instance = null;
     private String reportPath;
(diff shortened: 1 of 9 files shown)


Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -101,6 +103,13 @@ public void generateReport(String baseReportDir, ReportProgressPanel progressPan
         progressPanel.setIndeterminate(false);
         progressPanel.start();
         progressPanel.updateStatusLabel(NbBundle.getMessage(this.getClass(), "STIXReportModule.progress.readSTIX"));
+        String stixReportDir = baseReportDir + " " + STIX_REPORT;
```
rcordovano: The individual report modules should not modify the base report name passed in to the generateRpeort method because the reports do not "own" this path, so modifications to it could break something in the future. What we should do instead is have the ReportGenerator call ReportModule.getName and insert it into the base report path between the the case name and the time stamp. This means that ReportGenerator will no longer be able to construct the complete report path in its constructor. I would create a format string in the constructor instead, and then use it to make the actual base report path in the four places that it is used. In each of these places, there is a ReportModule that can be used for the call to ReportModule.getName. In connection with this, we should make sure all of our implementations of ReportModule,getName return a reasonable value for use in the base path as an aid to users finding the report they want. It looks like several of the getName implementations return a string that starts with "Results - ", which is not particularly helpful these days, so it can be discarded.

Core/src/org/sleuthkit/autopsy/modules/stix/STIXReportModule.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Autopsy Forensic Browser
  * 
- * Copyright 2013 Basis Technology Corp.
+ * Copyright 2018 Basis Technology Corp.
```
rcordovano: This copyright notice should change to "Copyright 2013-2018 Basis Technology Corp", preserving the original date.

Core/src/org/sleuthkit/autopsy/report/FileReportText.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Autopsy Forensic Browser
  *
- * Copyright 2013 Basis Technology Corp.
+ * Copyright 2018 Basis Technology Corp.
```
rcordovano: This copyright notice should change to "Copyright 2013-2018 Basis Technology Corp", preserving the original date.

Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Autopsy Forensic Browser
  *
- * Copyright 2013 Basis Technology Corp.
+ * Copyright 2018 Basis Technology Corp.
```
rcordovano: This copyright notice should change to "Copyright 2013-2018 Basis Technology Corp", preserving the original date.

Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -145,9 +138,17 @@ public void windowClosing(WindowEvent e) {
...
     void generateGeneralReport(GeneralReportModule generalReportModule) {
         if (generalReportModule != null) {
+            reportPathFormatString = String.format(reportPathFormatString, generalReportModule.getName());
+            // Create the root reports directory.
+            try {
+                FileUtil.createFolder(new File(reportPathFormatString));
+            } catch (IOException ex) {
+                errorList.add(NbBundle.getMessage(this.getClass(), "ReportGenerator.errList.failedMakeRptFolder"));
+                logger.log(Level.SEVERE, "Failed to make report folder, may be unable to generate reports.", ex); //NON-NLS
+            }
```
rcordovano: Moving the folder creation from the constructor to the various methods highlights a problem with this approach.  If there was a problem creating the directory, there is no point in calling into the ReportModule(s). I want to move this code change into the code base, but I will create a follow up story in JIRA for addressing this.

PR #1065, opened 2015-02-19
Stix report generation file validation 2

sidheshenator:

PR #3347, opened 2017-12-28
3208 delay writing to the database until the end of the ingest pipeline

APriestman:
Depends on TSK PR #1065


Core/src/org/sleuthkit/autopsy/ingest/FileIngestPipeline.java
```diff
@@ -136,6 +140,14 @@ Date getStartTime() {
...
                 }
             }
+            
+            // Save any properties that have not already been saved to the database
+            try{
+                file.save(Case.getCurrentCase().getSleuthkitCase());
+            } catch (TskCoreException ex){
+                Logger.getLogger(FileIngestPipeline.class.getName()).log(Level.SEVERE, "Failed to save data", ex); //NON-NLS
+            }
+
```
rcordovano: I think we should do two things here:

1. Move this code inside the job cancellation check conditional. We do not need to do a database round trip if nothing has changed.
2. Add the object id of the file to the error message to make debugging easier.

Core/src/org/sleuthkit/autopsy/modules/filetypeid/FileTypeDetector.java
```diff
@@ -474,7 +483,30 @@ private String detectAutopsyDefinedType(AbstractFile file) throws TskCoreExcepti
...
+     *         were uncertain, octet-stream is returned.
+     *
+     * @throws TskCoreException if detection is required and there is a problem
+     *                          writing the result to the case database.
+     * 
+     * @deprecated
+     */
+    @Deprecated
+    public String getFileType(AbstractFile file) throws TskCoreException {
+        return detect(file, true);
```
rcordovano: I would like us to handle the public API change a bit differently.  Instead of creating a new API with a flag, let's call the new API detectFileType(AbstractFile) and have it call detect(file, false). The module writers do not need to know about our deferred write mechanism.

PR #3350, opened 2018-01-04
3414: Generic Application Content Viewer

raman-bt:


Core/src/org/sleuthkit/autopsy/contentviewers/ApplicationContentViewer.java
```diff
@@ -0,0 +1,251 @@
...
+            return false;
+        } else {
+            return (getSupportingViewer(mimeType) != null);
+        }
+       
+    }
+
+    @Override
+    public int isPreferred(Node node) {
+        if (node == null) {
```
bcarrier: isPreferred() is called only if isSupported() returned true.  So, these checks can be dropped.

Core/src/org/sleuthkit/autopsy/contentviewers/ApplicationContentViewer.java
```diff
@@ -0,0 +1,251 @@
...
+    }// </editor-fold>//GEN-END:initComponents
+
+
+    // Variables declaration - do not modify//GEN-BEGIN:variables
+    // End of variables declaration//GEN-END:variables
+    
+    @Override
+    public void setNode(Node selectedNode) {
+        if (selectedNode == null) {
+            return;
```
bcarrier: reset should be called in this situation.

Core/src/org/sleuthkit/autopsy/contentviewers/ApplicationContentViewer.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Get the ContentViewer for a given mimetype
+     *
+     * @param mimeType
+     *
+     * @return ContentViewer, null if no known content viewer supports the
+     * mimetype
+     */
+    private ContentViewer getSupportingViewer(String mimeType) {
+        return mimeTypeToViewerMap.containsKey(mimeType) ? mimeTypeToViewerMap.get(mimeType) : null;
```
millmanorama: Map.get(key) already returns null if he key is not contained in the map.

PR #3359, opened 2018-01-08
Cvt visualize tab

millmanorama:
revive he Visualize Tab with basic support for pinning from the accounts table

should this go into accounts_relationships?

depends on https://github.com/sleuthkit/sleuthkit/pull/1077


Core/src/org/sleuthkit/autopsy/communications/AccountDeviceInstanceNode.java
```diff
@@ -0,0 +1,132 @@
...
+        return accountDeviceInstanceKey.getAccountDeviceInstance();
+    }
+
+    public AccountDeviceInstanceKey getAccountDeviceInstanceKey() {
+        return accountDeviceInstanceKey;
+    }
+
+    public CommunicationsManager getCommsManager() {
+        return commsManager;
+    }
```
rcordovano: I am not sure I am comfortable with making a node class a source for getting the CommunicationsManager.  Is the motivation for this to avoid having to do Case.getCurrentCase().getSleuthkitCase().getCommunicationsManager(), or are there other reasons?

Core/src/org/sleuthkit/autopsy/communications/AccountDeviceInstanceNode.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public long getMessageCount() {
+        return accountDeviceInstanceKey.getMessageCount();
+    }
+
+    public CommunicationsFilter getFilter() {
+        return accountDeviceInstanceKey.getCommunicationsFilter();
+    }
```
rcordovano: I would give the methods that are not overrides package access, since the class is package access.  That way, if we decide to go public, the developer who does this will need to think about what they want for the public API. This getFilter() method, for example, seems to have only the CommunicationsManager as its client.

Core/src/org/sleuthkit/autopsy/communications/CVTEvents.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.common.eventbus.EventBus;
+
+/**
+ * Provide he singleton EventBus.
+ */
+final class CVTEvents {
+
+    private final static EventBus cvtEventBus = new EventBus();
+
+    public static EventBus getCVTEventBus() {
```
rcordovano: See previous comment about public methods on package access classes. Also, I'm curious, why have the event carry a reference to the event bus?

Core/src/org/sleuthkit/autopsy/communications/AccountsBrowser.java
```diff
@@ -54,19 +67,24 @@ public AccountsBrowser() {
...
             } else if (ExplorerManager.PROP_EXPLORED_CONTEXT.equals(evt.getPropertyName())) {
                 SwingUtilities.invokeLater(this::setColumnWidths);
             }
         });
+
+        jSplitPane1.setRightComponent(new MessageBrowser(tableExplorerManager, messageBrowserEM));
+
+        proxyLookup = new ProxyLookup(
+                ExplorerUtils.createLookup(messageBrowserEM, getActionMap()),
+                ExplorerUtils.createLookup(accountsTableEM, getActionMap()));
```
rcordovano: Very interesting!

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * * A panel that goes in the Visualize tab of the Communications Visualization
+ * Tool. Hosts an JGraphX mxGraphComponent that host the communications network
+ * visualization and a MessageBrowser for viewing details of communications.
+ *
+ * The Lookup provided by getLookup will be proxied by the lookup of the
+ * CVTTopComponent when this tab is active allowing for context sensitive
+ * actions to work correctly.
+ */
+final public class VisualizationPanel extends JPanel implements Lookup.Provider {
```
rcordovano: Does this class need to be public?

Core/src/org/sleuthkit/autopsy/communications/AccountDeviceInstanceNode.java
```diff
@@ -0,0 +1,132 @@
...
+        setIconBaseWithExtension("org/sleuthkit/autopsy/communications/images/" + Utils.getIconFileName(account.getAccountType()));
+    }
+
+    public AccountDeviceInstance getAccountDeviceInstance() {
+        return accountDeviceInstanceKey.getAccountDeviceInstance();
+    }
+
+    public AccountDeviceInstanceKey getAccountDeviceInstanceKey() {
+        return accountDeviceInstanceKey;
+    }
```
rcordovano: I would give the methods that are not overrides package access, since the class is package access.  That way, if we decide to go public, the developer who does this will need to think about what they want for the public API. This method, for example, seems to be something that would only need package access.

Also, I am curious about why a node would need to keep a reference to its key. Could this make key refernces hang around after they could otherwise be garbage collected, i.e., do we understand the way the keys for which we pass of ownership to the NetBeans framework should be managed in memory?

Core/src/org/sleuthkit/autopsy/communications/AccountDeviceInstanceNode.java
millmanorama: that is exactly the motivation.  I particularly don't like calling get current case, as it always introduces a lot of noise to do the proper error handling.  However if you prefer that to the approach of passing he CommunicationsManager around with he nodes, I will change it.

Core/src/org/sleuthkit/autopsy/communications/CVTEvents.java
millmanorama: Yes, I will do a sweep for public methods.  This is not an event, i just provides static access to the eventbus.  Maybe it could go somewhere else...

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
millmanorama: yes for it to work properly in Netbeans/Matisse

Core/src/org/sleuthkit/autopsy/communications/AccountDeviceInstanceNode.java
millmanorama: I suppose the node does not need to keep a reference to its key.  I am no sure we have a complete understanding of how Netbeans manages keys in memory.  @esaunders  might have thoughts since he did most of the work to reduce memory consumption of retained keys and nodes.

PR #3361, opened 2018-01-09
3449: Failed to generate report if couldn't create a report directory.

zhhl:


Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -136,16 +136,9 @@ public void windowClosing(WindowEvent e) {
...
         if (generalReportModule != null) {
-            reportPathFormatString = String.format(reportPathFormatString, generalReportModule.getName());
-            // Create the root reports directory.
-            try {
-                FileUtil.createFolder(new File(reportPathFormatString));
-            } catch (IOException ex) {
-                errorList.add(NbBundle.getMessage(this.getClass(), "ReportGenerator.errList.failedMakeRptFolder"));
-                logger.log(Level.SEVERE, "Failed to make report folder, may be unable to generate reports.", ex); //NON-NLS
-            }
+            reportPathFormatString = createReportDirectory(generalReportModule, reportPathFormatString);
```
rcordovano: It would be better to not overwrite the format string with a formatted path. This assumes that the format string will only ever be used once. It would be better to make reportPathFormatString into a static final String REPORT_PATH_FMT_STR and use a local variable for the return value of createReportDirectory.

Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -194,16 +180,9 @@ void generateTableReport(TableReportModule tableReport, Map<BlackboardArtifact.T
...
         if (fileReportModule != null && null != enabledInfo) {
-            reportPathFormatString = String.format(reportPathFormatString, fileReportModule.getName());
-            // Create the root reports directory.
-            try {
-                FileUtil.createFolder(new File(reportPathFormatString));
-            } catch (IOException ex) {
-                errorList.add(NbBundle.getMessage(this.getClass(), "ReportGenerator.errList.failedMakeRptFolder"));
-                logger.log(Level.SEVERE, "Failed to make report folder, may be unable to generate reports.", ex); //NON-NLS
-            }
+            reportPathFormatString = createReportDirectory(fileReportModule, reportPathFormatString);
```
rcordovano: Please see prior comment.

Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -163,16 +156,9 @@ void generateGeneralReport(GeneralReportModule generalReportModule) {
...
         if (tableReport != null && null != artifactTypeSelections) {
-            reportPathFormatString = String.format(reportPathFormatString, tableReport.getName());
-            // Create the root reports directory.
-            try {
-                FileUtil.createFolder(new File(reportPathFormatString));
-            } catch (IOException ex) {
-                errorList.add(NbBundle.getMessage(this.getClass(), "ReportGenerator.errList.failedMakeRptFolder"));
-                logger.log(Level.SEVERE, "Failed to make report folder, may be unable to generate reports.", ex); //NON-NLS
-            }
+            reportPathFormatString = createReportDirectory(tableReport, reportPathFormatString);
```
rcordovano: Please see prior comment.

PR #3364, opened 2018-01-10
3408 multi user case explorer 2.11.0

wschaeferB:


wschaeferB:
Forgot to add context menu action to open log

Core/src/org/sleuthkit/autopsy/casemodule/Bundle.properties
```diff
@@ -223,17 +223,11 @@ ReviewModeCasePanel.MetadataFileHeaderText=Metadata File
...
-MultiUserCasesPanel.bnRefresh.text=&Refresh
 MultiUserCasesPanel.bnOpen.text=&Open
-MultiUserCasesPanel.rbGroupLabel.text=Show cases created in the last 10:
-MultiUserCasesPanel.rbMonths.text=Months
 CueBannerPanel.newCaseLabel.text=New Case
 CueBannerPanel.openCaseButton.text=
 CueBannerPanel.openCaseLabel.text=Open Case
 MultiUserCasesPanel.bnOpenSingleUserCase.text=Open Single-User Case...
 CueBannerPanel.newCaseButton.text=
+MultiUserCasesPanel.searchLabel.text=Start typing to search by case name
```
dgrove727: Just an initial thought as I run this: It says "Start typing to search by case name", but this doesn't work unless the user first selects a row in the table. It's a little misleading. Perhaps this should be changed to either only show that message when a row has been selected, or figure out a way to get the search working without making that initial selection.

wschaeferB: I will make it auto select the first element unless you see an issue with that idea.  Although they would still be able to prevent it from working by un-selecting the node with ctrl+click.  
Alternatively I would prefer to just make the message longer than to make the message changing based on selection.

rcordovano: I am fine with adding to the help text.

Core/src/org/sleuthkit/autopsy/casemodule/CaseBrowser.java
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Add a listener to changes in case selections in the table
+     *
+     * @param listener the ListSelectionListener to add
+     */
+    void addListSelectionListener(ListSelectionListener listener) {
+        outline.getSelectionModel().addListSelectionListener(listener);
+    }
+
+    String getCasePath() {
```
dgrove727: Missing JavaDoc for this this method.

Core/src/org/sleuthkit/autopsy/casemodule/CaseBrowser.java
```diff
@@ -0,0 +1,296 @@
...
+        int dateColumnIndex = 0;
+        for (int index = 0; index < columnModel.getColumnCount(); index++) {
+            //get indexes for created date column and path column
+            if (columnModel.getColumn(index).getHeaderValue().toString().equals(Bundle.CaseNode_column_metadataFilePath())) {
+                originalPathColumnIndex = index;
+            } else if (columnModel.getColumn(index).getHeaderValue().toString().equals(Bundle.CaseNode_column_createdTime())) {
+                dateColumnIndex = index;
+            }
+        }
+        //Hide path column by default will need to
```
dgrove727: Looks like you started documenting and forgot to finish your thought! =)

PR #3367, opened 2018-01-11
3238: Add a SQLite content viewer: List tables

raman-bt:


markmckinnon:
You should check to see if the SQLite database has a wal file associated with it.  If it does then you need to account for that as well when you write out the SQLite database otherwise you are potentially missing tables and data.

markmckinnon:
@raman-bt  since you need to run some tests I took the liberty to create some files for you that can be found here https://drive.google.com/drive/folders/1TC6C66kwWOWhl8LJmpIbzP0g3pFPFM7R?usp=sharing

The database Wal-test-1.db3 has 2 tables that were create but the table structure and data only exist in the wal file.  Wal-test-2.db3 has 1 table that exists in the database with data and 1 table with data that exists in the wal file.  Wal-test-3.db3 has 2 tables that exist in the database and no wal file.  Wal-test-4.db3 has 2 tables and data existing in the database but the data has been deleted in the wal file so actually no data should exist in the database if you open it up properly.

Main reason I know all this is because I ran into the same issue when I created my SQLite python plugin.

Enjoy the test data.

Mark

raman-bt:
@markmckinnon : THANK YOU! this will be very helpful

markmckinnon:
@raman-bt actually thank you as creating a content viewer for SQLite.  This was on my list of things to create this year and now I can can cross it off my list.

PR #3372, opened 2018-01-16
Updated Autopsy build for new jar files

rishwanth1995:


Core/nbproject/project.properties
```diff
@@ -18,7 +18,7 @@ file.reference.sqlite-jdbc-3.8.11.jar=release/modules/ext/sqlite-jdbc-3.8.11.jar
 file.reference.StixLib.jar=release/modules/ext/StixLib.jar
 file.reference.tika-core-1.14.jar=release/modules/ext/tika-core-1.14.jar
 file.reference.tika-parsers-1.14.jar=release/modules/ext/tika-parsers-1.14.jar
-file.reference.Tsk_DataModel_PostgreSQL.jar=release/modules/ext/Tsk_DataModel_PostgreSQL.jar
```
millmanorama: it seems bad that the version number is hardcoded here.  But maybe there is a plan and I don't know about it.

bcarrier: It's part of getting Autopsy and the TSK JAR files on Linux. The only impact should be that we need to update this file before or after a release.

PR #3379, opened 2018-01-17
3448 need to retain more log files

zhhl:


Core/src/org/sleuthkit/autopsy/corecomponents/AutopsyOptionsPanel.java
```diff
@@ -476,6 +485,24 @@ private boolean isMemFieldValid() {
...
+        logNumAlert.setText("");
+        if (count.length() == 0) {
+            logNumAlert.setText(Bundle.AutopsyOptionsPanel_logNumAlert_noValueEntered_text());
+            return false;
+        } 
+        if (count.replaceAll("[^\\d]", "").length() != count.length() || count.replaceFirst("^0+", "").length() == 0) {
+            logNumAlert.setText(Bundle.AutopsyOptionsPanel_logNumAlert_invalidInput_text());
+            return false;
+        }
+        return true;
```
zhhl: @rcordovano  Here the validation code is similar to isMemFieldValid(). The user interface looks better, because this way the Java NumberFormatException window is not pop up, only show the warning message. But this is not checking if the input is larger than the Integer.MAX_VALUE. 
I like the following way to check if a number is a positive integer, it will include checking empty input, fraction, or number>Integer.MAX_VALUE. 

        String count = logFileCount.getText();
         try {
             int count_num = Integer.parseInt(count);
             if (count_num < 1) {
                 logNumAlert.setText(Bundle.AutopsyOptionsPanel_logFileCountAlert_text());
                 return false;
             } 
         } catch (NumberFormatException e) {
             logNumAlert.setText(Bundle.AutopsyOptionsPanel_logFileCountAlert_text());
             return false;
         }
But this will popup Java Exception window besides show the warning message if the input is invalid. Please let me know which way is better here.

rcordovano: We definitely do not want a pop up as well as a message, but we should still try parsing the int and catching the exception. I am not seeing how the int parsing code can be throwing an exception we can't catch , so please put the code back to this way and show me how the pop up is happening, thanks.

Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
```diff
@@ -369,4 +372,26 @@ public static void setAppName(String name) {
...
+        return preferences.getInt(MAX_NUM_OF_LOG_FILE, LOG_FILE_NUM_INT);
+    }
+    
+    /**
+     * get the default number of log files to save
+     * @return LOG_FILE_COUNT
+     */
+    public static String getDefaultLogFileCount() {
+        return LOG_FILE_NUM_STRING;
+    }
```
rcordovano: Let's return an int here and let clients change it to a string if they want to; it is more natural to return an int.

Core/src/org/sleuthkit/autopsy/contentviewers/ApplicationContentViewer.form
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
rcordovano: It appears that the develop branch has been merged into the branch used for this pull request. We need the changes to the logging in the release-4.6.0 branch, but we need to keep the new application content viewer in the develop branch for eventual inclusion in the 4.7 release. Please address the additional review comments below and add the changed files to a release-4.6.0 branch and resubmit this pull request, thanks.

Core/src/org/sleuthkit/autopsy/coreutils/Logger.java
```diff
@@ -128,13 +96,10 @@ synchronized public static void setLogDirectory(String directoryPath) {
          * add/remove handler calls (currently, the base class handlers
          * collection is a CopyOnWriteArrayList).
          */
-        FileHandler newUserFriendlyHandler = createFileHandlerWithoutTraces(directoryPath);
-        FileHandler newDeveloperFriendlyHandler = createFileHandlerWithTraces(directoryPath);
+        FileHandler newUserFriendlyHandler = createFileHandlerWithTraces(directoryPath);
```
rcordovano: Let's thoroughly complete the removal of the concept of user friendly vs. developer friendly logging from this class - it has only been done partially. Please take a few more minutes to make sure that all variable names, etc. in the oced an all comments have been brought up to date to avoid confusing future maintainers of this file.

PR #3383, opened 2018-01-22
3415: SQLite Viewer -  Show Table Content

raman-bt:


Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -363,33 +410,103 @@ protected Integer doInBackground() throws Exception {
...
+                            + " LIMIT " + Integer.toString(numRowsToRead)
+                            + " OFFSET " + Integer.toString(startRow - 1)
+                    ); //NON-NLS
 
-        SQLColDef(String colName, SQLStorageClass sc) {
-            this.colName = colName;
-            this.storageClass = sc;
+                    return resultSetToArrayList(resultSet);
+                } catch (SQLException ex) {
+                    LOGGER.log(Level.SEVERE, "Failed to get data for table.", ex); //NON-NLS
```
rcordovano: It would be helpful for debugging purposes if we logged more information, i.e., at least the file name and the table name, and perhaps the other arguments as well. My concern is that as it is right now, we would have very little actionable information if we received a report of a problem from production or someone in the open source community.

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -363,33 +410,103 @@ protected Integer doInBackground() throws Exception {
...
+                super.done();
+                try {
+                    ArrayList<Map<String, Object>> rows = get();
+                    if (Objects.nonNull(rows)) {
+                        selectedTableView.setupTable(rows);
+                    }else{
+                        selectedTableView.setupTable(Collections.emptyList());
+                    }
+                } catch (InterruptedException | ExecutionException ex) {
+                    LOGGER.log(Level.SEVERE, "Unexpected exception while reading table.", ex); //NON-NLS
```
rcordovano: As above, the name of the file and the table would be helpful for debugging any issues that might come up when this code is released.

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -363,33 +410,103 @@ protected Integer doInBackground() throws Exception {
...
+        int columns = md.getColumnCount();
+        ArrayList<Map<String, Object>> arraylist = new ArrayList<>();
+        while (rs.next()) {
+            Map<String, Object> row = new LinkedHashMap<>(columns);
+            for (int i = 1; i <= columns; ++i) {
+                if (rs.getObject(i) == null) {
+                    row.put(md.getColumnName(i), "");
+                } else {
+                    if (md.getColumnTypeName(i).compareToIgnoreCase("blob") == 0) {
+                        row.put(md.getColumnName(i), "BLOB Data not shown...");
```
rcordovano: It would be better if the message about the BLOB data was implemented via NbBundle.Messages instead of hard-coded.  This will make it easier/less expensive for localization. Also, I would make the message just "BLOB data not shown", lower-casing "data" and omitting the ellipsis which is generally used when the user can invoke a dialog. I personally find the use of the ellipsis slightly unusual here.

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -363,33 +410,103 @@ protected Integer doInBackground() throws Exception {
...
+                    LOGGER.log(Level.SEVERE, "Unexpected exception while reading table.", ex); //NON-NLS
+                }
+            }
+        };
+
+        worker.execute();
+    }
+
+    private ArrayList<Map<String, Object>> resultSetToArrayList(ResultSet rs) throws SQLException {
+        ResultSetMetaData md = rs.getMetaData();
```
rcordovano: This would be a bit easier to read and more in line with our coding guidelines if we used a longer variable name than "md." I personally experienced having to look back at the variable definition to remind myself what "md" stands for while reading the code, short as it is. The use of "i" as an index and "rs" for a record set is so common the short names work, but I don't think "md" is a common abbreviation.

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -363,33 +410,103 @@ protected Integer doInBackground() throws Exception {
...
+        int columns = md.getColumnCount();
+        ArrayList<Map<String, Object>> arraylist = new ArrayList<>();
```
rcordovano: Renaming "arrayList" to "rowList would help readability.

PR #3389, opened 2018-01-25
3468 artifact hits

dgrove727:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/ExtractedContentViewer.java
```diff
@@ -104,41 +105,55 @@ public void setNode(final Node node) {
...
+            /*
+             * Results for ad-hoc search.
+             */
+            QueryContent queryContent = nodeLookup.lookup(QueryContent.class);
+            content = (AbstractFile) queryContent.getContent();
+            
+            if (null != content && solrHasContent(content.getId())) {
+                QueryResults queryResults = queryContent.getResults();
+                if (queryResults != null) {
+                    highlightedHitText = new HighlightedText(queryContent.getSolrObjectId(), queryResults);
```
rcordovano: I am not comfortable with the logic in this method, for reasons that are detailed in other comments. Please do the following instead, which assumes that we already know there is indexed text from isSupported, so we can eliminate some Solr queries in this method:
`Query lookup for QueryContent`
`Query lookup for BlackboardArtifact`
`Query lookup for AbstractFile`
`If lookup has QueryContent`
`Construct a HighlightedText using the Solr object id and the query results from the QueryContent` 
`else if lookup has a BlackboardArtifact and it is TSK_KEYWORD_HIT`
`Construct a HighlightedText using the artifact`
`else if lookup has a BlackboardArtifact and it is TSK_ACCOUNT and lookup has AbstractFile`
`Construct a HighlightedText using getAccountsText`
`if lookup has AbstractFile`
`Construct a RawText with file and its id`
`If lookup has a BlackboardArtifact`
`Construct a RawText with getRawArtifactText`

Note that in another story, we will seal up the holes in exception handling so that the user will see error messages when we have TskCoreExceptions, but we will consider that out of scope for this effort.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
```diff
@@ -55,7 +54,7 @@
  */
 class HighlightedText implements IndexedText {
 
-    private static final Logger logger = Logger.getLogger(HighlightedText.class.getName());
+    private static final Logger LOGGER = Logger.getLogger(HighlightedText.class.getName());
```
rcordovano: See previous comment on leaving this as "logger" per revised coding guidelines.

PR #3392, opened 2018-01-26
3469 3470 fix Excel and HTML report issues

zhhl:


zhhl:
Not ready yet.

Core/src/org/sleuthkit/autopsy/report/ReportExcel.java
```diff
@@ -34,6 +34,7 @@
 
     private static final Logger logger = Logger.getLogger(ReportExcel.class.getName());
     private static ReportExcel instance;
+    private static final int EXCEL_CELL_MAXIMUM_SIZE = 36767;
```
rcordovano: How did we arrive at this size? I notice that it is not near a power of two, which strikes me as unusual.

rcordovano: Zhaohui says this size is from documentation; she will add a link to the code as a comment.

PR #3393, opened 2018-01-26
3470: Excel report handles the data too large exception and removing …

zhhl:
…tags for non-HMTL preview data.


zhhl:
This is not what we want in our HTML report.

PR #3405, opened 2018-01-31
3445 date filter develop

wschaeferB:


wschaeferB:
Appears to have an issue with backwards compatibility of FileSetsDefinitions, figuring out how to fix it before re-opening PR

wschaeferB:
Serialized properties files aren't a shared resource that we currently support older versions reading settings from newer versions from.

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
```diff
@@ -737,6 +755,46 @@ public boolean passes(AbstractFile file) {
...
+            /**
+             * Construct a new DateCondition
+             *
+             * @param days - files created or modified more recently than this
+             *             number of days will pass
+             */
+            DateCondition(int days) {
+                daysIncluded = days;
+            }
+
```
raman-bt: What are the valid values for daysIncluded?  The UI allows for 0 as well as a negative number.

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
```diff
@@ -737,6 +755,46 @@ public boolean passes(AbstractFile file) {
 
         }
 
+        /**
+         * A class for checking whether a file's creation or modification
+         * occured in a specific range of time
+         */
+        static final class DateCondition implements FileAttributeCondition {
+
+            private final static long DAY_IN_S = 60 * 60 * 24;
```
raman-bt: Maybe rename DAY_IN_S to something more obvious - SECS_PER_DAY?

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
wschaeferB: Modifying it to allow 0 (just in case they want to see if any file has a bogus future modification or creation date) but to not allow negative numbers because they just add a layer of confusion, I can easily disallow 0 as well if that is deemed simpler for the user.

Core/src/org/sleuthkit/autopsy/modules/interestingitems/FilesSet.java
wschaeferB: Good suggestion, thanks.

rcordovano:
@wschaeferB and @raman-bt, it looks like there is no code to export the new rule condition to XML. Will, please add this; Raman, please review again and re-test, being sure to verify that everything still works for both interesting file sets and file ingest filters, including both binary serialization and XML persistence of settings file, thanks!

wschaeferB:
I have pushed the fix to write the conditions to the xml file when exporting, apologies for missing that part.

rcordovano:
@raman-bt, thanks. Did you also test import of something you exported?

raman-bt:
Yes, I did import an exported filter.

PR #3409, opened 2018-02-01
3284 disk image block size

dgrove727:


Core/src/org/sleuthkit/autopsy/casemodule/AddImageTask.java
```diff
@@ -126,9 +129,6 @@ public void run() {
             Thread progressUpdateThread = new Thread(new ProgressUpdater(progressMonitor, tskAddImageProcess));
             progressUpdateThread.start();
             runAddImageProcess(errorMessages);
-            if (null != progressUpdateThread) {
-                progressUpdateThread.interrupt();
-            }
```
APriestman: Why was this removed? It seems unrelated to sector size. Was it broken?

dgrove727: The reason I took this out was because, as the IDE indicates, 'progressUpdateThread' is never going to be null at this point. But in re-examining this again, I realize now that the thread interrupt should still be there, because the condition check is for *not* null. So I'll get a fix for this.

PR #3414, opened 2018-02-02
PMD: Testing the properties setting for comments rule.

zhhl:


bcarrier:
Applied as part of #3417.

PR #3430, opened 2018-02-09
Cvt 950 overide layout

millmanorama:
- locked nodes are ignored by layout algorithms
- scale labels and icons with vertex size
- improve fitting the graph to the view area.
- progress related cleanup (in preparation for cancelling)


Core/src/org/sleuthkit/autopsy/communications/FiltersPanel.java
```diff
@@ -196,8 +197,8 @@ private void updateAccountTypeFilter() {
         Account.Type.PREDEFINED_ACCOUNT_TYPES.forEach(type -> {
             if (type.equals(Account.Type.CREDIT_CARD)) {
                 //don't show a check box for credit cards
-            } else if (type.equals(Account.Type.DEVICE)) {
-                //don't show a check box fro device
+//            } else if (type.equals(Account.Type.DEVICE)) {
+//                //don't show a check box fro device
```
rcordovano: Is this just a temporary removal? If not, I think it would be better to delete rather than comment out.

millmanorama: removed

PR #3438, opened 2018-02-14
 modified look and feel to make timeline work on autopsy-linux 

rishwanth1995:
lookandfeel is set to crossplatform instead of native look and feel.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
+++ b/Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
@@ -31,6 +31,7 @@ import org.openide.modules.ModuleInstall;
 import org.openide.windows.WindowManager;
 import org.sleuthkit.autopsy.casemodule.StartupWindowProvider;
 import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.autopsy.coreutils.ModuleSettings;
 
 /**
  * Manages this module's life cycle. Opens the startup dialog during startup.
@@ -40,7 +41,6 @@ public class Installer extends ModuleInstall {
     private static final long serialVersionUID = 1L;
     private static final Logger logger = Logger.getLogger(Installer.class.getName());
     private static Installer instance;
-    public static boolean enableTimeline = true;
 
     public synchronized static Installer getDefault() {
         if (null == instance) {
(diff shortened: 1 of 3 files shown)


Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -111,4 +114,13 @@ private void setOSXLookAndFeel() {
             UIManager.put(entry.getKey(), entry.getValue());
         });
     }
+    
+    private void setUnixLookAndFeel(){
+        try {
+            //UIManager.setLookAndFeel(UIManager.getCrossPlatformLookAndFeelClassName());
```
rcordovano: Why is there a duplicate line, commented out?

Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -111,4 +114,13 @@ private void setOSXLookAndFeel() {
...
+            UIManager.setLookAndFeel(UIManager.getCrossPlatformLookAndFeelClassName());
+        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | UnsupportedLookAndFeelException ex) {
+            Exceptions.printStackTrace(ex);
```
rcordovano: Please log the error, don't print it to stdout. 

What happens to the LAF if an exception is thrown here?

Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -28,6 +28,7 @@
 import javax.swing.UnsupportedLookAndFeelException;
 import org.netbeans.swing.tabcontrol.plaf.DefaultTabbedContainerUI;
 import org.openide.modules.ModuleInstall;
+import org.openide.util.Exceptions;
```
rcordovano: This import is not needed, see below.

Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -40,6 +40,7 @@
     private static final long serialVersionUID = 1L;
     private static final Logger logger = Logger.getLogger(Installer.class.getName());
     private static Installer instance;
+    public static boolean enableTimeline = true;
```
rcordovano: Our coding guidelines prohibit global variables. Please put this value in a properties file for Timeline to read. There is a ModuleSettings utility class that could be used for managing the properties file.

Note that you will need to reset the property to enabled at the beginning of this set LAF operation.

Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -111,4 +114,18 @@ private void setOSXLookAndFeel() {
...
+        try {
+            UIManager.setLookAndFeel(UIManager.getCrossPlatformLookAndFeelClassName());
+        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | UnsupportedLookAndFeelException ex) {
+            logger.log(Level.WARNING, "Error setting crossplatform look-and-feel, setting default look-and-feel",ex);
+            try {
+                UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+                enableTimeline=false;
+            } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | UnsupportedLookAndFeelException ex1) {
+                 logger.log(Level.WARNING, "Error setting native look-and-feel",ex1);
+            }
```
rcordovano: Given that you say that the default system LAF will be used anyway, I think we should reduce code complexity here (i.e., eliminate the nested try-catch) and simply set the property to false after logging the error.

Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
bcarrier: I think we can solve that in the next release.  This is good enough for now.

PR #3442, opened 2018-02-15
2229: Make Case.getCurrentCase throw checked exceptions, part 1

zhhl:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -597,13 +598,32 @@ public static Case getCurrentCase() {
...
+     * @return The open case.
+     *
+     * @throws NoCurrentCaseException if there is no open case.
+     */
+    public static Case getOpenCase() throws NoCurrentCaseException {
+        try {
+            Case openCase = new Case(currentCase.getCaseType(), 
+                                currentCase.getCacheDirectory(), 
+                                new CaseDetails(currentCase.getDisplayName(), currentCase.getNumber(), currentCase.getExaminer(), currentCase.getExaminerPhone(), currentCase.getExaminerEmail(), currentCase.getCaseNotes()));
+            return openCase;
```
rcordovano: For this fix, we should be making a local copy of the reference in currentCase, checking the local reference for null, and finally returning the local reference. This does not prevent another thread from closing the case, but it does make things a bit better because we will at least not risk returning null and violating the contract of the method. Dealing with the possibility of the current case being closed while other objects have a reference to it is a much bigger problem for another story.

Core/src/org/sleuthkit/autopsy/casemodule/events/BlackBoardArtifactTagAddedEvent.java
```diff
@@ -45,7 +46,7 @@ public BlackBoardArtifactTagAddedEvent(BlackboardArtifactTag newTag) {
      * @throws TskCoreException
      */
     @Override
-    BlackboardArtifactTag getTagByID() throws IllegalStateException, TskCoreException {
-        return Case.getCurrentCase().getServices().getTagsManager().getBlackboardArtifactTagByTagID(getTagID());
+    BlackboardArtifactTag getTagByID() throws NoCurrentCaseException, TskCoreException {
+        return Case.getOpenCase().getServices().getTagsManager().getBlackboardArtifactTagByTagID(getTagID());
```
rcordovano: This is the right thing to do, but please include the change to the client (TagAddedEvent) in this pull request, so that we keep the incremental changes together.

PR #3445, opened 2018-02-16
3532: Add a html file to include the links to case output directory.

zhhl:


test/script/regression.py
```diff
@@ -106,6 +108,10 @@ def main():
     if not parse_result:
         Errors.print_error("The arguments were given wrong")
         exit(1)
+    # Remove the output_link_file
+    if file_exists(OUTPUT_DIR_LINK_FILE):
+        os.remove(OUTPUT_DIR_LINK_FILE)
+
```
rcordovano: Why would there be a link file to remove? These files are specific to a given test run, are they not stored in the output directory for the run?

PR #3448, opened 2018-02-19
Add URL to reports and Launch when opened

markmckinnon:
Add a URL to a report.  When the report is opened then open it in a browser.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/casemodule/Case.java
+++ b/Core/src/org/sleuthkit/autopsy/casemodule/Case.java
@@ -1529,7 +1529,11 @@ public class Case {
     public void addReport(String localPath, String srcModuleName, String reportName) throws TskCoreException {
         String normalizedLocalPath;
         try {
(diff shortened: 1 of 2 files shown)


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1529,7 +1529,11 @@ public void notifyBlackBoardArtifactTagDeleted(BlackboardArtifactTag deletedTag)
     public void addReport(String localPath, String srcModuleName, String reportName) throws TskCoreException {
         String normalizedLocalPath;
         try {
-            normalizedLocalPath = Paths.get(localPath).normalize().toString();
+            if (localPath.toLowerCase().contains("http:")) {
```
bcarrier: should use startsWith.

bcarrier:
Closing this in favor of the approach in url_viewer.

PR #3453, opened 2018-02-21
3454 auto ingest deprioritize jobs

dgrove727:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -530,7 +546,49 @@ void resume() {
...
+     *
+     * @throws AutoIngestManagerException If there is an error removing the
+     *                                    priority of the jobs for the case.
      */
+    void deprioritizeCase(final String caseName) throws AutoIngestManagerException {
+        if (state != State.RUNNING) {
+            return;
+        }
+
+        List<AutoIngestJob> prioritizedJobs = new ArrayList<>();
```
APriestman: I feel like naming this (and the other similar objects in the other methods) something like "jobsToDeprioritize" would make the code clearer.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestCaseDeprioritizedEvent.java
```diff
@@ -0,0 +1,65 @@
...
+package org.sleuthkit.autopsy.experimental.autoingest;
+
+import java.io.Serializable;
+import org.sleuthkit.autopsy.events.AutopsyEvent;
+
+/**
+ * Event published when an automated ingest manager deprioritizes all or part of
+ * a case.
+ */
+final class AutoIngestCaseDeprioritizedEvent extends AutopsyEvent implements Serializable {
```
APriestman: It looks like you forgot to add event handlers for these new events in AutoIngestManager.propertyChange() and AutoIngestMonitor.propertyChange(). But I would lean towards re-using AutoIngestCasePrioritizedEvent instead of making a new event in the first place, mostly because it will reduce the chance of errors like this. The handling of the two events is exactly the same and I can't see a case where a remote node would care whether a case was prioritized or de-prioritized; either way we need to rescan the directory.

PR #3456, opened 2018-02-22
added setlookandfeel to AWT-EventQueue and made the font look consistent

rishwanth1995:
I named the metallookandfeel class explicitly in line 144 because UIManager.getCrossPlatformLookAndFeel might not be consistent on all distros. I wanted the LAF to be metal on  all linux distros as we planned.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
+++ b/Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
@@ -18,11 +18,13 @@
  */
 package org.sleuthkit.autopsy.corecomponents;
 
+import java.awt.Font;
 import java.awt.Insets;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.logging.Level;
 import javax.swing.BorderFactory;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/corecomponents/Installer.java
```diff
@@ -115,6 +119,16 @@ private void setOSXLookAndFeel() {
...
+    public static void setUIFont (javax.swing.plaf.FontUIResource f){
+    java.util.Enumeration keys = UIManager.getLookAndFeelDefaults().keys();
+    while (keys.hasMoreElements()) {
+      Object key = keys.nextElement();
+      Object value = UIManager.getLookAndFeelDefaults().get(key);
+      if (value instanceof javax.swing.plaf.FontUIResource)
+        UIManager.getLookAndFeelDefaults().put(key, f);
+      }
+    } 
+
```
rcordovano: We are not ready to commit to this approach. Please make a PR without setUIFont, but keep the code with setUIFont in another branch until we decide whether or not we want it.

PR #3457, opened 2018-02-23
3544: Integrate Media content viewer into Application content viewer

raman-bt:


rcordovano:
This does not auto-merge, please address. 

Please take a few minutes to address the Codacy "No need to import a type that lives in the same package" issues.

raman-bt:
I've addressed the merge conflict, as well as the "No need to import a type that lives in the same package"  issues.  

Codacy still complains about other issues in these moved/refactored files.

PR #3460, opened 2018-02-23
3545 local disk image block size

dgrove727:


rcordovano:
Please address Codacy issues.

dgrove727:
The two things Codacy points out are both generated code. Do you still want them changed?

rcordovano:
I did not notice that the Codacy issues were in generated code. Don't worry about them.

PR #3464, opened 2018-02-26
872-fix accout type icons in the DirectoryTree

millmanorama:


rcordovano:
Does not auto-merge. Please address, thanks!

PR #3472, opened 2018-02-27
2229 part1get open case throws checked exception

zhhl:


Core/src/org/sleuthkit/autopsy/casemodule/NoCurrentCaseException.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ *
+ * Exception thrown when no current case is available
+ */
+@NbBundle.Messages({
+        "NoCurrentCaseException.message=No open case is available."
+    })
+public class NoCurrentCaseException extends Exception {
+
+    NoCurrentCaseException() {
```
rcordovano: Instead of providing a no arguments constructor, please provide one constructor with a message parameter, and one constructor with a message parameter and a cause parameter. You will find plenty of examples in other Autopsy code.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -583,24 +583,41 @@ public static boolean isCaseOpen() {
     }
 
     /**
+     * Deprecated. Use getOpenCase() instead.
+     *
```
rcordovano: This is not the correct way to comment on deprecation. Please use the @deprecated comment instead. There should be an example of this for other deprecated Case methods in this file.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -583,24 +583,41 @@ public static boolean isCaseOpen() {
...
+     * @throws NoCurrentCaseException if there is no open case.
+     */
+    public static Case getOpenCase() throws NoCurrentCaseException {
+        Case openCase = currentCase;
+        if (null != openCase) {
+            return openCase;
         } else {
-            throw new IllegalStateException(NbBundle.getMessage(Case.class, "Case.getCurCase.exception.noneOpen"));
+            throw new NoCurrentCaseException();
```
rcordovano: It is not a good idea to throw an exception with no message, e.g., it is awkward when exceptions are logged. See the comment below on the constructors this exception class should have.

zhhl:
Auto merged code shouldn't be included in this pull request.

PR #3478, opened 2018-02-27
2229: Part 1: Use getOpenCase() instead of getCurrentCase() and throws checke…

zhhl:
…d excepiton: NoCurrentCaseException.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/casemodule/Case.java
+++ b/Core/src/org/sleuthkit/autopsy/casemodule/Case.java
@@ -598,8 +598,7 @@ public class Case {
          *
          */
         try {
-            Case curCase = getOpenCase();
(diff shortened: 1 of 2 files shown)


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -583,24 +583,43 @@ public static boolean isCaseOpen() {
     }
 
     /**
+     * Deprecated. Use getOpenCase() instead.
+     *
```
rcordovano: As noted previously, this is not the correct way to document a deprecated method. Look, for example, at Case.addEventSubscriber.

rcordovano:
It looks like the Codacy issue can be easily fixed.

PR #3479, opened 2018-02-27
2229 part2 use getopencase in package action

zhhl:


Core/src/org/sleuthkit/autopsy/actions/DeleteFileBlackboardArtifactTagAction.java
```diff
@@ -95,7 +96,16 @@ protected void deleteTag(TagName tagName, BlackboardArtifactTag artifactTag, lon
...
+                TagsManager tagsManager;
+                try {
+                    tagsManager = Case.getOpenCase().getServices().getTagsManager();
+                } catch (NoCurrentCaseException ex) {
+                    logger.log(Level.SEVERE, "Error untagging artifact. No open case found.", ex); //NON-NLS
+                    Platform.runLater(()
+                            -> new Alert(Alert.AlertType.ERROR, Bundle.DeleteFileBlackboardArtifactTagAction_deleteTag_alert(artifactId)).show()
+                    );
+                    return null;  
+                }
```
rcordovano: There is an error in this code that we should fix, rather than repeat. The doInBackgroundMethod should just throw exceptions, rather than logging them and scheduling a task for the EDT to present the error to the user. This is why we have a done() method - to call get() to catch and handle exceptions thrown by doInBackground and to do things that need to be done in the EDT. Please eliminate the catches in doInBackground and put all of the exception handling in done(). Note that this may lead to unused imports to remove.

Core/src/org/sleuthkit/autopsy/actions/DeleteFileContentTagAction.java
```diff
@@ -95,7 +96,16 @@ protected void deleteTag(TagName tagName, ContentTag contentTag, long fileId) {
...
+                TagsManager tagsManager;
+                try {
+                    tagsManager = Case.getOpenCase().getServices().getTagsManager();
+                } catch (NoCurrentCaseException ex) {
+                    logger.log(Level.SEVERE, "Error untagging file. No open case found.", ex); //NON-NLS
+                    Platform.runLater(() ->
+                            new Alert(Alert.AlertType.ERROR, Bundle.DeleteFileContentTagAction_deleteTag_alert(fileId)).show()
+                    );
+                    return null;
+                }
```
rcordovano: See previous comment regarding use of SwingWorker.doInBackground() and done().

PR #3487, opened 2018-02-28
3568: error handling in FileTypeViewers

raman-bt:


Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -446,17 +467,26 @@ private void selectTable(String tableName) {
...
-                    Statement statement = connection.createStatement();
-                    ResultSet resultSet = statement.executeQuery(
+                    statement = connection.createStatement();
+                    resultSet = statement.executeQuery(
                             "SELECT count (*) as count FROM " + tableName); //NON-NLS
 
                     return resultSet.getInt("count");
                 } catch (SQLException ex) {
-                    LOGGER.log(Level.SEVERE, "Failed to get data for table.", ex); //NON-NLS
+                    LOGGER.log(Level.SEVERE, "Failed to get row count for table " + tableName, ex); //NON-NLS
```
rcordovano: Isn't this exception getting logged in done(), making this log message a duplicate?

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
```diff
@@ -414,28 +435,28 @@ private boolean findAndCopySQLiteMetaFile(AbstractFile sqliteFile, String metaFi
...
 
-                dbTablesMap.put(tableName, tableSQL);
-            }
-        } catch (SQLException e) {
-            LOGGER.log(Level.SEVERE, "Error getting table names from the DB", e); //NON-NLS
+            dbTablesMap.put(tableName, tableSQL);
         }
-        return true;
+        resultSet.close();
+	statement.close();
```
rcordovano: I think a finally block is needed here to ensure that the record set and the statement are always closed.

raman-bt: Addressed.

Core/src/org/sleuthkit/autopsy/contentviewers/SQLiteViewer.java
raman-bt: In this case, yes. So i removed the duplicate log. 

In general, I prefer to log a more specific error message as close to the point of occurrence as possible, since more for the context can be captured and logged.

rcordovano: I would prefer to leave logging to the top-level clients as part of their exception handling. What we can do is propagate info to the clients using exceptions, because I am all for putting as much context into the logs as possible for debugging. So if need be, we can define specific exception class to add context to wrap lower level exceptions like SQLException. The specific exception class names themselves can carry data, and we can construct them with informative, context-rich messages. To sum up: low level classes should throw information dense exceptions so that the high-level clients can decide how to handle errors based on the client use case, and can do the logging, so it is in one place. Multiple logs are confusing.

PR #3490, opened 2018-02-28
3437 read content input stream error 3

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/encryptiondetection/EncryptionDetectionFileIngestModule.java
```diff
@@ -228,6 +232,8 @@ private boolean isFileEncrypted(AbstractFile file) throws IOException, TskCoreEx
                 if (calculatedEntropy >= minimumEntropy) {
                     return true;
                 }
+            } catch (ReadContentInputStreamException ex) {
+                throw new ReadContentInputStreamException("Unable to calculate the entropy.", ex);
```
rcordovano: This error message does not really go well with this exception. Also, we are wrapping a  ReadContentInputStreamException in a ReadContentInputStreamException or an IOException in an IOException just because of how the code has been decomposed into methods.  Something similar happens below. I don't think these catches are necessary.

dgrove727: I was trying to get granular with the messages to help with troubleshooting, but as you have pointed out, the code is a little awkward with the double-wrapping of exceptions. The latest update to this pull request gets rid of those.

PR #3503, opened 2018-03-02
3578: Refactor ContentViewer

raman-bt:


raman-bt:
Deleting this PR, we decided *not *to move things back into CoreComponents to avoid bloating.

PR #3522, opened 2018-03-07
3537 correlation engine settings panel

dgrove727:


Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
```diff
@@ -181,6 +197,19 @@ public void shutDown() {
...
+         * Tell the IngestEventsListener to flag notable items based on the
+         * current module's configuration. This is a work around for the lack of
+         * an artifacts pipeline. Note that this can be changed by another
+         * module instance. All modules are affected by the value. While not
+         * ideal, this will be good enough until a better solution can be
+         * posited.
+         */
+        if (IngestEventsListener.getCeModuleInstanceCount() == 1 || !IngestEventsListener.isFlagNotableItems()) {
+            IngestEventsListener.setFlagNotableItems(flagTaggedNotableItems);
+        }
```
APriestman: Can you add a bit to this comment explaining our algorithm for when we allow the flag to change?

dgrove727: How about...

         * Note: Flagging cannot be disabled by any module instance except for
         * the first instance. This restriction is to prevent a confusing
         * collection of results where the one module is flagging notable items,
         * and the proceeding module causes the first to no longer flagging.

I think that sums it up pretty nicely, right?

PR #3527, opened 2018-03-08
added setup script to copy the sleuthkit-jar to autopsy folder

rishwanth1995:


build.xml
```diff
@@ -87,6 +87,12 @@
         <copy file="${basedir}/LICENSE-2.0.txt" tofile="${zip-tmp}/${app.name}/LICENSE-2.0.txt"/>
         <copy file="${basedir}/NEWS.txt" tofile="${zip-tmp}/${app.name}/NEWS.txt"/>
         <copy file="${basedir}/KNOWN_ISSUES.txt" tofile="${zip-tmp}/${app.name}/KNOWN_ISSUES.txt"/>
+        <if>
+            <equals arg1="${os.family}" arg2="unix"/>
+            <then>
+                <copy file="${basedir}/unix_setup.sh" tofile="${zip-tmp}/${app.name}/unix_setup.sh"/>
+            </then>
```
bcarrier: Do the replaceregexp here, on the copy. Then this file wont' be constantly changing and needing to be checked in each time we change the version.

Core/build.xml
```diff
@@ -6,7 +6,8 @@
     <description>Builds, tests, and runs the project org.sleuthkit.autopsy.core</description>
     <import file="nbproject/build-impl.xml"/>
     <import file="../BootstrapIvy.xml"/>
-    
+    <import file="../TSK_VERSION.xml"/>
```
bcarrier: I'd call it TSKVersion.xml  for consistency.

unix_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+    	echo "deleting $ext_jar_filepath"
+    	rm $ext_jar_filepath;
+    	if [ "$?" -gt 0 ]; then  #checking if remove operation failed
+        	echo "exiting .."
+        	exit 1
+    	else
+        	echo "Successfully removed $ext_jar_filepath"
+        	cp $sleuthkit_jar_filepath $ext_jar_filepath
+        	if [ "$?" -eq 0 ]; then # checking copy operation was successful
+            	echo "Successfully copied $sleuthkit_jar_filepath"
```
bcarrier: Can you indent this and the else?

build.xml
bcarrier: Move this one more line down.  Do the replace on the zip-tmp destination

PR #3528, opened 2018-03-08
3563 application options panel scroll bars

dgrove727:


Core/src/org/sleuthkit/autopsy/corecomponents/Bundle.properties
```diff
@@ -169,3 +169,13 @@ CriterionChooser.removeButton.text=Remove
 CriterionChooser.descendingRadio.text=\u25bc Descending
 AutopsyOptionsPanel.agencyLogoPathFieldValidationLabel.text=
 AutopsyOptionsPanel.logNumAlert.text=
+AutopsyOptionsPanel.runtimePanel.AccessibleContext.accessibleName=<AutopsyOptionsPanel.runtimePanel.border.title>
```
wschaeferB: Using a property to point at another property seems to add an unnecessary layer of complexity to these properties to me.  If possible we should probably have either the accessibleName property or the title property but not both and it should be the text we want to display.  This comment also applies to the AutopsyOptionsPanel.agencyLogoPreview.AccessibleContext.accessibleName property below.

dgrove727: This is an area where NetBeans doesn't quite work right. The IDE doesn't actually let me change these values you refer to, and because it's auto-generated, I ultimately left it alone. I can manually update this stuff, but that seems slightly dangerous to me.

wschaeferB: I agree you shouldn't manually modify auto-generated code but I believe you can change what is used to auto-generate the code through the properties menu, I will post screen shots in the Jira story of what I am thinking.

dgrove727:
The bundle message references should be better now. I do not see any more double references, and it shows up as expected in the editor now too.

PR #3537, opened 2018-03-12
2229 Part 27 use getOpenCase() instead of getCurrentCase()

zhhl:


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/MboxParser.java
```diff
@@ -267,8 +268,18 @@ private void handleTextBody(EmailMessage email, TextBody tb, String type, List<F
...
     private void handleAttachment(EmailMessage email, Entity e, long fileID, int index) {
-        String outputDirPath = ThunderbirdMboxFileIngestModule.getModuleOutputPath() + File.separator;
+        String outputDirPath;
+        String relModuleOutputPath;
+        try {
+            outputDirPath = ThunderbirdMboxFileIngestModule.getModuleOutputPath() + File.separator;
+            relModuleOutputPath = ThunderbirdMboxFileIngestModule.getRelModuleOutputPath() + File.separator;
+        } catch (NoCurrentCaseException ex) {
+            addErrorMessage(Bundle.MboxParser_handleAttch_noOpenCase_errMsg());
+            logger.log(Level.INFO, Bundle.MboxParser_handleAttch_noOpenCase_errMsg(), ex); //NON-NLS
```
rcordovano: This should be logged as a SEVERE error, not an INFO message.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/PstParser.java
```diff
@@ -237,7 +245,7 @@ private void extractAttachments(EmailMessage email, PSTMessage msg, long fileID)
                 attachment.setSize(attach.getFilesize());
                 attachment.setEncodingType(TskData.EncodingType.XOR1);
                 email.addAttachment(attachment);
-            } catch (PSTException | IOException | NullPointerException ex) {
+            } catch (PSTException | IOException | NullPointerException | NoCurrentCaseException ex) {
```
rcordovano: The NoCurrentCaseException should be in a separate catch so the error can be logged as SEVERE.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -76,13 +77,22 @@
     @Override
     public void startUp(IngestJobContext context) throws IngestModuleException {
         this.context = context;
-        fileManager = Case.getCurrentCase().getServices().getFileManager();
+        try {
+            fileManager = Case.getOpenCase().getServices().getFileManager();
+        } catch (NoCurrentCaseException ex) {
+            logger.log(Level.SEVERE, "Exception while getting open case.", ex);
+        }
```
rcordovano: Ingest modules need to throw IngestModuleException if there is an error in startUp, with a user readable exception message, since it is displayed to the user. You need to wrap the NoCurrentCaseException in an IngestModuleException and throw, not log here.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -408,11 +443,19 @@ private BlackboardArtifact addArtifact(EmailMessage email, AbstractFile abstract
...
-        AccountFileInstance senderAccountInstance = null;        
+        AccountFileInstance senderAccountInstance = null;
+
+        Case openCase;
+        try {
+            openCase = Case.getOpenCase();
+        } catch (NoCurrentCaseException ex) {
+            logger.log(Level.WARNING, "Exception while getting open case.", ex); //NON-NLS
+            return null; 
+        }
```
rcordovano: The addArtifact method is called in a loop. The exception needs to propagate, to stop processing. If the case has been closed, there is nothing more for this module to do, and continuing to loop does no good.

Core/src/org/sleuthkit/autopsy/directorytree/ExplorerNodeActionVisitor.java
```diff
@@ -23,12 +23,15 @@
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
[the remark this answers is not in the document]
zhhl: I have checked all other 26 PRs, and they don't have this mistake there.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -76,13 +77,23 @@
     @Override
     public void startUp(IngestJobContext context) throws IngestModuleException {
         this.context = context;
-        fileManager = Case.getCurrentCase().getServices().getFileManager();
+        try {
+            fileManager = Case.getOpenCase().getServices().getFileManager();
+        } catch (NoCurrentCaseException ex) {
+            logger.log(Level.SEVERE, "Exception while getting open case.", ex);
+            throw new IngestModuleException("Exception while getting open case.", ex);
```
rcordovano: IngestModuleExceptions messages should be created with Message annotations since they are displayed to users.

Core/src/org/sleuthkit/autopsy/directorytree/ExtractUnallocAction.java
```diff
@@ -69,14 +69,14 @@
     private long currentImage = 0L;
     private final boolean isImage;
 
-    public ExtractUnallocAction(String title, Volume volume) {
+    public ExtractUnallocAction(String title, Volume volume) throws NoCurrentCaseException {
```
rcordovano: Thinking about this some more, rather than propagate the NoCurrentCaseException from the constructors, lets catch and log it and cal setEnabled(false) on the action so it would be grayed out, but not omitted from the node context menu.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Extract.java
```diff
@@ -41,17 +41,23 @@
...
-    Extract() {
+    Extract() {        
     }
 
     void init() throws IngestModuleException {
+        try {
+            currentCase = Case.getOpenCase();
+            tskCase = currentCase.getSleuthkitCase();
+        } catch (NoCurrentCaseException ex) {
+            throw new IngestModuleException("Exception while getting open case.", ex);
```
rcordovano: IngestModuleExceptions messages should be created with Message annotations since they are displayed to users.

PR #3540, opened 2018-03-13
Added checks for better results in plugins

markmckinnon:
Added checks to get better results in dlllist, cmdline and handles

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
+++ b/Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
@@ -248,6 +248,10 @@ class VolatilityProcessor {
                 // if there is already the same entry with ".exe" in the set, just use that one
                 if (fileSet.contains(file + ".exe"))
                     continue;
+                // if plugin is handles then skip if filename does not have an extension helps with 
+                // cases when there really is no just a directory or if it truly does not have an extension
+                if (pluginName.matches("handles"))
+                    continue;
                 fileName = fileName + ".%";
             }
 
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
```diff
@@ -336,7 +339,13 @@ private String normalizePath(String filePath) {
         filePath = filePath.replaceAll("\\\\", "/");
         filePath = filePath.toLowerCase();
         filePath = filePath.replaceAll("/systemroot/", "/windows/");
+        // catches 1 type of file in cmdline
+        filePath = filePath.replaceAll("%systemroot%", "/windows/");
         filePath = filePath.replaceAll("device/","");
+        // helps with finding files in handles plugin
+        if (filePath.contains("harddiskvolume[0-9]/")){
+            filePath = filePath.substring(filePath.indexOf("harddiskvolume[0-9]/") -1);
```
bcarrier: Isn't this the same as the next line?

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
```diff
@@ -229,8 +229,8 @@ private void lookupFiles(Set<String> fileSet, String pluginName) {
             if (fileName.length() < 1) {
                 continue;
             }
-           // if there is no extension, add a wildcard to the end
-            if (fileName.contains(".") == false) {
+           // if there is no extension, add a wildcard to the end except for handles as it leads to duplicates and missed files
+            if (fileName.contains(".") == false && pluginName != "handles") {
```
bcarrier: Do you have examples of what was causing problems in handles? Should they be filtered out?

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
```diff
@@ -247,29 +247,32 @@ private void lookupFiles(Set<String> fileSet, String pluginName) {
...
-                        BlackboardArtifact volArtifact = resolvedFile.newArtifact(BlackboardArtifact.ARTIFACT_TYPE.TSK_INTERESTING_FILE_HIT);
-                        BlackboardAttribute att1 = new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SET_NAME, MODULE_NAME,
-                                "Volatility Plugin " + pluginName);
-                        BlackboardAttribute att2 = new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_COMMENT, MODULE_NAME,
-                                "Volatility Plugin " + pluginName);
-                        volArtifact.addAttribute(att1);
-                        volArtifact.addAttribute(att2);
-
+                    // check for slack files and do not include them may double the list of files found.
+                    if (!resolvedFile.getName().endsWith("-slack")){
```
bcarrier: If you merge with upstream, then you'll get my version of this fix which is to check the file type.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
markmckinnon: Users\Public\Desktop pulls back Desktop.ini
System Volume Information\{e15d4b08-1598-11e8-93e6-080027b5e733}{3808876b-c176-4e48-b7ae-04046e6cc752} is not found 

Several others that do not have file extensions were not found as well.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
markmckinnon: Actually after looking at it again get rid of the if and the line after the if and keep the substring as this will be better for this case \Device\clfs\Device\HarddiskVolume2\Users\mark\AppData\Local\Microsoft\Windows\UsrClass.dat{e15d4b01-1598-11e8-93e6-080027b5e733}.TM

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
markmckinnon: Will do this and fix the above.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
bcarrier: Ah.  Maybe this logic should then first try to lookup the passed in name w/out the wildcard and then we only add it if it failed to find anything.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/VolatilityProcessor.java
bcarrier: OK.  Can you get rid of the next line then?

PR #3558, opened 2018-03-14
3362 central repo options job completed events

dgrove727:


Core/src/org/sleuthkit/autopsy/centralrepository/optionspanel/GlobalSettingsPanel.java
```diff
@@ -450,7 +451,12 @@ protected void finalize() throws Throwable {
         @Override
         public void propertyChange(PropertyChangeEvent event) {
             if (AutopsyEvent.SourceType.LOCAL == ((AutopsyEvent) event).getSourceType()) {
-                ingestStateUpdated();
+                EventQueue.invokeLater(new Runnable() {
```
rcordovano: This is only a partial fix. This code is responding to all local events. It needs to enable and disable for ingest jobs. Take a look at HashLookupSettingsPanel, which does this right.

PR #3566, opened 2018-03-15
3620 Use device ID instead of data source name in CorrelationAttributeInst…

APriestman:
…ance.toString().
This will fix an issue where similar central repository artifacts were being discarded.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
@@ -111,7 +111,7 @@ public class CorrelationAttributeInstance implements Serializable {
     public String toString() {
         return this.getID()
                 + this.getCorrelationCase().getCaseUUID()
-                + this.getCorrelationDataSource().getName()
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
```diff
@@ -111,7 +111,7 @@ public Boolean equals(CorrelationAttributeInstance otherInstance) {
     public String toString() {
         return this.getID()
                 + this.getCorrelationCase().getCaseUUID()
-                + this.getCorrelationDataSource().getName()
+                + this.getCorrelationDataSource().getDeviceID()
```
rcordovano: Does this take into account the possibility that multiple data sources can have the same device id?

APriestman: I don't think they should.

rcordovano: Okay, thanks. There may be future work to do here.

PR #3567, opened 2018-03-15
3602 common case files r2

briangsweeney:
We think this is feature complete as per the story but have one concern documented in the changes files (CommonFilesSearchNode) area.


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesSearchNode.java
```diff
@@ -0,0 +1,45 @@
...
+ * Encapsulates data used to display common files search results in the top
+ * right pane.
+ */
+final class CommonFilesSearchNode extends AbstractNode {
+
+    private final Children children;
+
+    CommonFilesSearchNode(CommonFilesMetaData metaData) {
+        super(Children.create(new CommonFilesChildren(metaData), true), Lookups.singleton(metaData));
+        children = this.getChildren();
```
briangsweeney: Is this.getChildren() the best call now that we've implemented the child factory?

rcordovano: We don't need it.

PR #3572, opened 2018-03-19
Allow reports to be URLs

bcarrier:


Core/src/org/sleuthkit/autopsy/directorytree/ExternalViewerAction.java
```diff
@@ -171,4 +179,45 @@ public static void openFile(String mimeType, String ext, File file) {
...
+     * 
+     * @param path URL to open 
+     */
+    public static void openURL(String path) {
+        String url_path = path.replaceAll("\\\\","/");   
+        try {
+            Desktop.getDesktop().browse(new URI(url_path.replaceFirst("/","//")));          
+        } catch (IOException ex) {
+            logger.log(Level.WARNING, "Could not find a viewer for the given URL: " + url_path, ex); //NON-NLS
+            JOptionPane.showMessageDialog(null,
```
rcordovano: I have been nudging us towards using the message box part of MessageNotifyUtils instead of "raw" JOptionPanes.

PR #3575, opened 2018-03-19
3595 declutter local disk panel

dgrove727:


Core/src/org/sleuthkit/autopsy/casemodule/LocalDiskPanel.java
```diff
@@ -435,33 +407,33 @@ private void setPotentialImageWriterPath(LocalDisk disk) throws NoCurrentCaseExc
     }
 
     private boolean imageWriterPathIsValid() {
-        if ((!copyImageCheckbox.isSelected()) || !(diskTable.getSelectedRow() >= 0 && diskTable.getSelectedRow() < disks.size())) {
+        if ((!copyImageCheckbox.isSelected()) || !(localDisk != null)) {
```
APriestman: Please change the double negative in the second part

PR #3595, opened 2018-03-23
fix tagging in timeline window

millmanorama:


rcordovano:
@eugene7646, so the testing of the feature went well?

eugene7646:
There is no story referenced in the pull request so I wasn't sure what exactly to test. I assumed it's "just PR review".

millmanorama:
I discovered that tagging( and probably other actions that rely on the
Actions Global Context) were broken in the timeline window.  You could test
that tagging works as expected in the timeline window, the main autopsy
window and the cvt.

eugene7646:
So here are the things that I have tested:
1) Tag something in Autopsy and verify the tag shows up in Timeline - works.
2) Tag something in Timeline and verify the tag shows up in Autopsy - works.
3) Tag something in Autopsy and verify the tag shows up in CVT - works.
4) Tag something in CVT and verify the tag shows up in Autopsy - DOESN'T WORK. In fact, when I tag something in CVT, whether it's "Tag File" or "Tag Result", the tag doesn't show up in CVT either. The row doesn't get highlighted in yellow and nothing gets listed in the "Tags" column.

millmanorama:
Ok thanks for testing.  I will take another look.

millmanorama:
I just pushed a fix for this.  Would you test it one more time Eugene?

PR #3639, opened 2018-03-29
Fix for a race condition in Correlation Engine

eugene7646:
Made newCase() method synchronized, added a check for existing database

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
@@ -177,7 +177,6 @@ public abstract class AbstractSqlEamDb implements EamDb {
      */
     @Override
     public synchronized CorrelationCase newCase(CorrelationCase eamCase) throws EamDbException {
-        Connection conn = connect();
         
         // check if there is already an existing CorrelationCase for this Case
         CorrelationCase cRCase = getCaseByUUID(eamCase.getCaseUUID());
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -176,8 +176,14 @@ public void updateDbInfo(String name, String value) throws EamDbException {
...
     @Override
-    public CorrelationCase newCase(CorrelationCase eamCase) throws EamDbException {
+    public synchronized CorrelationCase newCase(CorrelationCase eamCase) throws EamDbException {
         Connection conn = connect();
+        
+        // check if there is already an existing CorrelationCase for this Case
+        CorrelationCase cRCase = getCaseByUUID(eamCase.getCaseUUID());
+        if (cRCase != null) {
+            return cRCase;
+        }
```
rcordovano: This should be moved inside the try block, or the Connection needs to be otherwise closed.

eugene7646: @rcordovano Unless we are talking about different things, I believe the connection for this call gets opened and closed inside the getCaseByUUID() method.

rcordovano: @eugene7646, look at the finally block of the code before the modification. The local conn is closed.

eugene7646: @rcordovano oh that connection! Good catch. Done.

PR #3644, opened 2018-03-30
3591 2nd test for file filter test

zhhl:


Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/IngestFileFiltersTest.java
```diff
@@ -156,6 +156,34 @@ public void testBasicDir() {
...
+        HashMap<String, Rule> rules = new HashMap<>();
+        rules.put("Rule", new Rule("testExtAndDirWithOneRule", new Rule.ExtensionCondition("jpg"), new MetaTypeCondition(MetaTypeCondition.Type.FILES), new ParentPathCondition("dir1"), null, null, null));
+        //Build the filter that ignore unallocated space and with one rule
+        FilesSet Files_Ext_Dirs_Filter = new FilesSet("Filter", "Filter to find all jpg files in dir1.", false, true, rules);
+        
+        try {
+            Case openCase = Case.getOpenCase();
+            runIngestJob(openCase.getDataSources(), Files_Ext_Dirs_Filter); 
+            FileManager fileManager = Case.getOpenCase().getServices().getFileManager();            
+            List<AbstractFile> results = fileManager.findFiles("%.jpg%");
```
rcordovano: This code is only checking the files with jpg extensions. That is not enough for this test case.

Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/IngestFileFiltersTest.java
```diff
@@ -156,6 +156,34 @@ public void testBasicDir() {
         }
     }
     
+    public void testExtAndDirWithOneRule() {
+        HashMap<String, Rule> rules = new HashMap<>();
+        rules.put("Rule", new Rule("testExtAndDirWithOneRule", new Rule.ExtensionCondition("jpg"), new MetaTypeCondition(MetaTypeCondition.Type.FILES), new ParentPathCondition("dir1"), null, null, null));
+        //Build the filter that ignore unallocated space and with one rule
+        FilesSet Files_Ext_Dirs_Filter = new FilesSet("Filter", "Filter to find all jpg files in dir1.", false, true, rules);
```
rcordovano: Please follow Java code naming conventions - use Camel-case with no initial capital letter for local variables.

Also, I don't see that the specification for the test calls for ignoring unallocated space. Maybe it is implied in the rest of the specification? I don't know enough about the test image to know, so please clarify.

Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/IngestFileFiltersTest.java
```diff
@@ -156,6 +156,34 @@ public void testBasicDir() {
...
+        
+        try {
+            Case openCase = Case.getOpenCase();
+            runIngestJob(openCase.getDataSources(), Files_Ext_Dirs_Filter); 
+            FileManager fileManager = Case.getOpenCase().getServices().getFileManager();            
+            List<AbstractFile> results = fileManager.findFiles("%%");
+            assertEquals(62, results.size());
```
rcordovano: I'm not sure of the purpose of this assert. DO you have doubts about whether or not FileManager.findFiles is working?

Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/IngestFileFiltersTest.java
```diff
@@ -156,6 +156,34 @@ public void testBasicDir() {
...
+            for (AbstractFile file : results) {
+                //Files with jpg extension in dir1 should have MIME Type
+                if (file.getParentPath().equalsIgnoreCase("/dir1/") && file.getNameExtension().equalsIgnoreCase("jpg")) {
+                    String errMsg = String.format("File %s (objId=%d) unexpectedly blocked by the file filter.", file.getName(), file.getId());
+                    assertTrue(errMsg, !(file.getMIMEType() == null || file.getMIMEType().isEmpty()));
```
rcordovano: I think this would assertion would be easier to read if you changed the conditional from 
!(file.getMIMEType() == null || file.getMIMEType().isEmpty()) 
to 
file.getMIMEType() != null && !file.getMIMEType().isEmpty()

Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/IngestFileFiltersTest.java
APriestman: To me this seems like a reasonable check. While I wouldn't expect findFiles to fail, if it did and returned, say, zero files, then the tests we want to run wouldn't execute at all and also wouldn't generate any warnings (since we looped over zero files).

rcordovano: Excellent point, @APriestman, thanks!

PR #3654, opened 2018-04-05
3611 Adding basic Services Health Monitor

APriestman:


Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    /**
+     * Check whether the health monitor database exists.
+     * Does not check the schema.
+     * @return true if the database exists, false otherwise
+     * @throws HealthMonitorException 
+     */
+    private boolean databaseExists() throws HealthMonitorException {       
+        try {
+            // Use the same database settings as the case
+            CaseDbConnectionInfo db = UserPreferences.getDatabaseConnectionInfo();
```
esaunders: Can the connection pool be used here?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    }
+    
+    /**
+     * Create a new health monitor database.
+     * @throws HealthMonitorException 
+     */
+    private void createDatabase() throws HealthMonitorException {
+        try {
+            // Use the same database settings as the case
+            CaseDbConnectionInfo db = UserPreferences.getDatabaseConnectionInfo();
```
esaunders: Same as above...can the connection pool be used?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+            connectionURL.append("/");
+            connectionURL.append(DATABASE_NAME);
+
+            connectionPool.setUrl(connectionURL.toString());
+            connectionPool.setUsername(db.getUserName());
+            connectionPool.setPassword(db.getPassword());
+
+            // tweak pool configuration
+            connectionPool.setInitialSize(3); // start with 3 connections
+            connectionPool.setMaxIdle(CONN_POOL_SIZE); // max of 10 idle connections
```
esaunders: There should only ever be 1 connection right? If so, might want to consider changing the initial size and max idle numbers.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+            // Make a shallow copy of the timing map. The map should be small - one entry
+            // per metric name.
+            timingMapCopy = new HashMap<>(timingInfoMap);
+            timingInfoMap.clear();
+        }
+        logger.log(Level.INFO, "Writing health monitor metrics to database");
+
+        String hostName;
+        try {
+            hostName = java.net.InetAddress.getLocalHost().getHostName();
```
esaunders: Don't know what happens under the covers when getting the host name of the local host but I've seen cases where this has to reach out to DNS servers and can be slow. Might want to consider getting this up front and storing it.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    
+    /**
+     * Add the timing metric data to the map.
+     * @param metric The metric to add. stopTiming() should already have been called.
+     */
+    private void addTimingMetric(TimingMetric metric) throws HealthMonitorException {
+
+        // Do as little as possible within the synchronized block to minimize
+        // blocking with multiple threads.
+        synchronized(this) {
```
esaunders: Could this synchronized block be eliminated if timingInfoMap was a ConcurrentHashMap?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+     * Write the collected metrics to the database.
+     * @throws HealthMonitorException 
+     */
+    private void writeCurrentStateToDatabase() throws HealthMonitorException {
+                
+        Map<String, TimingInfo> timingMapCopy;
+        
+        // Do as little as possible within the synchronized block since it will
+        // block threads attempting to record metrics.
+        synchronized(this) {
```
esaunders: Same as above. If timingInfoMap was a ConcurrentHashMap would there be a need for synchronization and copying here?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+        try {
+            hostName = java.net.InetAddress.getLocalHost().getHostName();
+        } catch (java.net.UnknownHostException ex) {
+            // Write it to the database but log a warning
+            logger.log(Level.WARNING, "Unable to look up host name");
+            hostName = "unknown";
+        }
+        
+        // Check if there's anything to report (right now we only have the timing map)
+        if(timingMapCopy.keySet().isEmpty()) {
```
esaunders: Could possibly be moved up top (depending on whether ConcurrentHashMap can be used).

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+            return;
+        }
+        
+        // Write to the database
+        CoordinationService.Lock lock = getSharedDbLock();
```
esaunders: I don't quite get the need for database locking here. Given that this is a multi-user/PostgreSQL only thing is database locking not sufficient? If 2 different processes attempt to insert data with the same name and host I'm guessing they will both succeed (since neither of those fields are key fields in the schema below).

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    private void initializeDatabaseSchema() throws HealthMonitorException {
+        Connection conn = connect();
+        if(conn == null) {
+            throw new HealthMonitorException("Error getting database connection");
+        }
+
+        try (Statement statement = conn.createStatement()) {
+            conn.setAutoCommit(false);
+            
+            StringBuilder createTimingTable = new StringBuilder();
```
esaunders: Minor comment but why use StringBuilder for both CREATE statements?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    /**
+     * Get an exclusive lock for the health monitor database.
+     * Acquire this before creating, initializing, or updating the database schema.
+     * @return The lock
+     * @throws HealthMonitorException 
+     */
+    private CoordinationService.Lock getExclusiveDbLock() throws HealthMonitorException{
+        try {
+            String databaseNodeName = DATABASE_NAME;
+            CoordinationService.Lock lock = CoordinationService.getInstance().tryGetExclusiveLock(CoordinationService.CategoryNode.HEALTH_MONITOR, databaseNodeName, 5, TimeUnit.MINUTES);
```
esaunders: Is it necessary to create a new String variable (databaseNodeName) here?

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+        }
+    } 
+    
+    /**
+     * Get an shared lock for the health monitor database.
+     * Acquire this before database reads or writes.
+     * @return The lock
+     * @throws HealthMonitorException 
+     */
+    private CoordinationService.Lock getSharedDbLock() throws HealthMonitorException{
```
esaunders: See comments above. Not sure why this is needed. Seems like database level locking might be sufficient.

Core/src/org/sleuthkit/autopsy/healthmonitor/HealthMonitorCaseEventListener.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.sleuthkit.autopsy.casemodule.Case;
+
+/**
+ * Listener for case events
+ */
+final class HealthMonitorCaseEventListener implements PropertyChangeListener {
+
+    private final ExecutorService jobProcessingExecutor;
```
esaunders: I find the names jobProcessingExecutor and CASE_EVENT_THREAD_NAME to be confusing here.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
```diff
@@ -0,0 +1,767 @@
...
+    private final static String MODULE_NAME = "ServicesHealthMonitor";
+    private final static String IS_ENABLED_KEY = "is_enabled";
+    private final static long DATABASE_WRITE_INTERVAL = 60; // Minutes
+    public static final CaseDbSchemaVersionNumber CURRENT_DB_SCHEMA_VERSION
+            = new CaseDbSchemaVersionNumber(1, 0);
+    
+    private static final AtomicBoolean isEnabled = new AtomicBoolean(false);
+    private static ServicesHealthMonitor instance;
+    
+    private ScheduledThreadPoolExecutor periodicTasksExecutor;
```
esaunders: I also find the name periodicTasksExecutor confusing.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
APriestman: I don't think so? The connection pool connects to the actual database so I don't think I can use it to create the database or cleanly test whether it exists.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
APriestman: No there can be multiple connections. The timer can start a database write as can the listener for case closing. Additionally displaying the metrics will use a database connection.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
APriestman: Yes inserting data normally by multiple nodes should be fine - that's why it's a shared lock. But eventually I'm sure we'll have to update the database schema during which time we'll need to stop the inserts by getting the exclusive lock.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
esaunders: Ok, I get it now.

Core/src/org/sleuthkit/autopsy/healthmonitor/ServicesHealthMonitor.java
esaunders: Ok.

APriestman:
I made the last few changes and did a test run.

Core/src/org/sleuthkit/autopsy/healthmonitor/HealthMonitorCaseEventListener.java
```diff
@@ -0,0 +1,53 @@
...
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.sleuthkit.autopsy.casemodule.Case;
+
+/**
+ * Listener for case events
+ */
+final class HealthMonitorCaseEventListener implements PropertyChangeListener {
```
rcordovano: It seems to me we can simplify things by making the monitor itself an event listener. We would register the monitor instance as a subscriber in getInstance, not the ServicesHealthMonitor constructor, to avoid leaking the "this" pointer. That would eliminate this class class with its additional thread pool and would allow ServicesHealthMonitor.DatabaseWriteTask to be private.

PR #3658, opened 2018-04-06
Switched from Java logger to Autopsy logger.

dgrove727:


Core/src/org/sleuthkit/autopsy/ingest/IngestMonitor.java
```diff
@@ -46,7 +46,7 @@
     public static final int DISK_FREE_SPACE_UNKNOWN = -1;
     private static final int INITIAL_INTERVAL_MS = 60000; //1 min.
     private static final int MAX_LOG_FILES = 3;
-    private static final java.util.logging.Logger MONITOR_LOGGER = java.util.logging.Logger.getLogger("monitor"); //NON-NLS
+    private static final java.util.logging.Logger monitorLogger = java.util.logging.Logger.getLogger("monitor"); //NON-NLS
```
wschaeferB: I assume this is a rare case where we want to continue using java.util.logging.Logger similar to KeywordSearch.java. Assuming that is the case a comment here explaining that we want to use java's logger and not ours as exists in KeywordSearch.java would be helpful if we ever run into this problem of the wrong logger being used elsewhere again in the futre.

PR #3672, opened 2018-04-11
3610 searchable auto ingest jobs

wschaeferB:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestJobsPanel.java
```diff
@@ -0,0 +1,210 @@
...
+    void customize() {
+        ((DefaultOutlineModel) outline.getOutlineModel()).setNodesColumnLabel(Bundle.AutoIngestJobsNode_caseName_text());
+        int indexOfColumn;
+        switch (status) {
+            case PENDING_JOB:
+                outlineView.setPropertyColumns(Bundle.AutoIngestJobsNode_dataSource_text(), Bundle.AutoIngestJobsNode_dataSource_text(),
+                        Bundle.AutoIngestJobsNode_jobCreated_text(), Bundle.AutoIngestJobsNode_jobCreated_text(),
+                        Bundle.AutoIngestJobsNode_priority_text(), Bundle.AutoIngestJobsNode_priority_text());
+                indexOfColumn = getColumnIndexByName(Bundle.AutoIngestJobsNode_priority_text());
+                if (indexOfColumn != -1) {
```
eugene7646: We should probably use a constant instead of just "-1". Something like "INVALID_COLUMN" or something.

PR #3687, opened 2018-04-13
3610 remove buttons add admin mode

wschaeferB:


wschaeferB:
Should only be reviewed after #3672 is integrated

PR #3700, opened 2018-04-18
  throw TskCoreException instead of swallowing ExcecutionException

millmanorama:


Core/src/org/sleuthkit/autopsy/timeline/datamodel/TimelineCacheException.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.timeline.datamodel;
+
+/**
+ * Exception thown when there is an error loading a cached value.
+ *
+ * NOTE: This should probably not be a RuntimeException. But the ripple effect
+ * of introducing this as a checked exception was spreading extremely far.
```
rcordovano: So does that mean that this exception is being treated as an unchecked exception?

rcordovano: So the argument is that having an unchecked exception propagate to the uncaught exception handler is better than calling the unchecked versions of the LoadingCache APIs? What is the impact on the state and continuing usability of the Timeline tool in each scenario?

Feel free to move this discussion to JIRA....

rcordovano: The motivation for wrapping the checked ExecutionExceptions from LoadingCache in a TimelineCacheException was to try to avoid confusion brought about by the association of ExecutionException with the Futures of background tasks. Basically, to remedy what in my opinion was a mistake in having LoadingCache throw ExecutionException instead of LoadingCacheException. However, I think replacing checked ExecutionExceptions with unchecked TimelineCacheExceptions is not a good solution. 

If we can satisfy the condition for using LoadingCache.getUnchecked in the [documentation](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/cache/LoadingCache.html#get(K))
(**"Warning: this method silently converts checked exceptions to unchecked exceptions, and should not be used with cache loaders which throw checked exceptions. In such cases use get(K) instead"**), I would be fine with calling LoadingCache.getUnchecked, and we can change EventRepository.countEvents and EventRepository.getEventStripes to be consistent with this strategy. Otherwise, we should stick with LoadingCache.get, in which case we will be dealing with either catching checked ExecutionExceptions or checked TimelineCacheExceptions, and I would still favor the latter.

millmanorama: That is just the thing, the cache loaders all do db calls and hence throw (unlikely and indicative of a serious problem) TskCoreExceptions.  It sounds like you will be most comfortable throwing a checked TimelineCacheException so I will update this PR to do that.

millmanorama:
looking at this again as I fix the conflicts.  I wonder if introducing this new exception is worth it.  There is no way to recover from it.  In all cases it ends up being handled identically to a TskCoreException:  log it, show a somewhat user friendly warning/notification, and then move on.

PR #3711, opened 2018-04-24
3754 Add metrics for hash lookup and a database query

APriestman:


Core/src/org/sleuthkit/autopsy/healthmonitor/EnterpriseHealthMonitor.java
```diff
@@ -276,12 +282,34 @@ public static void submitTimingMetric(TimingMetric metric) {
...
+    /**
+     * Submit the metric that was previously obtained through getTimingMetric(),
+     * incorporating a count that the time should be divided by.
+     * Call this immediately after the section of code being timed.
+     * This method is safe to call regardless of whether the Enterprise Health
+     * Monitor is enabled.
+     * @param metric The TimingMetric object obtained from getTimingMetric()
+     * @param normalization The number to divide the time by (a zero here will be treated as a one)
+     */
+    public static void submitNormalizedTimingMetric(TimingMetric metric, long normalization) {
```
rcordovano: I think it would be nice to standardize on either "submit" or "add", e.g., submitTimingMetric and submitNormalizedTimingMetric.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbIngestModule.java
```diff
@@ -182,8 +184,20 @@ public ProcessResult process(AbstractFile file) {
...
         if (md5Hash == null || md5Hash.isEmpty()) {
             try {
+                TimingMetric metric = EnterpriseHealthMonitor.getTimingMetric("Disk Reads: Hash calculation");
                 long calcstart = System.currentTimeMillis();
                 md5Hash = HashUtility.calculateMd5Hash(file);
+                if (file.getSize() > 0) {
+                    // Surprisingly, the hash calculation does not seem to be correlated that
+                    // strongly with file size until the files get large.
+                    // Only normalize if the file size is greater than ~1MB.
+                    if (file.getSize() < 1000000) {
```
rcordovano: I would prefer a named constant here, instead of a "magic number."

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbIngestModule.java
```diff
@@ -182,8 +184,20 @@ public ProcessResult process(AbstractFile file) {
...
+                        EnterpriseHealthMonitor.submitTimingMetric(metric);
+                    } else {
+                        // In testing, this normalization gave reasonable resuls
+                        EnterpriseHealthMonitor.submitNormalizedTimingMetric(metric, file.getSize() / 500000);
```
rcordovano: I would prefer a named constant here, instead of a "magic number." Also note typo in comment.

Core/src/org/sleuthkit/autopsy/healthmonitor/EnterpriseHealthMonitor.java
APriestman: There is a standard - "submit" is what clients do, those are our public methods. They're submitting a metric to the health monitor, and the metric might get ignored if the monitor isn't enabled. "add" is when the monitor adds the metric to its internal list.

PR #3723, opened 2018-04-27
3757 Group views in the case tree by data source

raman-bt:


rcordovano:
Closing because we have decided to change the structure of the case tree a bit. A new PR will replace this one soon.

PR #3724, opened 2018-04-27
3750 password protection access databases

dgrove727:


Core/test/qa-functional/src/org/sleuthkit/autopsy/modules/encryptiondetection/EncryptionDetectionTest.java
```diff
@@ -127,11 +79,28 @@ public void tearDown() {
     public void testPasswordProtection() {
```
wschaeferB: I just noticed an issue with this test that admittedly has existed since it's first version.  

It will pass with any data source, because it only expects all files of the given mime types to have the encryption detection artifact, so when those files don't exist it continues to pass and does not test them.  So if you don't delete the old image to get the new one it continues to pass.  

I believe the test requires us to hard code some knowledge about what it should expect to see in the image, but there might be a better way that I am not thinking of.

rcordovano:
Please address merge conflicts.

PR #3728, opened 2018-04-30
show warning in less intrusive JFX notifications instead of Message b…

millmanorama:
…oxes, suppress NPEs during painting.


Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -142,12 +147,38 @@
...
+            protected mxGraphComponent.mxGraphControl createGraphControl() {
+
+                return new mxGraphControl() {
+
+                    @Override
+                    public void paint(Graphics graphics) {
+                        try {
+                            super.paint(graphics);
+                        } catch (NullPointerException ex) { //NOPMD  We can't find the underlying cause in jgraphx
+                            logger.log(Level.WARNING, "There was a NPE while painting the VisualizationPanel", ex);
```
rcordovano: What does NOPMD mean? Why are NPEs acceptable here? Please add some more explanation.

millmanorama: NOPMD disables the warning that EeasyPMD would normaly show.  It is similar to the NO-NLS comment for internationalization.  I am not sure if codacy honors it.

I will improve the comment

PR #3730, opened 2018-05-01
Add metrics to correlation engine

APriestman:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -934,6 +936,8 @@ public void bulkInsertArtifacts() throws EamDbException {
                 if (bulkArtifactsCount == 0) {
                     return;
                 }
+                
+                TimingMetric timingMetric = EnterpriseHealthMonitor.getTimingMetric("Correlation Engine: Bulk insert");
```
eugene7646: I think the "Correlation Engine: Bulk insert" string should be declared a constant, given that we are using it as an identifier to start and stop the timer.

Core/src/org/sleuthkit/autopsy/centralrepository/ingestmodule/IngestModule.java
```diff
@@ -129,7 +131,9 @@ public ProcessResult process(AbstractFile abstractFile) {
          */
         if (abstractFile.getKnown() != TskData.FileKnown.KNOWN && flagTaggedNotableItems) {
             try {
+                TimingMetric timingMetric = EnterpriseHealthMonitor.getTimingMetric("Correlation Engine: Notable artifact query");
```
eugene7646: Same goes for "Correlation Engine: Notable artifact query" string, I think it should be a constant.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
APriestman: It's only used in creating the timer - stopping is done on the metric object so we don't need to match it up with anything. I don't think making it a constant gains us much since it will only be used in a single spot.

PR #3732, opened 2018-05-01
3668 intra case correlation content viewer

aziehl:
Pull for VIK-3688 completion.


APriestman:
Please resolve the merge conflicts

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -414,64 +422,130 @@ private AbstractFile getAbstractFileFromNode(Node node) {
...
+        ArtifactKey instKey = new ArtifactKey(dataSource.getDeviceID(), filePath);
+        CorrelationAttributeInstance caseDbInstance = new CorrelationAttributeInstance(caze, dataSource, filePath, "", caseDbFile.getKnown());
+        TskData.FileKnown knownStatus = caseDbInstance.getKnownStatus();
+        // If not known, check Tags for known and set
+        if (knownStatus.compareTo(TskData.FileKnown.KNOWN) != 0 && knownStatus.compareTo(TskData.FileKnown.BAD) != 0) {
+            List<ContentTag> fileMatchTags = openCase.getServices().getTagsManager().getContentTagsByContent(caseDbFile);
+            for (ContentTag tag : fileMatchTags) {
+                TskData.FileKnown tagKnownStatus = tag.getName().getKnownStatus();
+                if (tagKnownStatus.compareTo(TskData.FileKnown.KNOWN) == 0 || tagKnownStatus.compareTo(TskData.FileKnown.BAD) == 0) {
+                    caseDbInstance.setKnownStatus(TskData.FileKnown.BAD);
```
APriestman: Is this saying that if the tag status is either KNOWN or BAD, then set the known status to BAD? That doesn't seem right. I don't think tags can have the KNOWN status anyway so we can probably just test for BAD.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -414,64 +422,130 @@ private AbstractFile getAbstractFileFromNode(Node node) {
...
+                    break;
+                }
+            }
+        }
+
+        // If known, or not in CR, add
+        if (knownStatus.compareTo(TskData.FileKnown.KNOWN) == 0 || knownStatus.compareTo(TskData.FileKnown.BAD) == 0 || !artifactInstances.containsKey(instKey)) {
```
APriestman: I feel like using "equals" instead of "compareTo" would make this code (and other similar code) more readable. This would change to: 
if (knownStatus.equals(TskData.FileKnown.KNOWN) || knownStatus.equals(TskData.FileKnown.BAD) || ...

aziehl:
Ann, I resolved the merge conflicts and incorporated your feedback. I removed the Known checks since you were right, the functionality appears to be the same just checking equality on Bad.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/ArtifactKey.java
```diff
@@ -0,0 +1,70 @@
...
+package org.sleuthkit.autopsy.centralrepository.contentviewer;
+
+import java.util.Objects;
+
+/**
+ * Used as a key to ensure we eliminate duplicates from the result set by not overwriting CR correlation instances.
+ */
+final class ArtifactKey {
+    
+    private final String dataSource;
```
APriestman: I think this would be more clear if you renamed this to dataSourceDeviceID (or similar) and updated  the name of getDataSource(). The constructor is a bit ambiguous as it stands.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -414,64 +422,131 @@ private AbstractFile getAbstractFileFromNode(Node node) {
...
+                TskData.FileKnown tagKnownStatus = tag.getName().getKnownStatus();
+                if (tagKnownStatus.equals(knownBad)) {
+                    caseDbInstance.setKnownStatus(knownBad);
+                    break;
+                }
+            }
+        }
+
+        // If known, or not in CR, add
+        if (knownStatus.equals(knownBad) || !artifactInstances.containsKey(instKey)) {
```
APriestman: I think you need to get the known status from caseDbInstance here instead of using the original value, since it may have been updated if a bad tag was found?

APriestman:
Looks good - you should just need to fix the new conflict that popped up.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -414,64 +422,133 @@ private AbstractFile getAbstractFileFromNode(Node node) {
             } catch (EamDbException ex) {
                 LOGGER.log(Level.SEVERE, "Error connecting to DB", ex); // NON-NLS
             }
+
+        } else {
+            try {
+                // If EamDb not enabled, get the Files default correlation type to allow Other Occurances to be enabled.   
+                if(this.file != null) {
```
APriestman: Sorry one more thing. I think you need to do the same checks from above about the MD5 being null or empty. I'm getting stack traces in the logs when I click on things that have no hash (for example, an unalloc space file)

aziehl:
Thanks, I'll make that fix, resolve the conflict and bring this up to the latest develop.

aziehl:
I've added the md5 null check, updated the code to pass codeacy and merged in the latest develop. I changed getOpenCase() to use getCurrentCase(), hopefully that's the correct usage versus getCurrenCaseThrows().

PR #3751, opened 2018-05-08
3830 call eventbus.unregister on object given as arg instead of 0

wschaeferB:


millmanorama:
Ha!  I caught this too in my timeline work.

PR #3755, opened 2018-05-09
3784 bitlocker detection

dgrove727:


Core/test/qa-functional/src/org/sleuthkit/autopsy/testutils/IngestUtils.java
```diff
@@ -52,13 +52,14 @@ private IngestUtils() {
...
         try {
             DataSourceProcessorRunner.ProcessorCallback callBack = DataSourceProcessorRunner.runDataSourceProcessor(dataSourceProcessor, dataSourcePath);
-            List<String> callbackErrorMessageList = callBack.getErrorMessages();
-            String errorMessage = String.format("The data source processor callback produced %d error messages.", callbackErrorMessageList.size());
-            assertEquals(errorMessage, 0, callbackErrorMessageList.size());
+            /*
+             * Ignore the callback error messages. Sometimes it's perfectly
+             * valid for it to not be able to detect a file system, which is one
+             * of the errors that can be returned.
+             */
```
wschaeferB: I had this same issue and solved it in a different way on my pull request, us both having to change this same code is turning into a headache of merge conflicts, I will resolve it when mine goes in though. 

For reference and in case you see an issue with it the way I did was to have the addDataSource method return the List<String> of error messages so that each test could check that there were the number of error messages it expected if it desired and handle differences accordingly.

rcordovano: @dgrove727 and @wschaeferB, when I wrote DataSourceProcessorRunner.runDataSourceProcessor, I was working from the premise that we would be creating test data sources that we know how to process and there would be no such thing as acceptable DSP errors, which is how IngestUtils.addDataSource works before this PR.  Assuming that acceptable DSP errors actually is a legitimate use case for functional testing, unconditionally ignoring DSP errors for ALL tests that anyone writes in the forseeable future is definitely not the way to go. But I don't see counting error messages as a general solution, either. What we should do instead is change DataSourceProcessorRunner.ProcessorCallback so that it stores, and provides a getter for, the DataSourceProcessorCallback.DataSourceProcessorResult enum value it gets from the DSP, which could then be checked for CRITICAL_ERRORS vs NONCRITICAL_ERRORS or NO_ERRORS.

rcordovano: @dgrove727 and @wschaeferB, after discussion with Will, I will integrate this story to get things unblocked, but we will then need to go back and update the utils. I will make a story for that.

rcordovano:
@dgrove727, please resolve the Codacy errors for this PR. I will look into the Travis and Appveyor errors.

PR #3775, opened 2018-05-15
3836 ingest utils

dgrove727:


Core/test/qa-functional/src/org/sleuthkit/autopsy/modules/encryptiondetection/EncryptionDetectionTest.java
```diff
@@ -260,15 +258,7 @@ public void testVeraCryptSupport() {
...
-            List<String> errorMessages = IngestUtils.addDataSource(dataSourceProcessor, VERACRYPT_DETECTION_IMAGE_PATH);
-            String joinedErrors;
-            if (errorMessages.isEmpty()) {
-                joinedErrors = "Encrypted partition did not cause error, it was expected to";
-            } else {
-                joinedErrors = String.join(System.lineSeparator(), errorMessages);
-            }
-            //there will be 1 expected error regarding the encrypted partition not having a file system
-            assertEquals(joinedErrors, 1, errorMessages.size());
+            IngestUtils.addDataSource(dataSourceProcessor, VERACRYPT_DETECTION_IMAGE_PATH);
```
rcordovano: I am a bit confused by the changes to this file. The checks for the number of DSP errors have been removed, but not replaced with checking based on the new getResult method of the IngestUtils DSP callback. Also, the other existing functional test classes have not been updated to check, either. Note that JIRA-3836 states that removing the code from IngestUtils that asserted on all DSP errors meant it had to be replaced with checking of the DSP result code/errors as appropriate in the client code. We need to fulfill the DoD of the story.

dgrove727: The call to the new method occurs at lines 65-69 of IngestUtils. Putting it there allows the 'addDataSource()' method to look for errors. If the result is 'CRITICAL_ERRORS', it then asserts a failure. Then, all I had to do was remove the extra code that was added to the other classes that checked the count of error messages. Reading the DoD, I believe this fulfills the requirements. However, if you think the calling classes should do the error check and fail assertion, I can go back and make that change.

rcordovano: I'm sorry, I don't know how I missed that in such a small PR. This is perfect as is.

PR #3776, opened 2018-05-15
3815 aid21 refocusing

wschaeferB:


wschaeferB:
Going to make changes as discussed Eamonn to make this work in an improved manner.

PR #3777, opened 2018-05-15
Add ingest progress snapshot support to AID

esaunders:


Core/src/org/sleuthkit/autopsy/ingest/IngestTasksScheduler.java
```diff
@@ -825,8 +830,9 @@ int countRunningTasksForJob(long jobId) {
     /**
      * A snapshot of ingest tasks data for an ingest job.
      */
-    class IngestJobTasksSnapshot {
+    public static final class IngestJobTasksSnapshot implements Serializable {
```
rcordovano: A public inner class of a package access outer class?

PR #3779, opened 2018-05-16
Initial creation of plaso module

markmckinnon:
Initial creation of plaso module that will run plaso against an image.


Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoModuleFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import org.openide.util.lookup.ServiceProvider;
+import org.sleuthkit.autopsy.coreutils.Version;
+import org.sleuthkit.autopsy.ingest.IngestModuleFactoryAdapter;
+import org.sleuthkit.autopsy.ingest.DataSourceIngestModule;
+import org.sleuthkit.autopsy.ingest.IngestModuleFactory;
+import org.sleuthkit.autopsy.ingest.IngestModuleIngestJobSettings;
+
+/**
+  * An factory that creates data source ingest modules that verify the integrity
+ * of Expert Witness Format (EWF), i.e., .e01 files .
```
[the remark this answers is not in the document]
markmckinnon: Revamping things so will take everything into consideration you posted.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+                resultSet = tempdbconnect.executeQry(sqlStatement);
+                while (resultSet.next()) {
+                    Collection<BlackboardAttribute> bbattributes = new ArrayList<>();
+                    bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DATETIME, Bundle.PlasoWorker_event_datetime(), resultSet.getLong("TSK_DATETIME")));
+                    bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DESCRIPTION, Bundle.PlasoWorker_event_description(), resultSet.getString("TSK_DESCRIPTION")));
+                    
+                    List<AbstractFile> resolvedFiles = getAbstractFile(resultSet.getString("filename"));
+                    try {
+                        for (AbstractFile resolvedFile : resolvedFiles) {
+                            BlackboardArtifact bbart = resolvedFile.newArtifact(BlackboardArtifact.ARTIFACT_TYPE.TSK_EVENT);
```
millmanorama: https://github.com/sleuthkit/sleuthkit/pull/1260  renames TSK_EVENT to TSK_TL_EVENT.  It also introduces two new artifact types you can use the same way, but for more specific events: TSK_LOG_EVENT and TSK_REGISTRY_EVENT.

For TSK_TL_EVENTs. If possible please use the TSK_EVENT_TYPE attribute to indicate the event type.  The value should be the id of one of the instances of EventType available as static members of the interface.  i.g. EventType.EMAIL.getTypeID().  Use EventType.OTHER.getTypeID() if none of the available ones fit.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+    @Override
+    public ProcessResult process(Content dataSource, DataSourceIngestModuleProgress statusHelper) {
+        String imgName = dataSource.getName();
+        File log2TimeLineExecutable;
+        File psortExecutable;
+        Case currentCase;
+        String moduleOutputPath;
+        String tempOutputPath;
+        
+        Image img = (Image) dataSource;
```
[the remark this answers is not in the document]
rcordovano: Good catch, Jon. This cast is definitely not safe. Assuming Plaso only works on image files, it should do an instanceof Image check with an early return if the check does not pass.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        ProcessBuilder log2TimeLineCommand = buildLog2TimeLineCommand(log2TimeLineExecutable, tempOutputPath, imgFile[0]);
+        ProcessBuilder psortCommand = buildPsortCommand(psortExecutable, tempOutputPath);
+        //logger.log(Level.INFO, "Hash value stored in {0}: {1}", new Object[]{imgName, storedHash}); //NON-NLS
+
+        logger.log(Level.INFO, Bundle.PlasoIngestModule_startUp_message()); //NON-NLS
+
+        try {
+            plasoWorker processImage = new plasoWorker(log2TimeLineCommand, psortCommand, moduleOutputPath, tempOutputPath);
```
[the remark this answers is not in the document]
rcordovano: Yes, that's right.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+        moduleOutputPath = Paths.get(currentCase.getModuleDirectory(), PLASO).toString();
+        File directory = new File(String.valueOf(moduleOutputPath));
+        if (!directory.exists()) {
+            directory.mkdirs();   
+        }
+        tempOutputPath = Paths.get(currentCase.getTempDirectory(), PLASO, imgName).toString();
+        directory = new File(String.valueOf(tempOutputPath));
+        if (!directory.exists()) {
+            directory.mkdirs();
+        }
```
rcordovano: The preceding directory creation code looks incomplete. The data source object id is queried, but not used. However, although data source object ids are unique within a case, it is possible that ingest is run multiple times on the same data source, so it is safest to do what we talked about in stand up, i.e., to use a time stamp from TimeStampUtils.createTimeStamp instead (in combination with a data source if you want to be really safe) as a way to get a unique path. In any case, a unique suffix is not actually being used here for the module output directory, and the temp directory will not be guaranteed unique because image names are not guaranteed to be unique.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+
+        ProcessBuilder log2TimeLineCommand = buildLog2TimeLineCommand(log2TimeLineExecutable, tempOutputPath, imgFile[0]);
+        ProcessBuilder psortCommand = buildPsortCommand(psortExecutable, tempOutputPath);
+        //logger.log(Level.INFO, "Hash value stored in {0}: {1}", new Object[]{imgName, storedHash}); //NON-NLS
```
rcordovano: Please remove this commented out line. It looks like it was copied and pasted here as a temporary, expedient model for writing log messages.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+
+        logger.log(Level.INFO, Bundle.PlasoIngestModule_startUp_message()); //NON-NLS
+
+        try {
+            plasoWorker processImage = new plasoWorker(log2TimeLineCommand, psortCommand, moduleOutputPath, tempOutputPath);
+            processImage.execute();
+        } catch (Exception ex) {
+          //TODO Make a real exception
```
rcordovano: Please resolve the TODO, which should happen when you remove the SwingWorker. FYI for future reference, SwingWorker.execute never throws an exception. Instead, the SwingWorker framework already provides an exception firewall around SwingWorker.doInBackground. To get exceptions from SwingWorker.doInBackground, you need to call get() in done().

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+            logger.log(Level.SEVERE, Bundle.PlasoIngestModule_error_running(), ex);
+            }
+ 
+        return ProcessResult.OK;
+    }
+    
+    private ProcessBuilder buildLog2TimeLineCommand(File log2TimeLineExecutable, String tempOutputPath, String imageName) {
```
rcordovano: A small point, but it looks like this could be static.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+         * permissions Autopsy uses.
+         */
+        processBuilder.environment().put("__COMPAT_LAYER", "RunAsInvoker"); //NON-NLS
+        processBuilder.redirectOutput(new File(tempOutputPath + File.separator +  "log2timeline_output.txt"));
+        processBuilder.redirectError(new File(tempOutputPath + File.separator +  "log2timeline_err.txt"));  //NON-NLS
+
+        return processBuilder; 
+    }
+    
+    private ProcessBuilder buildPsortCommand(File psortExecutable, String tempOutputPath) {
```
rcordovano: A small point, but it looks like this could be static.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,410 @@
...
+        private void createPlasoArtifacts(String plasoDb) {
+            org.sleuthkit.datamodel.Blackboard blackboard;
+            SleuthkitCase sleuthkitCase = Case.getCurrentCase().getSleuthkitCase();
+            blackboard = sleuthkitCase.getBlackboard();
+            ResultSet resultSet;
+            String connectionString = "jdbc:sqlite:" + plasoDb; //NON-NLS
+            String sqlStatement = "select substr(filename,1) filename, strftime('%s', datetime) 'TSK_DATETIME', description 'TSK_DESCRIPTION' \n" +
+                                  "  from log2timeline where source != 'FILE';";
+//            String sqlStatement = "select datetime('unixepoch', datetime) 'TSK_DATETIME', description 'TSK_DESCRIPTION' "
+//                                +  " from log2timeline where source != 'REG';";
```
rcordovano: Please remove commented out code.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,350 @@
...
+    @Override
+    public ProcessResult process(Content dataSource, DataSourceIngestModuleProgress statusHelper) {
+        statusHelper.switchToIndeterminate();
+        Image img;
+        if (dataSource instanceof Image) {
+           img = (Image) dataSource;
+        } else {
+            logger.log(Level.SEVERE, Bundle.PlasoIngestModule_dataSource_not_an_image());
+            MessageNotifyUtil.Message.info(Bundle.PlasoIngestModule_error_running());
+            return ProcessResult.ERROR;
```
bcarrier: This should probably just be OK. ERROR is more for critical system errors, not files that we choose to ignore.

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,350 @@
...
+        Image img;
+        if (dataSource instanceof Image) {
+           img = (Image) dataSource;
+        } else {
+            logger.log(Level.SEVERE, Bundle.PlasoIngestModule_dataSource_not_an_image());
+            MessageNotifyUtil.Message.info(Bundle.PlasoIngestModule_error_running());
+            return ProcessResult.ERROR;                   
+        }
+
+        File log2TimeLineExecutable = locateExecutable(LOG2TIMELINE_EXECUTABLE);
```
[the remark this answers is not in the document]
markmckinnon: Which part, the cast check or the locateExecutable?

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
markmckinnon: Will do

Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -0,0 +1,350 @@
...
+
+        try {
+            statusHelper.progress(Bundle.PlasoIngestModule_running_log2timeline());
+            ExecUtil.execute(log2TimeLineCommand, new DataSourceIngestModuleProcessTerminator(context));
+            if (context.dataSourceIngestIsCancelled()) {
+                logger.log(Level.INFO, Bundle.PlasoIngestModule_log2timeline_cancelled()); //NON-NLS
+                MessageNotifyUtil.Message.info(Bundle.PlasoIngestModule_log2timeline_cancelled());
+                return ProcessResult.OK;
+            }
+            statusHelper.progress(Bundle.PlasoIngestModule_running_psort());
```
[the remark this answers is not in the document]
markmckinnon: I can check to see if the SQLite plaso file was created and if it was not then not run the rest.  I will do this for psort as well.

PR #3794, opened 2018-05-22
3861 npe other occurrences

briangsweeney:
fixes the npe when the CR is disabled and you try to use the other occurrences tab


Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -539,9 +540,21 @@ public boolean isSupported(Node node) {
         //  Is supported if this node
         //      has correlatable content (File, BlackboardArtifact) OR
         //      other common files across datasources.
-        return this.file != null
+        
+        EamDb eamDb = null;
+        try {
+            eamDb = EamDb.getInstance();
```
APriestman: Is there a reason not to use EamDb.isEnabled() here?

briangsweeney: Nah - I wasn't thinking.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -44,6 +44,7 @@
 import javax.swing.table.TableCellRenderer;
 import javax.swing.table.TableColumn;
 import org.openide.nodes.Node;
+import org.openide.util.Exceptions;
```
APriestman: Please remove the now unused import

PR #3804, opened 2018-05-23
3802: Group tree by data source

raman-bt:


Core/src/org/sleuthkit/autopsy/datamodel/DataSourceGroupingChildren.java
```diff
@@ -0,0 +1,101 @@
...
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Child factory for Data Source grouping nodes 
+ * A node is created for each data source in the case
+ * 
+ */
+public class DataSourceGroupingChildren extends Children.Keys<DataSource> {
```
rcordovano: This class is not used and should be discarded.

Core/src/org/sleuthkit/autopsy/datamodel/DataSourceGroupingNode.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class DataSourceGroupingNode extends DisplayableItemNode {
+
+    private static final Logger logger = Logger.getLogger(DataSourceGroupingNode.class.getName());
+
+    /**
+     * Creates the Data source grouping node for the given data source,  
+     * 
+     * @param dataSource specifies the data source 
+     */
```
rcordovano: This comment for the published, public API documentation is a little funky with respect to capitalization and punctuation.

Core/src/org/sleuthkit/autopsy/datamodel/DataSourceGroupingNode.java
```diff
@@ -0,0 +1,99 @@
...
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.Image;
+import org.sleuthkit.datamodel.LocalFilesDataSource;
+
+
+/**
+ *  Data source grouping node - an optional grouping node in the data tree view
+ * 
+ */
+public class DataSourceGroupingNode extends DisplayableItemNode {
```
rcordovano: Does this class have to be public? It has a package level constructor...

PR #3805, opened 2018-05-24
include modules and configuration of autopsy in html report

rishwanth1995:


Core/src/org/sleuthkit/autopsy/report/ReportHTML.java
```diff
@@ -1094,6 +1092,54 @@ private void writeSummary() {
...
+                if (out != null) {
+                    out.flush();
+                    out.close();
+                }
+            } catch (IOException ex) {
+            }
+        }
+    }
+    
+    private StringBuilder writeSummaryCaseDetails(){
```
bcarrier: Can you add JavaDocs for the new methods?

PR #3813, opened 2018-05-25
3832 fix aid job ordering

wschaeferB:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestJob.java
```diff
@@ -627,7 +629,21 @@ public int compare(AutoIngestJob o1, AutoIngestJob o2) {
...
+                    //if the manifest files were created at the same time compare with the jobs case name
+                    comparisonResult = -job.getManifest().getCaseName().compareTo(anotherJob.getManifest().getCaseName());
+                    if (comparisonResult == 0) {
+                        //if the case name is the same compare with the jobs datasource file name
+                        comparisonResult = -job.getManifest().getDataSourcePath().getFileName().toString().compareTo(anotherJob.getManifest().getDataSourcePath().getFileName().toString());
+                        //if they are still the same at this point they may be ordered incosistantly
+                    }
+                }
+            }
+            return comparisonResult;
```
rcordovano: I think it is confusing to have a PriorityComparator.compare that does general comparison and an AuotIngestJob.compareTo that compares manifest file dates. Please refactor this code so  that this code is in compareTo. If we no longer need PriorityComparator, let's get rid of it. Ditto for any of the other Comparator implementations.

wschaeferB: Are you saying to get rid of all Comparators in autopsy or just for the AID // AID 2.0?

rcordovano: The scope of removing comparators I am proposing is limited to the AutoIngestJob class. If we do not need to use the inner job priority comparator class any more because we are using AutoIngestJob.compareTo for collection sorts, we can delete the unused comparator. If the other inner comparator classes in AutoIngestJob are not used anywhere anymore, we can get rid of them, too. SO basically, I'm just talking about unused code removal.

PR #3816, opened 2018-06-01
Remove unsafe Cast of datasource

markmckinnon:
Remove unsafe cast of datasource.


markmckinnon:
Did not mean for all this to get put into my simple PR

PR #3818, opened 2018-06-01
Refactor other occurence tab for intra cases

bcarrier:
UI-based changes came in from NetBeans when I moved the inner class in....


Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -507,29 +520,49 @@ private AbstractFile getAbstractFileFromNode(Node node) {
...
                     break;
                 }
             }
         }
 
-        // If known, or not in CR, add
-        if (caseDbInstance.getKnownStatus().equals(knownBad) || !artifactInstances.containsKey(instKey)) {
-            artifactInstances.put(instKey, caseDbInstance);
+        // make a key to see if the file is already in the hashset
+        CorrelationCase correlationCase = EamDb.getInstance().getCase(autopsyCase);
```
APriestman: NPE here if the CR is not enabled

PR #3820, opened 2018-06-04
Added additional rules to qualify files to be processed.

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/encryptiondetection/EncryptionDetectionFileIngestModule.java
```diff
@@ -134,17 +136,39 @@ public void startUp(IngestJobContext context) throws IngestModule.IngestModuleEx
...
-                    if (mimeType.equals("application/octet-stream")) {
-                        if (isFileEncryptionSuspected(file)) {
-                            return flagFile(file, BlackboardArtifact.ARTIFACT_TYPE.TSK_ENCRYPTION_SUSPECTED,
-                                    String.format(Bundle.EncryptionDetectionFileIngestModule_artifactComment_suspected(), calculatedEntropy));
+                    if (!file.isMetaFlagSet(TskData.TSK_FS_META_FLAG_ENUM.UNALLOC)
+                            && !file.isDirNameFlagSet(TskData.TSK_FS_NAME_FLAG_ENUM.UNALLOC)) {
+                        /*
+                         * Is the file in FILE_IGNORE_LIST?
+                         */
+                        boolean ignoreFile = false;
```
rcordovano: Personally, I find code that sets flags and then checks them in later conditionals can quickly become hard to read and maintain. I'm thinking we can just return immediately if we find one of the files we are ignoring.

PR #3822, opened 2018-06-05
new method added in autopsy centralrepository datamodel to get all known bad artifact instances

rishwanth1995:


Core/nbproject/project.xml
```diff
@@ -314,6 +314,7 @@
                 <package>org.sleuthkit.autopsy.casemodule</package>
                 <package>org.sleuthkit.autopsy.casemodule.events</package>
                 <package>org.sleuthkit.autopsy.casemodule.services</package>
+                <package>org.sleuthkit.autopsy.centralrepository.datamodel</package>
```
bcarrier: This should be removed.

PR #3826, opened 2018-06-07
3846 add view comments central repository

dgrove727:


Core/src/org/sleuthkit/autopsy/centralrepository/AddEditCentralRepoCommentAction.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Create a Add/Edit dialog for the correlation attribute file instance
+     * comment. The comment will be updated in the database if the file instance
+     * exists there, or a new file instance will be added to the database with
+     * the comment attached otherwise.
+     */
+    public void addEditCentralRepoComment() {
+        CentralRepoCommentDialog centralRepoCommentDialog = new CentralRepoCommentDialog(correlationAttribute);
+        centralRepoCommentDialog.display();
```
APriestman: I think the dialog would look nicer with a title. Our logo would be nice, too, if it's not too much work.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1087,38 +1092,156 @@ public void bulkInsertCases(List<CorrelationCase> cases) throws EamDbException {
...
+     *
+     * @throws EamDbException
+     */
+    @Override
+    public void updateAttributeInstanceComment(CorrelationAttribute eamArtifact) throws EamDbException {
+        if (eamArtifact == null) {
+            throw new EamDbException("Correlation attribute is null");
+        }
+
+        CorrelationAttributeInstance eamInstance = eamArtifact.getInstances().get(0);
```
APriestman: Add check that eamInstance is not null

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1087,38 +1092,156 @@ public void bulkInsertCases(List<CorrelationCase> cases) throws EamDbException {
...
+     *
+     * @return The correlation attribute if it exists; otherwise null.
+     *
+     * @throws EamDbException
+     */
+    @Override
+    public CorrelationAttribute getCorrelationAttribute(CorrelationAttribute.Type type, CorrelationCase correlationCase,
+            CorrelationDataSource correlationDataSource, String value, String filePath) throws EamDbException {
+
+        if (type == null) {
```
APriestman: Add checks that the other four parameters are not null

Core/src/org/sleuthkit/autopsy/centralrepository/AddEditCentralRepoCommentAction.java
```diff
@@ -0,0 +1,144 @@
...
+     *             derived.
+     *
+     * @return The instance.
+     * 
+     * @throws EamDbException
+     * @throws NoCurrentCaseException
+     * @throws TskCoreException
+     */
+    @Messages({"AddEditCentralRepoCommentAction.menuItemText.addEditCentralRepoComment=Add/Edit Central Repository Comment"})
+    public static AddEditCentralRepoCommentAction createAddEditCentralRepoCommentAction(AbstractFile file)
```
APriestman: It might be cleaner for this (or a method down the chain) to wrap all the exception types into one instead of throwing three different types that you then have to check for individually.

Core/src/org/sleuthkit/autopsy/centralrepository/AddEditCentralRepoCommentAction.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Private constructor to create an instance given an AbstractFile.
+     *
+     * @param file  The file from which a correlation attribute to modify is
+     *              derived.
+     * @param title The text for the menu item.
+     */
+    private AddEditCentralRepoCommentAction(AbstractFile file, String title) throws EamDbException, NoCurrentCaseException, TskCoreException {
```
APriestman: Looking at this again, I don't think anything in here actually throws an exception and you can get rid of the entire "throws" part. This should allow you to remove the exception handling from createAddEditCentralRepoCommentAction, and possibly get rid of the new exception class if it isn't used anywhere else.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1241,7 +1241,7 @@ public CorrelationAttribute getCorrelationAttribute(CorrelationAttribute.Type ty
                 correlationAttribute.addInstance(artifactInstance);
             }
         } catch (SQLException ex) {
-            throw new EamDbException("Error getting notable artifact instances.", ex); // NON-NLS
+            throw new EamDbException(Bundle.AbstractSqlEamDb_exceptionMessage_errorGettingNotableArtifactInstances(), ex); // NON-NLS
```
APriestman: Exceptions should not be bundle messages

dgrove727: Removed

PR #3838, opened 2018-06-08
3819 7 zip depth fix

wschaeferB:


Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/EmbeddedFileExtractorIngestModule.java
```diff
@@ -44,12 +47,14 @@
 })
 public final class EmbeddedFileExtractorIngestModule extends FileIngestModuleAdapter {
 
-    static final String[] SUPPORTED_EXTENSIONS = {"zip", "rar", "arj", "7z", "7zip", "gzip", "gz", "bzip2", "tar", "tgz",}; // "iso"}; NON-NLS
+    private static final ConcurrentHashMap<Long, ConcurrentHashMap<Long, Archive>> mapOfDepthTrees = new ConcurrentHashMap<>();
```
bcarrier: Can you add a comment about what they keys are for this map?  I.e. "maps object id to map, which maps Y to Z".

PR #3847, opened 2018-06-11
query api to process each instances in the instances table.

rishwanth1995:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1526,6 +1526,45 @@ public boolean isArtifactKnownBadByReference(CorrelationAttribute.Type aType, St
...
+        ResultSet resultSet = null;
+        String tableName = EamDbUtil.correlationTypeToInstanceTableName(type);
+        StringBuilder sql = new StringBuilder();
+        sql.append("select * from ");
+        sql.append(tableName);
+
+        try {
+            preparedStatement = conn.prepareStatement(sql.toString());
+            resultSet = preparedStatement.executeQuery();
+            while (resultSet.next()) {
```
bcarrier: Let's change this so that the resultset is passed in once to the callback and it is up to the callback to iterate through.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1526,6 +1526,45 @@ public boolean isArtifactKnownBadByReference(CorrelationAttribute.Type aType, St
...
+                instanceTableCallback.process(resultSet);
+            }
+        } catch (SQLException ex) {
+            throw new EamDbException("Error getting notable artifact instances.", ex);
```
bcarrier: Reword.

PR #3848, opened 2018-06-11
3894: Cleanup  'Tag and Comment' dialog

raman-bt:


raman-bt:
We decided to reienstate the New Tag button on this dialog. The easiest way is for me to close this PR and and work on a new one.

PR #3855, opened 2018-06-14
3881: Normalize the content for index and query.

zhhl:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Keyword.java
```diff
@@ -57,7 +58,7 @@
      *                    than a substring.
      */
     Keyword(String searchTerm, boolean isLiteral, boolean isWholeWord) {
```
bcarrier: Can you have this constructor call the 5 parameter one instead of it assigning the values?

PR #3856, opened 2018-06-14
3870 progress bar callable

aziehl:
upstream develop with just the progress bar and callable wrapper changes for 3870.


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesChildren.java
```diff
@@ -0,0 +1,49 @@
...
+    protected Node[] createNodes(AbstractFile t) {
+        
+        if(t.isDir()){
+            //TODO log an error ???
+            
+        } else {
+            
+        }
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
+    }
```
rcordovano: I"m not sure what is going on here, but this looks broken.

Core/src/org/sleuthkit/autopsy/datamodel/Md5Node.java
```diff
@@ -53,16 +54,34 @@
...
+        public Children call() throws Exception {
+            //Check that the key has children,
+            //if it doesn't have children, return a leaf:
+            if (key.getMetadata().isEmpty()) {
+                return Children.LEAF;
+            } else {
+                return Children.create(new FileInstanceNodeFactory(key), true);
+            }
+        }
+    }
```
rcordovano: My understanding of lazy Children creation is that it is unrelated to lazy child node creation. Lazy Children creation seems to me to be for cases where you want to make different types of Children based on the context at the time of node creation, and it is not something provides a performance gain. Do we have a measurable gain from this? If not, it seems to add complexity for little or no gain.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesPanel.java
```diff
@@ -164,6 +166,9 @@ protected void done() {
         "CommonFilesPanel.search.results.titleAll=Common Files (All Data Sources)",
         "CommonFilesPanel.search.results.titleSingle=Common Files (Match Within Data Source: %s)",
         "CommonFilesPanel.search.results.pathText=Common Files Search Results",
+        "CommonFilesPanel.search.done.searchProgress1=Gathering Common Files Search Results.",
+        "CommonFilesPanel.search.done.searchProgress2=Generating Common Files Search Results.",
+        "CommonFilesPanel.search.done.searchProgress3=Displaying Common Files Search Results.",
```
rcordovano: These should me more descriptively named than just searchProgress1, 2, 3.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesPanel.java
```diff
@@ -202,7 +208,11 @@ private Long determineDataSourceId() {
 
             @Override
             @SuppressWarnings({"BoxedValueEquality", "NumberEquality"})
-            protected CommonFilesMetadata doInBackground() throws TskCoreException, NoCurrentCaseException, SQLException {
+            protected CommonFilesMetadata doInBackground() throws TskCoreException, NoCurrentCaseException, SQLException, EamDbException {
+                progress = ProgressHandle.createHandle(Bundle.CommonFilesPanel_search_done_searchProgress1());
+                progress.start();
+                progress.switchToIndeterminate();
+
```
rcordovano: This is interesting, I don't think we have a precedent for providing a progress bar in the lower right hand corner when doing search operations. I'm not sure the user will notice. We do have a new search feature where we provide a progress bar on the search dialog, but I don't want to undertake that additional complexity at this time. We need to think over whether this is the right approach.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesPanel.java
```diff
@@ -237,8 +247,8 @@ protected CommonFilesMetadata doInBackground() throws TskCoreException, NoCurren
             protected void done() {
                 try {
                     super.done();
-
-                    CommonFilesMetadata metadata = get();
+                    progress.setDisplayName(Bundle.CommonFilesPanel_search_done_searchProgress2());
```
rcordovano: I don't think that the difference between "gathering results" and "generating results" says much. I also suspect that the code between the generating and the displaying messages runs very quickly. Have you experienced otherwise?

aziehl: Removed the second text change.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesPanel.java
aziehl: Removed for now.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesPanel.java
aziehl: Updated variable names.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonFilesChildren.java
aziehl: Removed, this was an unmerged fragment

Core/src/org/sleuthkit/autopsy/datamodel/Md5Node.java
aziehl: Removed for now.

PR #3861, opened 2018-06-15
3888 Refactor other occurrences table

APriestman:


Core/src/org/sleuthkit/autopsy/centralrepository/AddEditCentralRepoCommentAction.java
```diff
@@ -102,7 +106,9 @@ public void addEditCentralRepoComment() {
             } catch (EamDbException ex) {
                 logger.log(Level.SEVERE, "Error connecting to Central Repository database.", ex);
             }
+            return centralRepoCommentDialog.getComment();
         }
+        return null;
```
bcarrier: To prevent NPEs, I'd rather that this always return the comment (or empty string if the user did not add a comment).

APriestman: The issue is that there are two pieces of data coming back from the dialog:
- Whether the user hit OK or Cancel
- What the comment is
I don't see any easy way to differentiate between "The user hit cancel so we're returning an empty string and should not update the display" and "The user removed the comment and hit OK so we're returning an empty string and should update the display". 

The other option may be to have OtherOccurrenceNodeData.getComment() pull from the original artifact instance if it is present. since I believe that instance should be updated in the course of the Add Comment action.

PR #3864, opened 2018-06-15
added try-catch for exception in jython module loader

rishwanth1995:


Core/src/org/sleuthkit/autopsy/python/JythonModuleLoader.java
```diff
@@ -66,12 +68,20 @@
...
-        PythonInterpreter interpreter = new PythonInterpreter();
-
+        PythonInterpreter interpreter = null;
+        // This method has previously thrown unchecked exceptions when it could not load because of non-latin characters.
+        try {
+            interpreter = new PythonInterpreter();
+        } catch (Exception ex) {
+            logger.log(Level.SEVERE, "Failed to load python Intepreter. Cannot load python modules", ex);
+            MessageNotifyUtil.Notify.show(Bundle.JythonModuleLoader_pythonInterpreterError_title(),Bundle.JythonModuleLoader_pythonInterpreterError_msg(), MessageNotifyUtil.MessageType.ERROR);
+            return objects;
```
rcordovano: @bcarrier and @rishwanth1995, this is not really the right approach. This code runs in the context of auto ingest jobs, not just ingest jobs and report generation attended by a user. As is generally the case with low-level code, the exception should be propagated to the higher level client code so that it can deal with it appropriately. In the ingest use case, this means that the ingest job should be stopped, since our policy is to stop ingest if a module configured to run cannot run. Also, note that this code runs not only in the context of an attended ingest job, but also in the context of an unattended ingest job, i.e., an auto ingest job, where creating a UI notification is inappropriate.

Core/src/org/sleuthkit/autopsy/python/JythonModuleLoader.java
```diff
@@ -91,7 +101,9 @@
                             if (line.startsWith("class ") && filter.accept(line)) { //NON-NLS
                                 String className = line.substring(6, line.indexOf("("));
                                 try {
-                                    objects.add(createObjectFromScript(interpreter, script, className, interfaceClass));
+                                    if (interpreter != null) {
```
rcordovano: This check adds a level of cyclomatic complexity that does not really do anything, given that we have already returned if there was an exception creating the interpreter, right?

Core/src/org/sleuthkit/autopsy/python/JythonModuleLoader.java
rcordovano: @bcarrier and @rishwanth1995, after further review of the code, I think that propagating the exception as I propose above is too much to undertake where we are in the current release cycle. I will make a story in JIRA about it. In the mean time, we need to make the logging of the notifications conditional on a check of RuntimeProperties.runningWithGUI, so that we do not do   notifications for auto ingest jobs.

PR #3883, opened 2018-06-21
3899 update docs for common files search, object detection, volatility, group by data source, and tagging

APriestman:


docs/doxygen-user/common_files.dox
```diff
@@ -0,0 +1,23 @@
...
+
+\image html common_files_dialog.png
+
+You can choose to find any files with multiple copies in the whole case, or specify that at least one of the copies has to be in the selected data source(s).
+
+\image html common_files_data_source.png
+
+You can also choose to restrict the search to only pictures and videos and/or documents.
+
+Once the search is run, the matching files are displayed in the results tab and are grouped by hash.
```
rcordovano: They are now grouped by number of instances, then hash.

docs/doxygen-user/volatility_dsp.dox
```diff
@@ -0,0 +1,32 @@
+/*! \page volatility_dsp_page Volatility Data Source Processor
+
+\section Overview
+
+The Volatility data source processor runs Volatility on a memory image and saves the individual module results. If the disk image associated with the memory image is also available, it will create Intesting Item artifacts linking the Volatility results to files in the disk image.
+
```
rcordovano: Intesting => Interesting

Might also want to say "individual Volatility module results" just to be clear we are talking about their modules, not ours.

docs/doxygen-user/volatility_dsp.dox
```diff
@@ -0,0 +1,32 @@
...
+\section Usage
+
+If you have a disk image associated with your memory image, ingest the disk image into the case first. Then go to "Add Data Source" and select "Memory Image File". 
+
```
rcordovano: Does the user have to ingest a corresponding disk image? From what follows below, it sounds like they will have critical errors if they don't, but I read this sub section to mean the ingest of the disk image is optional...

docs/doxygen-user/common_files.dox
APriestman: I know - it's on my todo list to update this.

docs/doxygen-user/volatility_dsp.dox
APriestman: They're non-critical errors that appear at the end of the data source wizard. Essentially if they don't have the image they'll get the Volatility module reports but nothing else.

docs/doxygen-user/common_files.dox
rcordovano: Okay.

docs/doxygen-user/volatility_dsp.dox
rcordovano: Sorry, I misread "non-critical" as "critical." Hopefully, my new glasses will be ready soon. ;-)

Let's go ahead and mention the consequences of not ingesting the data source first, please.

PR #3922, opened 2018-06-28
3994: Image gallery does not work with Postgres

raman-bt:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/grouping/GroupManager.java
```diff
@@ -351,11 +352,22 @@ public synchronized DrawableGroup removeFromGroup(GroupKey<?> groupKey, final Lo
...
+                        // Use the group_concat function to get a list of files for each mime type.  
+                        // This has different syntax on Postgres vs SQLite
+                        String groupConcatClause;
+                        if (DbType.POSTGRESQL == controller.getSleuthKitCase().getDatabaseType()) {
+                            groupConcatClause = " array_to_string(array_agg(obj_id), ',') as object_ids";
+                        }
+                        else {
+                            groupConcatClause = "select group_concat(obj_id) as object_ids";
+                        }
+                        String query = "select " + groupConcatClause + " , mime_type from tsk_files group by mime_type ";
```
esaunders: For SQLite doesn't this produce the query "select select group_concat(obj_id)..."?
Is the double "select" intentional for some reason? Perhaps SQLite doesn't care.

raman-bt:
Actually SQLite does care. Thanks for catching this hawkeye!!

PR #3926, opened 2018-06-29
3985 intra case test update

briangsweeney:


briangsweeney:
Codacy is likely going to complain of star syntax in imports.  If I get rid of the star it will complain of too many static imports.  If I don't go with either of those options, I must use the fully qualified name of a whole bunch of different variables which will make the code very ugly.  Leaving it as is, is the lesser of several evils, IMHO.

PR #3957, opened 2018-07-12
3788 intercase intracase merge

aziehl:
Feature complete for 3788 and 3905. Still requires code hygiene, code cleanup, and minor renames from 3904. Other Occurances Tab has been disabled again for intercase search.


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -384,6 +384,46 @@ public CorrelationCase getCaseByUUID(String caseUUID) throws EamDbException {
...
     }
+    /**
+     * Retrieves Case details based on Case ID
+     *
+     * @param caseID unique identifier for a case
+     *
+     * @return The retrieved case
+     */
+    @Override
+    public CorrelationCase getCaseById(int caseId) throws EamDbException {
```
bcarrier: Should not be public

aziehl: Should we extend the InstanceCallback interface instead and have query results join cases by id so we can retrieve the UUID and get the case that way? Same for datasources?

bcarrier: Oh yea, never mind.  You need it outside of the package and we are still keeping it from being exported from the NBM.

PR #3980, opened 2018-07-20
Changed form layout to conform with Add Keywords.

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/hashdatabase/Bundle_ja.properties
```diff
@@ -207,3 +206,4 @@ HashLookupSettingsPanel.nameLabel.text=\u30cf\u30c3\u30b7\u30e5\u30bb\u30c3\u30c
 HashLookupSettingsPanel.hashDatabasesLabel.text=\u30cf\u30c3\u30b7\u30e5\u30c7\u30fc\u30bf\u30d9\u30fc\u30b9\uff1a
 HashLookupSettingsPanel.importDatabaseButton.text=\u30c7\u30fc\u30bf\u30d9\u30fc\u30b9\u3092\u30a4\u30f3\u30dd\u30fc\u30c8
 HashDbCreateDatabaseDialog.databasePathLabel.text=\u30c7\u30fc\u30bf\u30d9\u30fc\u30b9\u30d1\u30b9\uff1a
+AddHashValuesToDatabaseDialog.okButton.text_2=\u30c7\u30fc\u30bf\u30d9\u30fc\u30b9\u306b\u30cf\u30c3\u30b7\u30e5\u3092\u8ffd\u52a0
```
wschaeferB: Do we want to use the existing Japanese text for the button which has been relabeled "OK" in English or do we want to use the English OK for the Japanese version as well?

My opinion would be to use OK as we appear to be doing elsewhere based on the occasional OK in Bundle_ja.properties file.

APriestman: It looks like every "okButton" in that file uses "OK", so I would agree that we should use that over the full 'add hash to database' string

PR #3987, opened 2018-07-25
4066 prevent duplicate hash set hits

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbIngestModule.java
```diff
@@ -90,11 +88,23 @@ private static synchronized IngestJobTotals getTotalsForIngestJobs(long ingestJo
...
         this.settings = settings;
-        skCase = Case.getCurrentCaseThrows().getSleuthkitCase();
     }
 
+    @Messages({
+        "HashDbIngestModule.noKnownBadHashDbSetMsg=No notable hash set.",
+        "HashDbIngestModule.knownBadFileSearchWillNotExecuteWarn=Notable file search will not be executed.",
+        "HashDbIngestModule.noKnownHashDbSetMsg=No known hash set.",
+        "HashDbIngestModule.knownFileSearchWillNotExecuteWarn=Known file search will not be executed."
+    })
```
rcordovano: I don't think we should be converting existing Bundle.properties messages to annotation form. If we do this, we lose the work that has already been done to translate the existing messages into Japanese.

millmanorama:
As long as you leave the japanese translations in the Bundle_ja.properites
and don't change the names of the keys, it should still work.

Core/src/org/sleuthkit/autopsy/modules/hashdatabase/HashDbIngestModule.java
dgrove727: What about exception messages that have been translated? I made two of those literal because we usually don't want to keep those as Bundle messages.

rcordovano: Please see revised general guidance in email.

Regarding the exception messages, there is a small subset of messages that get displayed to user. If the messages in question go into the log only, then the localization is indeed a mistake and you should proceed.

PR #3989, opened 2018-07-25
4072 remove cache

briangsweeney:


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributeValue.java
```diff
@@ -74,9 +74,8 @@ void addInstance(AbstractCommonAttributeInstance metadata) {
         this.fileInstances.add(metadata);
     }
 
-    void addFileInstanceMetadata(AbstractCommonAttributeInstance metadata, String caseName) {
+    void addFileInstanceMetadata(AbstractCommonAttributeInstance metadata) {
```
bcarrier: Do we need this method? Seems to do the same as addInstance().

briangsweeney: Looks like an auto-merge error or an oversight on our part.

PR #3990, opened 2018-07-25
1001 ig multiuser

raman-bt:
Interim PR for review.


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -411,6 +430,57 @@ public synchronized void reset() {
...
+        if ((null == db) || (null == sleuthKitCase)) {
+            return false;
+        }
+        
+        try {
+            Set<Long> knownDataSourceIds= db.getKnownDataSourceIds();
+            List<DataSource> dataSources = sleuthKitCase.getDataSources();
+            Set<Long> caseDataSourceIds = new HashSet<>();
+            dataSources.forEach((dataSource) -> {
+                caseDataSourceIds.add(dataSource.getId());
```
millmanorama: I would do this as 

Set<Long> caseDataSourceIds = dataSources.map(DataSource::getId).collect(Collectors.toSet());

But that is just me.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -411,6 +430,57 @@ public synchronized void reset() {
...
+        }
+ 
+    }
+    
+    /**
+     * Update the datasources table in drawable DB.
+     * 
+     * @return true if data_sources table is stale
+     */
+    void updateDataSourcesTable() {
```
rcordovano: I think this method can be private.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -688,6 +704,65 @@ private void insertOrUpdateFile(DrawableFile f, @Nonnull DrawableTransaction tr,
         }
     }
 
+
+   /**
+     * Gets all known data source object ids from data_sources table
+     *
+     * @return list of known data source object ids
+     */
+    public Set<Long> getKnownDataSourceIds() throws TskCoreException {
```
rcordovano: I think this should just be getDataSourceIds. "Known" is too overloaded a term to use here.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -688,6 +704,65 @@ private void insertOrUpdateFile(DrawableFile f, @Nonnull DrawableTransaction tr,
...
+        return ret;
+    }
+    
+    
+    /**
+     * Insert given data source object id into data_sources table
+     * 
+     * @param dsObjectId data source object id to insert
+     */
+    public void  insertDataSource(long dsObjectId) {
```
rcordovano: Perhaps indicate that this is insert if not exists. At least in the documentation.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -750,10 +828,14 @@ public void run() {
                 updateMessage("");
                 updateProgress(-1.0);
             }
-            cleanup(true);
+            cleanup(taskCompletionStatus);
         }
 
         abstract ProgressHandle getInitialProgressHandle();
+        
+        void setTaskCompletionStatus(boolean status) {
```
rcordovano: I think this should be protected, not package access.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -794,10 +877,14 @@ void processFile(AbstractFile f, DrawableDB.DrawableTransaction tr) {
                     if (FileTypeUtils.hasDrawableMIMEType(f)) {  //supported mimetype => analyzed
                         taskDB.updateFile(DrawableFile.create(f, true, false), tr);
                     } else { //unsupported mimtype => analyzed but shouldn't include
+                        // if mimetype of the file hasnt been ascertained, ingest might not have completed yet.
+                        if (null == f.getMIMEType()) {
+                            this.setTaskCompletionStatus(false);
+                        }
```
rcordovano: AbstractFile.getMIMEType will never return null after a call to FileTypeUtils.hasDrawableMIMEType, so the call to setTaskCompletionStatus will never be executed and we do not need the taskCompletionStatus stuff. @bcarrier, apparently this was your idea, what is the behavior you are trying to create?

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryTopComponent.java
```diff
@@ -88,6 +89,31 @@
...
+    /**
+     * Returns the top component window.
+     * 
+     * @return Image gallery top component window, null if it's not open
+     */
+    @ThreadConfined(type = ThreadConfined.ThreadType.JFX)
+    public static TopComponent getTopComponent() {
+        return WindowManager.getDefault().findTopComponent(PREFERRED_ID);  
+    }
+
```
rcordovano: The annotation on these two methods are not correct. The methods are called in the EDT.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/actions/OpenAction.java
```diff
@@ -172,7 +172,15 @@ public void performAction() {
...
-                    ImageGalleryController.getDefault().setListeningEnabled(true);
+                    
+                    if (currentCase.getCaseType() == Case.CaseType.SINGLE_USER_CASE) {
+                        // toggling listening to ON automatically triggers a rebuild
+                        ImageGalleryController.getDefault().setListeningEnabled(true);
+                    }
+                    else {
+                        ImageGalleryController.getDefault().rebuildDB();
+                    }
+
```
rcordovano: We should comment this to explain the two different behaviors, as discussed in conversation.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -411,6 +430,57 @@ public synchronized void reset() {
...
 
+    /**
+     * Checks if the datasources table in drawable DB is stale.
+     * 
+     * @return true if datasources table is stale
+     */
+    boolean isDataSourcesTableStale() {
+        
+        // no current case open to check
+        if ((null == db) || (null == sleuthKitCase)) {
```
rcordovano: These two instance variables are accessed by multiple threads, but they are only partially guarded using the intrinsic monitor of the ImageGalleryController object. The most basic solution is to be sure that we never access these variables directly unless we are holding the lock, which is easily done by ALWAYS calling the properly synchronized getDatabase and the corresponding getter for the case database reference. I am a bit concerned that we rely on finalize to close the drawable db rather than closing it when the current case is closed, so please make a story about this so we can look into the consequences later, thanks!

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
raman-bt: We realized that hasDrawableMIMEType() will determine the mimetype if unknown.  
So we changed the condition to:
if ( null != f.getMIMEType() && FileTypeUtils.hasDrawableMIMEType(f)) {
...
}

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -890,28 +979,32 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+                                            //this file should be included and we don't already know about it from hash sets (NSRL)
+                                            queueDBTask(new UpdateFileTask(file, db));
+                                        } else if (FileTypeUtils.getAllSupportedExtensions().contains(file.getNameExtension())) {
+                                            //doing this check results in fewer tasks queued up, and faster completion of db update
+                                            //this file would have gotten scooped up in initial grab, but actually we don't need it
+                                            queueDBTask(new RemoveFileTask(file, db));
+                                        }
                                     }
+                                } catch (TskCoreException | FileTypeDetector.FileTypeDetectorInitException ex) {
+                                    //TODO: What to do here?
```
rcordovano: Please remove the TODO.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -712,16 +783,22 @@ public void run() {
...
-                    if (isCancelled() || Thread.interrupted()) {
+                    if (isCancelled()) {
                         LOGGER.log(Level.WARNING, "Task cancelled: not all contents may be transfered to drawable database."); //NON-NLS
                         progressHandle.finish();
                         break;
                     }
+                    
+                    if (Thread.interrupted()) {
+                        LOGGER.log(Level.WARNING, "BulkTransferTask interrupted. Ignoring it to update the contents of drawable database."); //NON-NLS
+                    }
```
rcordovano: Please write a JIRA about this - the interrupt should be equivalent to a cancellation, so we need to find out how this thread is being unexpectedly interrupted. When the JIRA is created please add a comment of the form "TODO (EUR-XXXX)" to explain why this code is the way is it.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -712,16 +783,22 @@ public void run() {
...
                 updateProgress(0.0);
+                
+               taskCompletionStatus = true;
 
                 //do in transaction
                 DrawableDB.DrawableTransaction tr = taskDB.beginTransaction();
                 int workDone = 0;
                 for (final AbstractFile f : files) {
-                    if (isCancelled() || Thread.interrupted()) {
+                    if (isCancelled()) {
```
rcordovano: I think that the taskCompletionStatus flag should be false when the task is cancelled

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -783,21 +865,28 @@ protected void cleanup(boolean success) {
...
+                    if ( null != f.getMIMEType() && FileTypeUtils.hasDrawableMIMEType(f)) {
                         taskDB.updateFile(DrawableFile.create(f, true, false), tr);
-                    } else { //unsupported mimtype => analyzed but shouldn't include
+                    }
+                    else { //unsupported mimtype => analyzed but shouldn't include
+                        
+                        // if mimetype of the file hasn't been ascertained, ingest might not have completed yet.
+                        if (null == f.getMIMEType()) {
+                            this.setTaskCompletionStatus(false);
+                        }
```
rcordovano: The logic has changed here, I suspect incorrectly. If the MIME type is not set, we should not remove the file from the drawable database, since we will want to process it later when ingest completes.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -890,28 +979,32 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+                                            "Unable to determine if file is drawable and not known.  Not making any changes to DB.  See the logs for details.");
                                 }
-                            } catch (TskCoreException | FileTypeDetector.FileTypeDetectorInitException ex) {
-                                //TODO: What to do here?
-                                LOGGER.log(Level.SEVERE, "Unable to determine if file is drawable and not known.  Not making any changes to DB", ex); //NON-NLS
-                                MessageNotifyUtil.Notify.error("Image Gallery Error",
-                                        "Unable to determine if file is drawable and not known.  Not making any changes to DB.  See the logs for details.");
                             }
+                        } else {   //TODO: keep track of what we missed for later
+                            setStale(true);
```
rcordovano: Please remove the TODO.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -943,12 +1036,14 @@ public void propertyChange(PropertyChangeEvent evt) {
...
-                    } else {//TODO: keep track of what we missed for later
-                        setStale(true);
+                    //For a data source added on the local node, prepopulate all file data to drawable database
+                    if (((AutopsyEvent) evt).getSourceType() == AutopsyEvent.SourceType.LOCAL) {
+                        Content newDataSource = (Content) evt.getNewValue();
+                        if (isListeningEnabled()) {
+                            queueDBTask(new PrePopulateDataSourceFiles(newDataSource, ImageGalleryController.this, getDatabase(), getSleuthKitCase()));
+                        } else {//TODO: keep track of what we missed for later
+                            setStale(true);
+                        }
```
rcordovano: Please remove the TODO.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/actions/OpenAction.java
rcordovano: Excellent comment!!

PR #3993, opened 2018-07-26
4070 remove db methods

aziehl:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1894,57 +1895,17 @@ public void processInstanceTable(CorrelationAttribute.Type type, InstanceTableCa
...
      *
      * @param type EamArtifact.Type to search for
-     * @param correlationCase  CorrelationCase to filter by
      * @param instanceTableCallback callback to process the instance
+     * @param whereClause query string to execute
      * @throws EamDbException
      */
     @Override
-    public void processCaseInstancesTable(CorrelationAttribute.Type type, CorrelationCase correlationCase, InstanceTableCallback instanceTableCallback) throws EamDbException {
+    public void processInstanceTableWhere(CorrelationAttribute.Type type, String whereClause, InstanceTableCallback instanceTableCallback) throws EamDbException {
```
bcarrier: I'd just call this processInstanceTable()

Core/src/org/sleuthkit/autopsy/commonfilesearch/IntraCasePanel.java
```diff
@@ -40,7 +41,7 @@
     private boolean singleDataSource;
     private String selectedDataSource;
     private ComboBoxModel<String> dataSourcesList = new DataSourceComboBoxModel();
-    private Map<Long, String> dataSourceMap;
+    private final Map<Long, String> dataSourceMap;
```
bcarrier: Can you add a comment about what the keys and values are?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
aziehl: There already is a processInstanceTable() which was added with the callback Interface, unless this method should supplant that. processInstanceTable() is a select * from <given table type>.

aziehl:
Added comments to IntraCasePanel. Codeacy wasn't able to run after previous commit, but this was limited in scope so I don't expect much. Otherwise, ready to be merged.

PR #4005, opened 2018-08-01
4092: Refactored SQLite parsing functionality out of SQLiteViewer and into SQLiteReader

dannysmyda:


Core/src/org/sleuthkit/autopsy/sqlitereader/SQLiteReader.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.stream.Collectors;
+import org.apache.commons.io.FilenameUtils;
+import org.openide.util.NbBundle;
+
+/**
+ * Abstracts the need to open (and close) statements and result sets from the
+ * JDBC. Classes using the SQLiteExtractor must maintain a reference to the 
+ * Connection object generated in getDatabaseConnection.
+ */
+public final class SQLiteReader {
```
rcordovano: I would like to see more user-friendly class documentation, i.e., something that describes the purpose and usage of the class at a higher level of abstraction for potential clients.

Core/src/org/sleuthkit/autopsy/sqlitereader/SQLiteReader.java
```diff
@@ -0,0 +1,191 @@
...
+    private SQLiteReader(){}
+
+    public static Connection getDatabaseConnection(String dbPath) 
+            throws ClassNotFoundException, SQLException {
+        
+        // Load the SQLite JDBC driver, if necessary.
+        Class.forName("org.sqlite.JDBC"); //NON-NLS  
+        return DriverManager.getConnection(
+                "jdbc:sqlite:" + dbPath); //NON-NLS
+    }
```
rcordovano: I think that we should relieve clients of the need to deal with the database connection. Instead, lets convert the class methods to instance methods, make the class implement AutoCloseable, and have the instances of the class hide and manage the connection on behalf of clients. I would add the writing of the database files to disk and the opening of the connection to the constructor. The user can supply the location for the write as a constructor argument.

Core/src/org/sleuthkit/autopsy/sqlitereader/SQLiteReader.java
```diff
@@ -0,0 +1,191 @@
...
+                        + " OFFSET " + Integer.toString(startRow - 1))) {
+            return resultSetToList(resultSet);
+        }
+    }
+    
+    @NbBundle.Messages({"SQLiteExtractor.exportTableToCsv.write.errText=Failed to export table content to csv file.",
+                        "SQLiteExtractor.exportTableToCsv.FileName=File name: ",
+                        "SQLiteExtractor.exportTableToCsv.TableName=Table name: "
+    })
+    public static void exportTableToCSV(File file, String tableName,
```
rcordovano: I think this method should stay in the file viewer, it is specific to its use case.

dannysmyda:
Hopefully this looks better!

PR #4008, opened 2018-08-02
1007: Ingest hangs with SQLite DB when Image Gallery is enabled

raman-bt:
- undo the use transaction for updates to groups table


bcarrier:
No longer needed after TSK issue 1297

PR #4009, opened 2018-08-02
Cleanup.

dgrove727:


Core/src/org/sleuthkit/autopsy/directorytree/ExtractAction.java
```diff
@@ -289,22 +315,23 @@ protected void done() {
...
+         * @return The number of work units.
+         */
+        /*
+         * private int calculateProgressBarWorkUnits(AbstractFile file) { int
+         * workUnits = 0; if (file.isFile()) { workUnits += file.getSize(); }
+         * else { try { for (Content child : file.getChildren()) { if (child
+         * instanceof AbstractFile) { workUnits +=
+         * calculateProgressBarWorkUnits((AbstractFile) child); } } } catch
+         * (TskCoreException ex) { logger.log(Level.SEVERE, "Could not get
+         * children of content", ex); //NON-NLS } } return workUnits;
```
rcordovano: If we don't use this code we should delete it, not comment it out.

dgrove727: I meant to mention this up-front. I commented it out because Brian has some comments in the class that seem to relate to this method. So I felt funny about removing the method but leaving his notes. Starting on line 284 of the changed code, he has the following:

/*
 * @@@ Add back in -> Causes exceptions int workUnits = 0; for
 * (FileExtractionTask task : extractionTasks) { workUnits +=
 * calculateProgressBarWorkUnits(task.source); }
 * progress.switchToDeterminate(workUnits);
 */

PR #4010, opened 2018-08-02
3964 test zip bomb rev2

wschaeferB:
Replaces PR #3931


Core/test/qa-functional/src/org/sleuthkit/autopsy/ingest/EmbeddedFileTest.java
```diff
@@ -92,31 +94,47 @@ public void tearDown() {
...
+                    assertEquals("Password protected zip file " + file.getName() + " has incorrect number of artifacts", 1, artifacts.size());
                     for (BlackboardArtifact artifact : artifacts) {
-                        assertEquals(artifact.getArtifactTypeID(), BlackboardArtifact.ARTIFACT_TYPE.TSK_ENCRYPTION_DETECTED.getTypeID());
+                        assertEquals("Artifact for password protected zip file " + file.getName() + " has incorrect type ID", artifact.getArtifactTypeID(), BlackboardArtifact.ARTIFACT_TYPE.TSK_ENCRYPTION_DETECTED.getTypeID());
                         passwdProtectedZips++;
                     }
+                } else if (file.getName().equalsIgnoreCase(depthZipBomb) || file.getName().equalsIgnoreCase(ratioZipBomb)){
+                    ArrayList<BlackboardArtifact> artifacts = file.getAllArtifacts();
+                    assertEquals("Zip bomb " + file.getName() + " has incorrect number of artifacts", 1, artifacts.size());
+                    for (BlackboardArtifact artifact : artifacts) {
```
dgrove727: I always think a good best practice to follow is to put "List" at the name of a List object. So here, you would have 'artifactsList'. It makes it stand out a little better so the person reading the code doesn't misread it.

APriestman: I've personally always found it clear enough to interpret it as plural -> some kind of collection.

rcordovano: @wschaeferB and @dgrove727, I agree with @APriestman . I find reading "artifacts" to be easier than reading "artifactsList." To me its a matter of level of abstraction. Reading the code to understand it, all I need to know is there is some collection of artifacts. If I care about the type of the collection, I look at the variable declaration, although that is rarely necessary. I try to make variable names be like what we would use in conversation or pseudo-code, to make code closer to speech...

PR #4015, opened 2018-08-03
warn user for case folder, if they are running autopsy in target system

rishwanth1995:


Core/src/org/sleuthkit/autopsy/coreutils/PathValidator.java
```diff
@@ -39,11 +41,45 @@ public static boolean isValid(String path, Case.CaseType caseType) {
...
         } else {
             // single user case - no validation needed
-        }
+            }
 
         return true;
     }
-
+    
+    public static boolean isCasedataPersistable(String path) {
```
rcordovano: Let's make the names a little more consistent and descriptive:

1. Rename isValid to isValidForMultiUserCase. Because isValid is public, we will need to keep the old isValid and deprecate it, making it call the "new" method.
2. Rename isCaseDataPersitable to isValidForRunningOnTarget.

PR #4023, opened 2018-08-03
4103 added icons back to AID 2.0 as part of figuring out how to do 4095

wschaeferB:


dannysmyda:
Check the Codacy PR suggestions

PR #4037, opened 2018-08-10
1009: Rebuild IG at data-source level

raman-bt:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -403,7 +400,15 @@ public synchronized void setCase(Case theNewCase) {
      * 
      */
     public void rebuildDB() {
-          queueDBTask(new CopyAnalyzedFiles(instance, db, sleuthKitCase));
+         
+        // queue a rebuild task for each stale data source
+        Set<Long> staleDataSources = getStaleDataSourceIds();
+        if (!staleDataSources.isEmpty()) {
```
millmanorama: this is already a no-op if stateDataSources is empty.  I would skip the explicit test my self...

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -200,11 +202,6 @@ void setStale(Boolean b) {
         Platform.runLater(() -> {
             stale.set(b);
         });
-        try {
-            new PerCaseProperties(Case.getCurrentCaseThrows()).setConfigSetting(ImageGalleryModule.getModuleName(), PerCaseProperties.STALE, b.toString());
-        } catch (NoCurrentCaseException ex) {
-            Logger.getLogger(ImageGalleryController.class.getName()).log(Level.WARNING, "Exception while getting open case."); //NON-NLS
```
millmanorama: Am I correct that we are thowing out the idea of staleness for the whole database and instead keeping track per datasource.  That means the only property left in PerCaseProperties is 'enabled'  do we want a whole class just for this?  Is there somewhere else it could go?

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -403,7 +400,11 @@ public synchronized void setCase(Case theNewCase) {
      * 
      */
     public void rebuildDB() {
-          queueDBTask(new CopyAnalyzedFiles(instance, db, sleuthKitCase));
+        // queue a rebuild task for each stale data source
```
millmanorama: personally, I would inline staleDatataSources.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -437,48 +438,57 @@ public synchronized void reset() {
      * @return true if datasources table is stale
      */
     boolean isDataSourcesTableStale() {
+        return (getStaleDataSourceIds().isEmpty() == false);
+    }
```
millmanorama: I have started using apache CollectionUtils.isNotEmpty() to avoid the == false and the codacy warning it often generates.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
```diff
@@ -739,27 +752,35 @@ public void run() {
...
         final DrawableDB taskDB;
         final SleuthkitCase tskCase;
+        final long dataSourceObjId;
 
         ProgressHandle progressHandle;
-        
         private boolean taskCompletionStatus;
 
-        BulkTransferTask(ImageGalleryController controller, DrawableDB taskDB, SleuthkitCase tskCase) {
+        BulkTransferTask(long dataSourceObjId, ImageGalleryController controller, DrawableDB taskDB, SleuthkitCase tskCase) {
```
millmanorama: I wonder if taskDB is too close to tskDB at a quick glance

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
raman-bt: Well we got rid of tracking the DB staleness as a whole via the property.  The DB is stale is any of the data sources is stale (or there are new data sources not synced)

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryController.java
raman-bt: Possibly.  I'll make a note of it and clean it up in a separate cleanup commit as part of EUR-1010.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -763,20 +773,20 @@ private void insertOrUpdateFile(DrawableFile f, @Nonnull DrawableTransaction tr,
...
-        Set<Long> ret = new HashSet<>();
+        Map<Long, DrawableDbBuildStatusEnum> map = new HashMap<>();
         dbReadLock();
         try {
             statement = con.createStatement();
-            rs = statement.executeQuery("SELECT ds_obj_id FROM datasources "); //NON-NLS
+            rs = statement.executeQuery("SELECT ds_obj_id, drawable_db_build_status FROM datasources "); //NON-NLS
             while (rs.next()) {
-                ret.add(rs.getLong(1));
+                map.put(rs.getLong(1), DrawableDbBuildStatusEnum.valueOf(rs.getString(2)));
```
millmanorama: I strongly advise against using indices to get the values out of a result set.  It is easy to change the query and forget to change the indices leading to bugs.  In addition I am not sure if the column order is guaranteed to be consistent across JDBC drivers.

raman-bt:
I addressed the critical comment.  Couldn't get the CollectionUtils.isNotEmpty() to work , I have to figure out how to add the library.

millmanorama:
It is already pulled in to CoreLibs, but the package is not exported from
the module.

raman-bt:
@rcordovano: this is the unmerged PR I was talking about this am

rcordovano:
Thanks, @raman-bt. I am waiting for an approval from @millmanorama. @millmanorama, if you now approve this PR, please mark it so as the reviewer, thanks!

PR #4040, opened 2018-08-14
Post artifacts fallout: round 1

millmanorama:
The first round of modifying Modules to use Blackboard.postArtifact(s) instead of 
index and then firing a ModuleDataEvent


Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -279,37 +278,41 @@ private void createPlasoArtifacts(String plasoDb, DataSourceIngestModuleProgress
...
                         resolvedFile = image;
                     }
-                    
+
                     try {
                         Collection<BlackboardAttribute> bbattributes = new ArrayList<>();
-                        bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DATETIME, Bundle.PlasoIngestModule_event_datetime(), resultSet.getLong("epoch_date")));
-                        bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DESCRIPTION, Bundle.PlasoIngestModule_event_description(), resultSet.getString("description")));
+                        bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DATETIME, MODULE_NAME, resultSet.getLong("epoch_date")));
+                        bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DESCRIPTION, MODULE_NAME, resultSet.getString("description")));
```
raman-bt: Is MODULE_NAME a suitable replacement for the strings here?

millmanorama: I am not sure.  Almost all modules put the module name here.  The old values were "Event Date Time" and "Event Description" which I didn't think were very informative.

PR #4041, opened 2018-08-14
3868 normalize cr

briangsweeney:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttribute.java
```diff
@@ -117,11 +113,16 @@ public String getCorrelationValue() {
...
+     * been set (for data normalization purposes).
+     * 
      * @param correlationValue the correlationValue to set
      */
-    public void setCorrelationValue(String correlationValue) {
-        // Lower-case all values to normalize and improve correlation hits, going forward make sure this makes sense for all correlation types
-        this.correlationValue = correlationValue.toLowerCase();
+    public void setCorrelationValue(String correlationValue) throws CorrelationAttributeNormalizationException {
+        if(this.getCorrelationType() == null){
+            throw new IllegalStateException("Correlation Type must be set before calling setCorrelationValue");
```
APriestman: While this does seem like an illegal state, unchecked exceptions tend to cause trouble. It's fine to use one of the normal CR ones.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Verify MD5 is the correct length and values. Make lower case.
+     */
+    private static String normalizeMd5(String data) throws CorrelationAttributeNormalizationException {
+        final String errorMessage = "Data purporting to be an MD5 was found not to comform to expected format: %s";
+        if(data == null){
+            throw new CorrelationAttributeNormalizationException(String.format(errorMessage, data));
+        }
+        final String validMd5Regex = "^[a-fA-F0-9]{32}$";
```
APriestman: There's no point in matching the upper case hex digits since you convert the string to lower case before applying the regex.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
```diff
@@ -0,0 +1,164 @@
...
+        String phoneNumber = data.replaceAll("[^0-9\\+]", "");
+        if(phoneNumber.matches("\\+?[0-9]+")){
+            return phoneNumber;
+        } else {
+            throw new CorrelationAttributeNormalizationException(String.format("Data was expected to be a valid phone number: %s", data));
+        }
+    }
+
+    /**
+     * USB ID is of the form: hhhh:hhhh where h is a hex digit.  Convert to lower case.
```
APriestman: Is this right? I'm not sure it matches my sample data

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -1625,10 +1611,9 @@ public Long getCountArtifactInstancesKnownBad(CorrelationAttribute.Type aType, S
      * @throws EamDbException
      */
     @Override
-    public List<String> getListCasesHavingArtifactInstancesKnownBad(CorrelationAttribute.Type aType, String value) throws EamDbException {
-        if (aType == null) {
-            throw new EamDbException("Correlation type is null");
-        }
+    public List<String> getListCasesHavingArtifactInstancesKnownBad(CorrelationAttribute.Type aType, String value) throws EamDbException, CorrelationAttributeNormalizationException {
+        
+        String normalizeValue = CorrelationAttributeNormalizer.normalize(aType, value);
```
APriestman: Please change all the "normalizeValue" variables to "normalizedValue" for consistency

Core/nbproject/project.properties
```diff
@@ -15,6 +15,7 @@ file.reference.postgresql-9.4.1211.jre7.jar=release/modules/ext/postgresql-9.4.1
 file.reference.Rejistry-1.0-SNAPSHOT.jar=release/modules/ext/Rejistry-1.0-SNAPSHOT.jar
 file.reference.sevenzipjbinding-AllPlatforms.jar=release/modules/ext/sevenzipjbinding-AllPlatforms.jar
 file.reference.sevenzipjbinding.jar=release/modules/ext/sevenzipjbinding.jar
+file.reference.SparseBitSet-1.1.jar-1=release/modules/ext/SparseBitSet-1.1.jar
```
APriestman: I think this is a duplicate

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -209,11 +197,7 @@ private static CorrelationAttribute getCorrelationAttributeFromBlackboardArtifac
             return null;
         }
 
-        if (null != value) {
-            return new CorrelationAttribute(correlationType, value);
-        } else {
-            return null;
-        }
+        return new CorrelationAttribute(correlationType, value);
```
APriestman: I think we still need that null check/null return. The idea is that we return null if the correlation type doesn't match any data for type of artifact. It's not an error that a keyword hit doesn't have an MD5 hash.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -179,17 +181,23 @@ private void showCommonalityDetails() {
...
-                    msg.append(Bundle.DataContentViewerOtherCases_correlatedArtifacts_byType(percentage,
+                    try {
+                        percentage = dbManager.getFrequencyPercentage(eamArtifact);
+                        msg.append(Bundle.DataContentViewerOtherCases_correlatedArtifacts_byType(percentage,
                             eamArtifact.getCorrelationType().getDisplayName(),
                             eamArtifact.getCorrelationValue()));
+                    } catch (CorrelationAttributeNormalizationException ex) {
+                        String message = String.format("Unable to determine commonality for artifact %s", eamArtifact.toString());
+                        LOGGER.log(Level.SEVERE, message, ex);
+                        Exceptions.printStackTrace(ex);
```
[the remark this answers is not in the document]
briangsweeney: It's the getFrequencyPercentage count that fails in this scope so that's why I used that message, but I guess the underlying issue is failed validation.  The message attached to the exception will contain notes on validation failure, and the log message will explain the effects of this failure in this particular case.  This is the approach I've generally tried to take throughout this feature.  Does this seem more reasonable now, or do you still think I should change it?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
briangsweeney: I found (on the dubious internet) USB IDs in a variety of formats, thus the flexible validation you see in the regex.  Can I see your test data, or do you have any suggestions on how to create my own?

Core/nbproject/project.properties
briangsweeney: I blame netbeans for this one :(

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
APriestman: I found this site which matches more of what I see:
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/standard-usb-identifiers
Mine look more like this: 4&24d6eb65&0
If we can't find a good source it might be better to skip validation than to possibly lose data. There aren't usually that many device ID attributes anyway.

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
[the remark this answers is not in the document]
APriestman: Yeah you can leave that as-is

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
APriestman: If you want to switch to an Optional instead of returning null, that would work too.

briangsweeney: Optional is new to me.  So the method signature would become like this: private static Optional<CorrelationAttribute> getCorrelationAttributeFromBlackboardArtifact(...) and then clients would call the get() function?

APriestman: I think you do an isPresent() check and then get() if it's true. There's an example in your normalize() method

briangsweeney: Right, I was mostly asking if it seemed ok to change the method signature and clients of that code?

APriestman: Ah sorry. Yes central repo isn't exported to the public API, so we can change anything we want

briangsweeney: Cool, I'm happy to do that; I'm just hesitant to go making too much in the way of outward facing changes.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
briangsweeney: From MS: USB\VID_v(4)&PID_d(4)&REV_r(4)_

So in your interpretation of this spec, are they just dropping the insignificant digits, so it really should specify up to 4 digits and not exactly four digits?

It goes on to discuss these various INF models.  Should we be expecting and accepting those as well?

bcarrier:
@briangsweeney Unfortunately, this has some merge conflicts after the refactor story was merged in.

briangsweeney:
@bcarrier That's ok I have a couple more changes to make here anyway before we close this.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            throw new CorrelationAttributeNormalizationException(String.format("Data was expected to be a valid email address: %s", data));
+        }
+    }
+
+    /**
+     * Verify it is only numbers and '+'. Strip spaces, dashes, and parentheses.
+     */
+    private static String normalizePhone(String data) throws CorrelationAttributeNormalizationException {
+        String phoneNumber = data.replaceAll("[^0-9\\+]", "");
```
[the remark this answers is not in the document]
briangsweeney: Not quite, I think.  The subsequent line also ensures that the '+' appears one or 0 times and that the digits appear at least once.  In any case, what you propose seems like better validation.

briangsweeney:
Just pushed - still merging and testing; nobody bother to review/merge this just yet...

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
```diff
@@ -0,0 +1,163 @@
...
+            return data.toLowerCase();
+        } else {
+            throw new CorrelationAttributeNormalizationException(String.format("Data was expected to be a valid email address: %s", data));
+        }
+    }
+
+    /**
+     * Verify it is only numbers and '+'. Strip spaces, dashes, and parentheses.
+     */
+    private static String normalizePhone(String data) throws CorrelationAttributeNormalizationException {
```
APriestman: This and normalizeUsbId should have null checks. Alternately, you could do this up front before the switch statement in normalize().

briangsweeney: I think the normalize email one actually gracefully handles null in the library I'm using - regardless I just put the check up front before the switch just for simplicity.

APriestman: I was referring to the phone number one - it threw an npe

APriestman:
Please resolve the appveyor/travis errors 
unreported exception CorrelationAttributeNormalizationException; must be caught or declared to be thrown
 [nb-javac]                 int frequencyPercentage = eamDb.getFrequencyPercentage(new CorrelationAttributeInstance(fileAttributeType, value.getValue()));
 [nb-javac]                                                                        ^
 [nb-javac] C:\projects\autopsy\Core\src\org\sleuthkit\autopsy\commonfilesearch\CommonAttributeSearchResults.java:116: error: unreported exception CorrelationAttributeNormalizationException; must be caught or declared to be thrown
 [nb-javac]                 int frequencyPercentage = eamDb.getFrequencyPercentage(new CorrelationAttributeInstance(fileAttributeType, value.getValue()));

PR #4049, opened 2018-08-16
 update ThunderbirdMboxFileIngestModule for postArtifacts and cleanup

millmanorama:
should come after PR 4040


millmanorama:
conflict solved

PR #4052, opened 2018-08-16
4130 add excel to selector module

dannysmyda:


Core/src/org/sleuthkit/autopsy/tabulardatareader/ExcelReader.java
```diff
@@ -0,0 +1,282 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import com.monitorjbl.xlsx.StreamingReader;
+import org.apache.poi.hssf.OldExcelFormatException;
+
+
+/**
+ * Reads excel files and implements the abstract reader api for interfacing with the 
+ * content. Supports .xls and .xlsx files.
+ */
+public class ExcelReader extends AbstractReader {
```
wschaeferB: I think we want to make the ExcelReader and other children of AbstractReader such as the SQLiteReader final, however I am not certain.

PR #4055, opened 2018-08-17
3863 remove correlationattribute

aziehl:


aziehl:
Thanks Ann! I've pushed fixes for what you called out.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
```diff
@@ -361,16 +361,16 @@ public int hashCode() {
...
+            str.append("(id=")
+                    .append(getId());
+            str.append(", displayName=")
+                    .append(getDisplayName());
+            str.append(", dbTableName=")
+                    .append(getDbTableName());
+            str.append(", supported=")
+                    .append(isSupported().toString());
+            str.append(", enabled=")
+                    .append(isEnabled().toString());
```
APriestman: I think what codacy is complaining about is the reuse of the "str" variable. You should be able to chain all of these together in one statement (so you'll only use "str" once at the beginning).

PR #4063, opened 2018-08-23
Data source filter ui -sync list as datasources are added and removed

millmanorama:


millmanorama:
this is included in https://github.com/sleuthkit/autopsy/pull/4075

PR #4065, opened 2018-08-23
4114 add has comment column

wschaeferB:
Requires Sleuthkit PR https://github.com/sleuthkit/sleuthkit/pull/1312 to avoid Comment being the name of two columns.


Core/src/org/sleuthkit/autopsy/commonfilesearch/CaseDBCommonAttributeInstanceNode.java
```diff
@@ -18,28 +18,30 @@
...
-    
+
     private final String caseName;
     private final String dataSource;
 
     /**
      * Create a node which can be used in a multilayer tree table and is based
      * on an <code>AbstractFile</code>.
-     * 
+     *
```
dannysmyda: Should we include the missing param with some descriptions?

Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
```diff
@@ -82,6 +86,7 @@
 
     private static final long serialVersionUID = 1L;
     private static final Logger LOGGER = Logger.getLogger(DataResultViewerTable.class.getName());
+    private static final ImageIcon COMMENT_ICON = new ImageIcon(ImageUtilities.loadImage("org/sleuthkit/autopsy/images/notepad16.png", false));
```
dannysmyda: I'm imagining a situation where this icon is reused somewhere else and in the future is eventually changed, then links like these will be broken. I might not fully understand the use case, but would it be bloat to have an enum for retrieving the icon location string?

wschaeferB: Management of our icons in a more organized manner would be nice as talked about, but if done should probably be done as a larger story so we can do it everywhere and in the process reduce icons which we have duplicate copies of. 

I am going to split the string off into its own variable because I believe it makes it a bit easier to read.

PR #4066, opened 2018-08-23
4102 filter by frequency

briangsweeney:


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributeSearchResults.java
```diff
@@ -48,32 +58,103 @@
...
+        for(Entry<Integer, List<CommonAttributeValue>> listOfValues : Collections.unmodifiableMap(this.instanceCountToAttributeValues).entrySet()){
+            
+            final Integer key = listOfValues.getKey();
+            final List<CommonAttributeValue> values = listOfValues.getValue();
+            
+            for(CommonAttributeValue value : values){
+                
+                int frequencyPercentage = eamDb.getFrequencyPercentage(new CorrelationAttributeInstance(fileAttributeType, value.getValue()));
+                
+                if(frequencyPercentage < minimumPercentageThreshold){
```
APriestman: Is this comparison backwards? I thought the idea was to exclude files that have a frequency greater than the threshold - i.e., exclude files that are common and appear many times in the central repo.

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributePanel.java
```diff
@@ -587,76 +667,91 @@ private void cancelButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-
...
+            this.errorText.setText(this.errorManager.getErrors().get(0));
             this.errorText.setVisible(true);
         } else {
             this.searchButton.setEnabled(true);
             this.errorText.setVisible(false);
         }
     }
 
-    private void manageCheckBoxState() {
+    private void enablePercentageOptions() {
```
APriestman: In this and the disable section, you need to add the second half of the text (jLabel1)

PR #4071, opened 2018-08-28
Input scan time fix

eugene7646:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -1398,10 +1398,6 @@ private void addPendingJob(Manifest manifest, AutoIngestJobNodeData nodeData) th
             AutoIngestJob job;
             if (nodeData.getVersion() == AutoIngestJobNodeData.getCurrentVersion()) {
                 job = new AutoIngestJob(nodeData);
-                Path caseDirectory = PathUtils.findCaseDirectory(rootOutputDirectory, manifest.getCaseName());
```
esaunders: For some unknown reason this code *really* wanted to make sure the case directory path was set on the job, so much so that there are 3 identical chunks of code to accomplish it (in the body of the 'if', in the body of the 'else' and immediately after the 'if-else'). This change eliminates 2 of them. Given that you have also made a change to openCase() to update the node data after the directory is set again, do we even need to keep the 3rd version of the code here? If there isn't a really good reason I would vote to eliminate it to avoid future confusion.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -1595,54 +1587,53 @@ private void doRecoveryIfCrashed(Manifest manifest, AutoIngestJobNodeData jobNod
...
-                    /**
-                     * Use the manifest rather than the node data here to create
-                     * a new AutoIngestJob instance because the AutoIngestJob
-                     * constructor that takes a node data object expects the
-                     * node data to have fields that do not exist in earlier
-                     * versions.
-                     */
-                    job = new AutoIngestJob(manifest);
-                    job.setCaseDirectoryPath(caseDirectoryPath);
+            Path caseDirectoryPath = nodeData.getCaseDirectoryPath();
```
esaunders: The only potential pitfall here is that there is a completed job node that predates the existence of the case directory path in the node. I guess the worst that will happen is that the job will not appear in the completed list.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
```diff
@@ -2456,6 +2447,7 @@ private Case openCase() throws CoordinationServiceException, CaseManagementExcep
                             Thread.sleep(AutoIngestUserPreferences.getSecondsToSleepBetweenCases() * 1000);
                         }
                         currentJob.setCaseDirectoryPath(caseDirectoryPath);
+                        updateCoordinationServiceManifestNode(currentJob);  // update case directory path
```
esaunders: Other places that call updateCoordinationServiceManifestNode() get an exclusive lock on the manifest file node. This block of code has a lock on the case node.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
eugene7646: I really wanted to keep the number of code changes to a minimum so I decided not to touch the old ZK node support code. Perhaps we should delete that all together. But sure, I can modify that code as well.  I also didn't modify the doRecoveryIfCrashed() method which calls PathUtils.findCaseDirectory(), simply because that code gets executed very rarely so who cares that we will scan the output directory when doing a job recovery.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
eugene7646: Actually all existing completed jobs already have case directory path set. The code simply ignored it and tried to figure out the case directory path again, every single scan. This makes no sense. What's even worse, it's entirely possible for a PathUtils.findCaseDirectory() to find a directory that was not the case output directory for that completed job. The method simply finds "a case directory" that fits the naming schema. Given that we keep doing that again and again, somebody could delete the original case output folder and a new case output folder could get created that fits the naming scheme.

esaunders: Understood, but in theory there can be version 0 nodes for completed jobs that were created by an earlier version before the caseDirectoryPath member was added to AutoIngestNodeData.
You may not have them in your environment but that doesn't mean they can't or don't exist elsewhere.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestManager.java
eugene7646: Originally I thought so too but turns out that's not the case. For example, PeriodicJobStatusEventTask.run() calls updateCoordinationServiceManifestNode() without obtaining the lock. Same goes for processJob(), which actually calls the openCase() method (the one I modified). So I think obtaining the lock seems only necessary when creating a new ZK node, as a mechanism for actual node creation, or perhaps for the purposes of not creating duplicate ZK nodes?

esaunders: Okay, I give up trying to wrap my head around it then.

PR #4079, opened 2018-08-30
Add docs to building autopsy from source on linux debian

rishwanth1995:


eugene7646:
Develop branch contains proper Linux instructions at this point

PR #4094, opened 2018-09-13
Fix error resulting from merge issue of exception being added to method

wschaeferB:


Core/src/org/sleuthkit/autopsy/contentviewers/AnnotationsContentViewer.java
```diff
@@ -235,7 +236,7 @@ private void populateCentralRepositoryData(StringBuilder html, BlackboardArtifac
                 if (commentDataFound == false) {
                     addMessage(html, "There is no comment data for the selected content in the central repository.");
                 }
-            } catch (EamDbException | TskCoreException ex) {
+            } catch (EamDbException | TskCoreException | CorrelationAttributeNormalizationException ex) {
```
rcordovano: The existing error message does not match this new exception.

wschaeferB: I don't know that I made it any more helpful of a message but it will no longer have the possible inaccuracy. 

 I don't really understand what is going on here to cause the exception so this change is the best I can do in any sort of reasonable time frame.

PR #4098, opened 2018-09-14
4195 remove columns from results

dgrove727:


dgrove727:
The aforementioned line 404 with the bundle annotation is still used by the method, so I left it in.

PR #4102, opened 2018-09-17
4176 column level analysis

dannysmyda:


dannysmyda:
Don't take this unless the code using this as a dependency has been reviewed fully.

PR #4110, opened 2018-09-19
fix resolving bug and always go lower case. Progress update

bcarrier:


Core/src/org/sleuthkit/autopsy/modules/plaso/PlasoIngestModule.java
```diff
@@ -356,7 +363,8 @@ private AbstractFile getAbstractFile(String file) {
                 return abstractFiles.get(0);
             }
             for (AbstractFile resolvedFile : abstractFiles) {
-                if (filePath.matches(resolvedFile.getParentPath().toLowerCase())) {
+                // double check its an exact match
+                if (filePath.toLowerCase().matches(resolvedFile.getParentPath().toLowerCase())) {
```
millmanorama: why are we even using matches()?  there is no regex invlolved.  we could just use equalsIgnoreCase()

bcarrier: Done.

bcarrier:
Duplicated now by #4115

PR #4114, opened 2018-09-19
4223 Move common files enabling test off the EDT

APriestman:


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributeSearchAction.java
```diff
@@ -60,18 +63,7 @@ private CommonAttributeSearchAction() {
...
     public boolean isEnabled() {
-        boolean shouldBeEnabled = false;
-        try {
-            //dont refactor any of this to pull out common expressions - order of evaluation of each expression is significant
-            shouldBeEnabled
-                    = (Case.isCaseOpen()
-                    && Case.getCurrentCase().getDataSources().size() > 1)
-                    || CommonAttributePanel.isEamDbAvailableForIntercaseSearch();
-
-        } catch (TskCoreException ex) {
```
wschaeferB: I think we want to leave this condition for checking if the action should be enabled in.

I believe it ensures the action not available when the current case contains fewer than 2 data sources AND the central repository would not be available as a search option on the panel.

PR #4125, opened 2018-09-20
path groups get their datasource and not the one from the grouping co…

millmanorama:
…ntrols.


raman-bt:
I've reviewed and tested this PR and approve it for merge.

PR #4148, opened 2018-09-25
Added Index.isCompatible() method to determine whether a particular S…

esaunders:
…olr index is compatible with a given version number. Compatibility is based on the major version number.


eugene7646:
Eamonn is on vacation so I added a minor fix and created a new PR:
https://github.com/sleuthkit/autopsy/pull/4165

PR #4159, opened 2018-09-27
Cache IG data, change order of loading

bcarrier:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -766,6 +779,112 @@ public void insertFile(DrawableFile f, DrawableTransaction tr, CaseDbTransaction
...
+        }
+        
+        areCachesLoaded = true;
+    }
+    
+    /**
+     * Add a file to cache of files that have EXIF data
+     * @param objectID ObjId of file with EXIF
+     */
+    public void addExifCache(long objectID) {
```
millmanorama: shouldn't the addXXXCache methods be synchronized too?

bcarrier: Added syncronization in latest commit.

PR #4160, opened 2018-09-27
1033 backwards compatibility

millmanorama:
blow away old drawable.db if it doesn't have data_source_obj_id column


raman-bt:
@millmanorama: can you please resolve the conflicts in this?

PR #4171, opened 2018-10-01
4260 limiting cr comment

dannysmyda:


eugene7646:
Feature works, code looks good. The only comment i have is we should probably add the "@Override" annotation to the methods of CentralRepoCommentLengthFilter class. It helps explain why those methods have to be public. Otherwise it's ready to go.

PR #4174, opened 2018-10-03
1062 not on jfx thread

millmanorama:


Core/src/org/sleuthkit/autopsy/datamodel/DhsImageCategory.java
```diff
@@ -124,13 +124,25 @@ public String toString() {
...
             region.setPrefSize(16, 16);
             region.setBorder(new Border(new BorderStroke(getColor().darker(), BorderStrokeStyle.SOLID, CORNER_RADII_4, BORDER_WIDTHS_2)));
             Scene scene = new Scene(region, 16, 16, Color.TRANSPARENT);
-            snapshot = region.snapshot(null, null);
+            return region.snapshot(null, null);
+        } else {
+            //if we are not already on the JFX thread, do the generation on the javafx thread and block until done.
+            SettableFuture<Image> imageFuture = SettableFuture.create();
+            Platform.runLater(() -> generateSnapshot());
+            return Futures.getUnchecked(imageFuture); //block until snapshot is generated. It should be quick.
```
raman-bt: Quite likely I'm missing something here, but doesn't imageFuture need to be set somewhere for it to unblock from here?

millmanorama: gah, of course.  I had that, and then i was cleaning up and I must have missed it.

millmanorama:
@raman-bt  can you test this?

PR #4175, opened 2018-10-03
4273 inter case search provide disabled feedback

wschaeferB:


Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributePanel.java
```diff
@@ -670,10 +675,102 @@ private void intraCaseRadioActionPerformed(java.awt.event.ActionEvent evt) {//GE
...
-        search();
-        SwingUtilities.windowForComponent(this).dispose();
+        checkDataSourcesAndSearch();
+        this.dispose();
     }//GEN-LAST:event_searchButtonActionPerformed
 
+    /**
+     * If the settings reflect that a inter-case search is being performed,
+     * checks that the data sources in the current case have been processed with
+     * Correlation Engine enabled and exist in the central repository. Promting
```
APriestman: Typo - prompting

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributePanel.java
```diff
@@ -670,10 +675,102 @@ private void intraCaseRadioActionPerformed(java.awt.event.ActionEvent evt) {//GE
...
+        "CommonAttributePanel.incompleteResults.title=Search may be incomplete"
+    })
+    private void checkDataSourcesAndSearch() {
+        new SwingWorker<List<String>, Void>() {
+            //if the eamdb is enabled and an instance is able to be retrieved check if each data source has been processed into the cr 
+            @Override
+            protected List<String> doInBackground() throws Exception {
+                List<String> unCorrelatedDataSources = new ArrayList<>();
+                if (interCaseRadio.isSelected()) {
+                    if (EamDb.isEnabled() && EamDb.getInstance() != null) {
```
APriestman: You could reduce the nesting by checking the opposite of these two ifs and returning immediately, i.e., 
if (! radio.isSelected()) return xxx
if(! (that eam enabled thing) return xxx
(continue)

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributePanel.java
```diff
@@ -793,4 +890,9 @@ private void checkFileTypeCheckBoxState() {
         this.updateErrorTextAndSearchButton();
     }
 
+    private enum CorrelatedStatus {
```
APriestman: Add a quick comment block

Core/src/org/sleuthkit/autopsy/commonfilesearch/CommonAttributePanel.java
wschaeferB: Alright I generally try to avoid having multiple returns if I can, but I agree in this case it will improve readability quite a bit due to the depth of the nesting.

PR #4183, opened 2018-10-03
1069 - Placeholder for empty lists/selections

millmanorama:


raman-bt:
@millmanorama: is this ready or does this need to be merged to develop now?  Is there a JIRA number for it?

millmanorama:
There is jira 1069 which describes it some. I think it is good to go, but it is not important.

PR #4186, opened 2018-10-04
1054 make sure all db queries for widget initialization happen on bg threads.

millmanorama:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/utils/TaskUtils.java
```diff
@@ -42,8 +47,26 @@ protected T call() throws Exception {
         };
     }
 
-   public  static ListeningExecutorService getExecutorForClass(Class<?> clazz) {
+    public static ListeningExecutorService getExecutorForClass(Class<?> clazz) {
         return MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor(
                 new ThreadFactoryBuilder().setNameFormat("Image Gallery " + clazz.getSimpleName() + " BG Thread").build()));
     }
+
```
raman-bt: This is just a utils class but I think adding method headers here would be useful

bcarrier:
NPE fixed in #4198

PR #4199, opened 2018-10-09
Appropriate seven zip extractor memory fix

dannysmyda:


dannysmyda:
DEPENDS ON: https://github.com/sleuthkit/sleuthkit/pull/1338

Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/SevenZipExtractor.java
```diff
@@ -789,138 +787,100 @@ boolean unpack(AbstractFile archiveFile, ConcurrentHashMap<Long, Archive> depthM
...
-            }
-
-        }
-
-        public abstract long getSize();
+    private final static class MutableEncodedFileOutputStream extends EncodedFileOutputStream 
+            implements AutoCloseable {
+        
+        private static final int HEADER_LENGTH = 32;
+        private static final String HEADER = "TSK_CONTAINER_XOR1_xxxxxxxxxxxxx";
```
APriestman: While we can't change them at this point, having the XOR1 header and encoding method hard-coded here seems wrong. But they also don't seem like the kind of thing we want in the public API. I wonder if it would make more sense to create a public MutableEncodedFileOutputStream class over in the same package as EncodedFileOutputStream since the getHeader and encodeByte methods are package scope.

dannysmyda: @rcordovano your suggestion has now been implemented. I ran a test archive to verify that it is functioning correctly. Its now a candidate for final review by @APriestman and integration.

Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/SevenZipExtractor.java
```diff
@@ -795,140 +793,58 @@ boolean unpack(AbstractFile archiveFile, ConcurrentHashMap<Long, Archive> depthM
...
-
+    
     /**
-     * Stream used to unpack the archive to local file
+     * UnpackStream used by the SevenZipBindings to do archive extraction. A memory
+     * leak exists in the SevenZip library that will not let go of the Streams until
+     * the entire archive extraction is complete. So, to compensate this UnpackStream
+     * uses a MutableEncodedFileOutputStream, which supports setting a new disk 
+     * location for the contents to be written to (as opposed to creating a new steam
+     * for each file).
```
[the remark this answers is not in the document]
dannysmyda: Done.

PR #4205, opened 2018-10-11
Fix EFE handling of container exceptions

dannysmyda:
…ndling is fixed


eugene7646:
Code looks good to me. I will test this in the upcoming release candidate build.

@rcordovano FYI

dannysmyda:
@eugene7646 I don't think this is going in until 2.15, if that's what you mean by 'upcoming release'. There is now an autopsy release branch for 4.9.0.

eugene7646:
Ok I see. I was actually kind of hoping to have it in next release to lower the number of severe errors we see in logs.

PR #4218, opened 2018-10-15
4262 handle domain special characters

dgrove727:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractIE.java
```diff
@@ -571,24 +590,25 @@ private boolean executePasco(String indexFilePath, String outputFileName) {
...
+     * checks to detect invalid URLs.
      * 
-     * @param url The URL to test.
+     * @param url The URL from which to extract the domain.
      * 
-     * @return True if the URL should be ignored; otherwise false.
+     * @return The domain.
      */
-    private boolean isIgnoredUrl(String url) {
+    private String extractDomain(String url) {
```
APriestman: It doesn't look like any of the code calls this? The references are all to the Util version

dgrove727: Correction made.

PR #4229, opened 2018-10-23
1093 Timeline event mgr

millmanorama:


raman-bt:
@bcarrier, @rcordovano: I've built and sanity tested general Autopsy functionality and in particular Timeline, and didn't come across any blocking issues (except for the time its taking to run the Plaso module on XP  - I'll investigate that separately.)  I approve this PR, and the corresponding Sleuthkit PR.

PR #4231, opened 2018-10-23
Update Plaso Module to newer release and Change log2timeline run options

markmckinnon:
Update plaso from release 20180127 to 20180818.

Change run options in log2timeline to run with no dependency check

Add list of parsers to run so it will run all plugins except filestat


MariasStory:
Why is this commit not in dev?
It would be cool to have plaso in autopsy.

PR #4232, opened 2018-10-24
4303 timezone option

dgrove727:


Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
```diff
@@ -45,6 +46,7 @@
     public static final String HIDE_SLACK_FILES_IN_DATA_SRCS_TREE = "HideSlackFilesInDataSourcesTree"; //NON-NLS 
     public static final String HIDE_SLACK_FILES_IN_VIEWS_TREE = "HideSlackFilesInViewsTree"; //NON-NLS 
     public static final String DISPLAY_TIMES_IN_LOCAL_TIME = "DisplayTimesInLocalTime"; //NON-NLS
+    public static final String CUSTOM_TIME_ZONE = "CustomTimeZone"; //NON-NLS
```
rcordovano: @dgrove727 and @wschaeferB, does this need to be public? I realize that the precedent is mixed, but it looks to me like the previous practice of exposing of the names for name-value pairs in the underlying NbPreferences object was not necessary nor desirable. Am I missing something?

Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
```diff
@@ -181,6 +183,14 @@ public static boolean displayTimesInLocalTime() {
...
         preferences.putBoolean(DISPLAY_TIMES_IN_LOCAL_TIME, value);
     }
+    
+    public static String getCustomTimeZone() {
+        return preferences.get(CUSTOM_TIME_ZONE, TimeZone.GMT_ZONE.getID());
+    }
+    
+    public static void setCustomTimeZone(String timeZone) {
+        preferences.put(CUSTOM_TIME_ZONE, timeZone);
+    }
```
rcordovano: @dgrove727 and @wschaeferB, I am concerned that this name is not very descriptive. I find myself asking "Set custom time zone for what?" and  "What is a custom time zone, i.e., what makes it custom and if it is custom why is the default GMT?" Instead, it seems to me that it would be nice to try to make the implicit link between the DISPLAY_TIMES_IN_LOCAL_TIME and this setting more explicit. Perhaps something like set/getTimeZoneForDisplays. Then we would have code like

 if (!shouldDisplayTimesInLocalTime()) {
       return TimeZone.getTimeZone(UserPreferences.getTimeZoneForDisplays());
}

Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
dgrove727: This is used by DirectoryTreeTopComponent to refresh whenever the time zone option is changed.

PR #4256, opened 2018-11-02
4361 sco clean up and background tasks

dannysmyda:


dannysmyda:
Don't take until @wschaeferB changes to reducing number of EamDb.useCentralRepo() calls are in. He has made changes that I'd like to integrate into this PR.

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -457,4 +582,36 @@ protected static String getHashSetHitsCsvList(AbstractFile file) {
...
     }
+
+    /**
+     * Fill map with AbstractFile properties
+     *
+     * @param map     map with preserved ordering, where property names/values
+     *                are put
+     * @param content The content to get properties for.
+     */
+    @Deprecated
```
rcordovano: Since we are still using this method, please remove the deprecation and make a technical debt story about removing this method in the future. You can reference the story in the code in a comment using the form "TODO JIRA-XXXX:"

Core/src/org/sleuthkit/autopsy/datamodel/ToggleableNodeProperty.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Allows a property to be either enabled or disabled. When creating a sheet,
+     * this method is used to filter out from displaying in the UI.
+     * 
+     * @return boolean denoting the availiability of this property. True by default.
+     */
+    public boolean isEnabled() {
+        return true;
+    }
+}
```
rcordovano: I am not convinced that the benefits of this class are worth the complexity it introduces. It is certainly a creative bit of code, but it seems that the use case is limited to the creation of instances of anonymous inner classes that override the enabled method as arguments to the updateProperty(Sheet) method, and more or less boils down to avoiding some if statements. Let's remove it, please.

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -174,11 +268,78 @@ private void removeListeners() {
...
+     * disrupt their UI.
+     *
+     * Race condition if not synchronized. Only one update should be applied at
+     * a time. The timing of currSheetSet.getProperties() could result in
+     * wrong/stale data being shown!
+     *
+     * @param newProps New file property instances to be updated in the current
+     *                 sheet.
+     */
+    private synchronized void updateProperty(ToggleableNodeProperty... newProps) {
```
rcordovano: I think that this method is better named as updateSheet or updatePropertySheet. It calls setSheet.

I could also go with updateProperties.

Core/src/org/sleuthkit/autopsy/datamodel/SCOAndTranslationTask.java
```diff
@@ -0,0 +1,144 @@
...
+                results));
+
+            //Once we've got the SCO columns, then lets fire the translation result.
+            //Updating of this column is significantly lower priority than 
+            //getting results to the SCO columns!
+            listener.propertyChange(new PropertyChangeEvent(
+                AutopsyEvent.SourceType.LOCAL.toString(),
+                AbstractAbstractFileNode.NodeSpecificEvents.TRANSLATION_AVAILABLE.toString(),
+                null,
+                FileNodeUtil.getTranslatedFileName(content)));
```
rcordovano: Is it possible that we are dispatching a task to the EDT that will often do nothing because there is no translation provider or the translation has already been done? I don't know that it really merits much attention to optimization, but we could make the firing of the property change event conditional on the result of getTranslatedFileName.

dannysmyda:
Don't take this yet, I messed up the XML format while doing merge conflicts. I'm working on fixing it

PR #4265, opened 2018-11-07
make EventAddedEvent 'reconstitutable'' after transmision over the ne…

millmanorama:
…twork.


Core/src/org/sleuthkit/autopsy/timeline/events/EventAddedEvent.java
```diff
@@ -18,15 +18,62 @@
...
+
+    private transient TimelineEvent addedEvent;
+    private final long addedEventID;
 
     public EventAddedEvent(org.sleuthkit.datamodel.TimelineManager.EventAddedEvent event) {
-        super(Case.Events.EVENT_ADDED.name(), null, event.getEvent());
+        super(Case.Events.EVENT_ADDED.name(), null, event.getAddedEvent().getEventID());
+        addedEvent = event.getAddedEvent();
+        addedEventID = addedEvent.getEventID();
+    }
```
rcordovano: It seems like passing the event id to the super class constructor and also making it a field might be a little confusing. Personally, I prefer the approach used in DataSourceAddedEvent where the id of the object is obtained in getNewValue via long id = (Long) super.getNewValue() and the "extra" field is not required. Thoughts?

PR #4269, opened 2018-11-08
4390 text extraction stream impl

dannysmyda:


Core/src/org/sleuthkit/autopsy/coreutils/SQLiteTableReader.java
```diff
@@ -0,0 +1,560 @@
...
+     * Last ditch effort to close the connections during garbage collection.
+     *
+     * @throws Throwable
+     */
+    @Override
+    protected void finalize() throws Throwable {
+        try {
+            close();
+        } catch (SQLiteTableReaderException ex) {
+            //Do nothing, we tried out best to close the connection.
```
rcordovano: It is fine to not propagate the exception here, but I would prefer that we log it.

Core/src/org/sleuthkit/autopsy/coreutils/SQLiteTableReader.java
```diff
@@ -0,0 +1,560 @@
...
+            if (Objects.nonNull(statement)) {
+                statement.close();
+            }
+            if (Objects.nonNull(queryResults)) {
+                queryResults.close();
+            }
+            liveResultSet = false;
+        } catch (SQLException ex) {
+            //Do nothing, can't close.. tried our best.
+        }
```
rcordovano: It is fine to not propagate the exception here, but I would prefer that we log it.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SqliteTextExtractor.java
```diff
@@ -91,267 +84,251 @@ boolean isSupported(Content file, String detectedFormat) {
...
             try {
-                connection.close();
-            } catch (SQLException ex) {
-                //Non-essential exception, user has no need for the connection 
-                //object at this stage so closing details are not important
-                logger.log(Level.WARNING, "Could not close JDBC connection", ex);
+                reader.close();
+            } catch (SQLiteTableReaderException ex) {
+                //Done reading, but couldn't close the resources. Nothing we can 
+                //do as a recovery.
```
rcordovano: It is fine to not propagate the exception here, but I would prefer that we log it.

Core/src/org/sleuthkit/autopsy/coreutils/SQLiteTableReader.java
```diff
@@ -0,0 +1,560 @@
...
+         */
+        public SQLiteTableReader build() {
+            return new SQLiteTableReader(this);
+        }
+    }
+
+    private final AbstractFile file;
+    private final Builder builder;
+    
+    private final String SELECT_ALL_QUERY = "SELECT * FROM \"%s\"";
```
rcordovano: This can be static.

PR #4273, opened 2018-11-08
4354 add object id to cr

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/SqliteEamDb.java
```diff
@@ -1156,4 +1170,25 @@ private void releaseSharedLock() {
         rwLock.readLock().unlock();
     }
 
+    @Override
+    boolean doesColumnExist(Connection conn, String tableName, String columnName) throws SQLException {
+        final String tableInfoQueryTemplate = "PRAGMA table_info(%s)";  //NON-NLS
+        ResultSet resultSet;
+        Statement statement = conn.createStatement();
+        boolean columnExists = false;
+        resultSet = statement.executeQuery(String.format(tableInfoQueryTemplate, tableName));
```
APriestman: I think this should be in the try-catch block to make sure everything gets closed

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresEamDb.java
```diff
@@ -189,36 +191,57 @@ protected Connection connect() throws EamDbException {
...
         }
     }
 
+    @Override
+    boolean doesColumnExist(Connection conn, String tableName, String columnName) throws SQLException {
+        final String objectIdColumnExistsTemplate = "SELECT EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name='%s' AND column_name='%s')";  //NON-NLS
+        ResultSet resultSet;
+        Statement statement = conn.createStatement();
+        boolean columnExists = false;
+        resultSet = statement.executeQuery(String.format(objectIdColumnExistsTemplate, tableName, columnName));
```
APriestman: Same as the other place - this should probably be in the try-catch block to close everything.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/AbstractSqlEamDb.java
```diff
@@ -808,9 +808,9 @@ public void addArtifactInstance(CorrelationAttributeInstance eamArtifact) throws
         String sql
                 = "INSERT INTO "
                 + tableName
-                + "(case_id, data_source_id, value, file_path, known_status, comment) "
+                + "(case_id, data_source_id, value, file_path, known_status, comment, object_id) "
```
rcordovano: Let's use src_file_obj_id fr the column name.

rcordovano: Actually, per Brian Carrier, let's use file_obj_id.

PR #4274, opened 2018-11-09
1889 Add CASE/UCO report generation 

eugene7646:


Core/src/org/sleuthkit/autopsy/guiutils/DataSourceComboBoxModel.java
```diff
@@ -17,7 +17,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.sleuthkit.autopsy.commonfilesearch;
+package org.sleuthkit.autopsy.guiutils;
```
wschaeferB: Just leaving a note that this move from an non-exported package to an exported package will represent an addition to the public API for this and the other File being moved, it certainly makes more sense with organization though since the new use of this data source list box is not related to common file search.

Core/src/org/sleuthkit/autopsy/modules/case_uco/ReportCaseUco.java
```diff
@@ -0,0 +1,458 @@
+ /*
+ *
+ * Autopsy Forensic Browser
+ * 
+ * Copyright 2012-2018 Basis Technology Corp.
```
wschaeferB: Copy paste of 2012- into new file

Core/src/org/sleuthkit/autopsy/modules/case_uco/ReportCaseUco.java
```diff
@@ -0,0 +1,458 @@
...
+ * 
+ * Copyright 2012 42six Solutions.
+ * Contact: aebadirad <at> 42six <dot> com
+ * Project Contact/Architect: carrier <at> sleuthkit <dot> org
```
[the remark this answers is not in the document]
eugene7646: Hmmm, interesting! The code was initially copy/pasted from ReportBodyFile report module, which contains this copyright info for some reason. Further more, @rcordovano when I search Autopsy for "42six" - 17 different Autopsy Java files come up, all having their copyright info in our header!!! Who the hell are these guys? Google says it's some software company from Columbia, MD... I removed the copyright info from the CASE/UCO class. We should probably do the same for the other files.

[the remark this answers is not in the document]
wschaeferB: My understanding after asking Brian is that 42six reflects the original team that was him and another person (persons?) and exists in some of the oldest code to attribute the work in those files to the dev team as it existed prior to Basis.

Core/src/org/sleuthkit/autopsy/guiutils/DataSourceComboBoxModel.java
rcordovano: Yes, I did propose this change. But thanks for being alert about public API changes, @wschaeferB!

PR #4275, opened 2018-11-09
fix what look likes accidental fall though with new wifi artifact type

millmanorama:


millmanorama:
this is the same fix as https://github.com/sleuthkit/autopsy/pull/4276

PR #4284, opened 2018-11-17
4410 html report header footer

dgrove727:


dgrove727:
Of course if breaks compatibility. I should have applied more scrutiny to the instruction I was given, because I'm certain Richard never intended me to break compatibility. I'll discuss this further in JIRA and try and come to some sort of consensus on how we will do this.

Thanks, Johnathan.

Core/src/org/sleuthkit/autopsy/report/GeneralReportModule.java
```diff
@@ -33,13 +31,4 @@
...
     public void generateReport(String baseReportDir, ReportProgressPanel progressPanel);
 
-    /**
-     * Returns the configuration panel for the report, which is displayed in the
-     * report configuration step of the report wizard.
-     *
-     * @return Configuration panel or null if the module does not need
-     *         configuration.
-     */
-    public JPanel getConfigurationPanel();
```
wschaeferB: I am noting that this is a public API change removing this method as the report package is exported.

Core/src/org/sleuthkit/autopsy/report/ReportHTMLConfigurationPanel.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+package org.sleuthkit.autopsy.report;
+
+import org.sleuthkit.autopsy.coreutils.ModuleSettings;
+
+/**
+ * The panel shown for all TableReportModules when configuring report modules.
+ */
+@SuppressWarnings("PMD.SingularField") // UI widgets cause lots of false positives
+public class ReportHTMLConfigurationPanel extends javax.swing.JPanel {
```
wschaeferB: This class and its methods can be package scope I believe as they do not appear to be accessed outside of the report package.

Core/src/org/sleuthkit/autopsy/report/ReportHTML.java
```diff
@@ -430,6 +452,40 @@ public void endDataType() {
...
+     * Write HTML-formatted page header text based on the text provided in the
+     * configuration panel.
+     * 
+     * @return The HTML-formatted text.
+     */
+    private String writePageHeader() {
+        StringBuilder output = new StringBuilder();
+        String pageHeader = configPanel.getHeader();
+        if (pageHeader.isEmpty() == false) {
+            output.append("<p style=\"text-align: center\">")
```
wschaeferB: So I would kinda like to see the size of this text a little bigger it sounded like Dimitry thought it should be about the same size as the section heading, and if we are going to do that I think we should define the size and style stuff in the CSS we write around lines 1130 for the summary and lines 915 for the sections.

Something like 
"#headerFooter {width:100%; padding: 10px; line-height: 25px; text-align: center; font-size: 20px;}\n"

So that in the write page header and write page footer code we would be appending "<div id=\"headerFooter\">"  before the text and "</div>\n" after the text.

PR #4293, opened 2018-11-21
update to match tsk API

millmanorama:


millmanorama:
this PR is redundant now

PR #4297, opened 2018-11-21
1133: cache the group seen status.

raman-bt:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -801,6 +802,16 @@ public void process(ResultSet resultSet) {
      */
     public void markGroupSeen(GroupKey<?> groupKey, boolean seen, long examinerID) throws TskCoreException {
 
+        /*
+         * Check the groupSeenCache to see if the seen status for this group was set recently.
+         * If recently set to the same value, there's no need to update it
+         */
+        String cacheKey = groupKey.getDataSourceObjId() + "_" + groupKey.getValueDisplayName() + "_" + groupKey.getAttribute().getDisplayName();
+        Boolean cachedValue = groupSeenCache.getIfPresent(cacheKey);
```
millmanorama: is there a reason we don't just use the groupKey as the cacheKey directly?

PR #4315, opened 2018-11-29
4395 change E01 module to also calculate missing hashes

APriestman:


Core/src/org/sleuthkit/autopsy/modules/dataSourceIntegrity/Bundle.properties
```diff
@@ -1,15 +1,19 @@
 OpenIDE-Module-Name=ewfVerify
-EwfVerifyIngestModule.moduleName.text=E01 Verifier
-EwfVerifyIngestModule.moduleDesc.text=Validates the integrity of E01 files.
+EwfVerifyIngestModule.moduleName.text=Data Source Integrity
```
dgrove727: I would have thought these would get renamed the same way the module name and class file name got renamed. It might provide for better clarity as to what one is looking at in case they're not familiar with the history of the module.

Core/src/org/sleuthkit/autopsy/modules/dataSourceIntegrity/DataSourceIntegrityIngestSettingsPanel.java
```diff
@@ -0,0 +1,120 @@
...
+import org.sleuthkit.autopsy.ingest.IngestModuleIngestJobSettingsPanel;
+
+/**
+ * Ingest job settings panel for the Correlation Engine module.
+ */
+@SuppressWarnings("PMD.SingularField") // UI widgets cause lots of false positives
+final class DataSourceIntegrityIngestSettingsPanel extends IngestModuleIngestJobSettingsPanel {
+
+    /**
+     * Creates new form IngestSettingsPanel
```
dgrove727: Since you renamed this, the comment should probably reflect the new name.

Core/src/org/sleuthkit/autopsy/modules/dataSourceIntegrity/DataSourceIntegrityModuleFactory.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2014 Basis Technology Corp.
```
dgrove727: Should probably update the copyright.

PR #4329, opened 2018-12-04
4438 empty string exif attributes

dgrove727:


Core/src/org/sleuthkit/autopsy/modules/exif/ExifParserFileIngestModule.java
```diff
@@ -196,13 +195,19 @@ ProcessResult processFile(AbstractFile f) {
             ExifIFD0Directory devDir = metadata.getFirstDirectoryOfType(ExifIFD0Directory.class);
             if (devDir != null) {
                 String model = devDir.getString(ExifIFD0Directory.TAG_MODEL);
-                if (model != null && !model.isEmpty()) {
-                    attributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DEVICE_MODEL, ExifParserModuleFactory.getModuleName(), model));
+                if (model != null) {
```
millmanorama: this could also be done with (apache) StringUtils.isNotBlank()

PR #4330, opened 2018-12-04
4425 - Move TextExtractors out of KWS and into Core.

dannysmyda:


Core/src/org/sleuthkit/autopsy/textextractors/ContentTextExtractor.java
```diff
@@ -73,36 +75,72 @@
...
 
+    /**
+     * Accepts context instance to allow for extractor configuration on specific 
+     * file types. 
+     * 
+     * See extractionconfigs package for available file type configurations.
+     * 
+     * @param context Instance that contains config classes 
+     */
+    public ContentTextExtractor(ExtractionContext context) {
```
esaunders: Nothing being done with context

Core/src/org/sleuthkit/autopsy/textextractors/ExtractionContext.java
```diff
@@ -0,0 +1,72 @@
...
+public class ExtractionContext {
+
+    ClassToInstanceMap<Object> extractionConfigs;
+
+    public ExtractionContext() {
+        extractionConfigs = MutableClassToInstanceMap.create();
+    }
+
+    /**
+     * Internally stores a class-instance pair.
```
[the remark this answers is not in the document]
dannysmyda: No, these settings are able to be set at run time. The existence of the provider does not matter.

Core/src/org/sleuthkit/autopsy/textextractors/SqliteTextExtractor.java
```diff
@@ -32,26 +32,62 @@
...
 
+    /**
+     * Determines if this extractor is fit to run.
+     * 
+     * @return Flag indicating if it should or shouldn't be run.
+     */
     @Override
     public boolean isDisabled() {
-        return false;
+        try {
```
esaunders: Is it possible for the SQLite JDBC class to be unavailable when this method is first called and subsequently be available at some point later when this method is called again?
If not I'd suggest only attempting to load the class once (and reduce the number of times the message appears in the logs). Does this really call for a SEVERE log message?

rcordovano:
@dannysmyda, this PR no longer auto-merges. Please add a merge conflict resolution commit.

Core/src/org/sleuthkit/autopsy/textextractors/SqliteTextExtractor.java
dannysmyda: I agree it shouldn't be done every time that the method is called. My next commit will have this fixed.

PR #4345, opened 2018-12-08
4396 User can enter image data source hashes

rcordovano:


Core/src/org/sleuthkit/autopsy/casemodule/AddImageTask.java
```diff
@@ -233,6 +241,48 @@ private void commitOrRevertAddImageProcess(Case currentCase, List<String> errorM
...
                         }
                         newDataSources.add(newImage);
+                        if (!StringUtils.isBlank(md5)) {
+                            try {
+                                newImage.setMD5(md5);
+                            } catch (TskCoreException | TskDataException ex) {
+                                /*
+                                 * Treat both exceptions as the same since this
+                                 * is a new image and the hash should not
+                                 * already be set.
```
APriestman: Is that true? I thought the MD5 was set by the ewf code in TSK when opening the image.

rcordovano: Good catch! I was not thinking about the possibility that a user would enter hashes in the panel for an EWF data source. I will fix this.

PR #4368, opened 2018-12-13
4519 - Make tesseract.exe spawns for images cancellable

dannysmyda:


dannysmyda:
I'm kicking off a regression test to be sure.

Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -110,7 +120,7 @@
 
     private static final java.util.logging.Logger tikaLogger = java.util.logging.Logger.getLogger("Tika"); //NON-NLS
 
-    private final ExecutorService tikaParseExecutor = Executors.newSingleThreadExecutor();
+    private final ExecutorService executorService = Executors.newSingleThreadExecutor();
```
rcordovano: We should pass a ThreadFactory to this executor to give the threads meaningful names.

PR #4375, opened 2018-12-14
1130: determine when a path group is analyzed.

raman-bt:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/grouping/GroupManager.java
```diff
@@ -238,7 +243,8 @@ synchronized public void reset() {
         setGroupBy(DrawableAttribute.PATH);
         setSortOrder(SortOrder.ASCENDING);
         setDataSource(null);
-
+        resetLastUpdatedPathGroup();
```
rcordovano: @bcarrier and @raman-bt, I don't think it is correct to call this here. The case is closed and the controller is shutting down. This should just set lastUpdatedPathGroup to null.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/DrawableDB.java
```diff
@@ -1285,29 +1305,50 @@ public Boolean areFilesAnalyzed(Collection<Long> fileIds) throws SQLException {
...
+     * 
+     * @param groupKey group key.
+     * @return true if the group is analyzed.
+     * @throws SQLException
+     * @throws TskCoreException 
+     */
+    public Boolean isGroupAnalyzed(GroupKey<?> groupKey) throws SQLException, TskCoreException {
+        
+        // Callback to process result of isAnalyzed query
+        class IsGroupAnalyzedQueryResultProcessor extends CompletableFuture<Boolean> implements CaseDbAccessQueryCallback {
```
rcordovano: This is actually not needed.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/ImageGalleryModule.java
```diff
@@ -378,6 +378,7 @@ public void propertyChange(PropertyChangeEvent event) {
                              * database.
                              */
                             if (controller.isListeningEnabled()) {
+                                controller.getGroupManager().resetLastUpdatedPathGroup();
```
rcordovano: @bcarrier and @raman-bt, the need for this call concerns me a bit. As I understand it, we need some way to mop up after the last path is completed. That means that this inner event handler class in the ImageGalleryModule class is tightly coupled to implementation details inside of the GroupManager. At the very least we need to comment this. But I have a growing conviction that instead of having event handlers in ImageGalleryModule that try to manipulate the rest of the system, the interested components should have their own event handlers. I think the code would be simpler and sounder then.

raman-bt:
Closing this PR. First need to address the hang, allow for rebuild task to update the isAnalyzed state of a group, and address Richard's comments.

PR #4405, opened 2018-12-19
Added section on OCR language packs

APriestman:


docs/doxygen-user/keyword_search.dox
```diff
@@ -52,6 +52,20 @@ The "Indexed Text" tab shows the results when running the keyword search module
...
 
+By default, OCR is only configured for English text. Its configuration depends on the presence of language files (called "traineddata" files) 
+that exist in a location that Autopsy can understand. To add support for more languages, you will need to download additional "traineddata" 
+and move them to the right location. The following steps breakdown this process for you:
+
+<ol>
+<li>Navigate to https://github.com/tesseract-ocr/tesseract/wiki/Data-Files.
+<li>Under the section titled "Data Files for Version 4.00 (November 29, 2016)" you will find a table containing files that represent each language. These files have the extension ".traineddata".
+<li>To download the desired language, click on the links in the far right-hand column of the table. You may download as many as you like. Note that you must only choose from this table. Language files under any other sections are not guaranteed to work in Autopsy.
+<li>Once you've downloaded your language files, simply drag and drop them into the "\autopsy\Tesseract-OCR\tessdata" folder of your Autopsy install location (typically in Program Files).
```
rcordovano: @APriestman and @dannysmyda, I this is the wrong location. We should be putting things like this in the user's AppData at the top-level as discussed previously, i.e., ~\AppData\Roaming\autopsy. PlatformUtil.getUserDirectory should get to the right place in the code.

rcordovano: The good thing about AppData is that it persists when new versions of Autopsy are installed.

PR #4425, opened 2018-12-26
1130 - is group analyzed

raman-bt:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/grouping/GroupManager.java
```diff
@@ -107,6 +107,11 @@
 
     private final ImageGalleryController controller;
 
+    /**
+     * Keeps track of the last path group
+     *  - a change in path indicates the last path group is analyzed
+     */
+    private GroupKey<?> lastUpdatedPathGroup = null;
```
rcordovano: Is it possible that this member variable may be accessed by multiple threads?

raman-bt: Ugh, yes.  Thanks for catching this, I'll fix it.

rcordovano: I see that there is now an annotation indicating that the intrinsic lock (monitor) of the GroupManager is used to control access to lastUpdatedPathGroup. However, the access in checkForPathGroupChange is only guarded indirectly by having the sole call to it happening from a synchronized method. It would be safer and more clear to synchronize checkForPathGroupChange as well. Also, there is no guarding at all in resetLastUpdatedPathGroup, which is not called from a synchronized method.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/datamodel/grouping/GroupManager.java
```diff
@@ -618,14 +624,65 @@ synchronized public void handleFileUpdate(Collection<Long> updatedFileIDs) {
...
+     * we have moved from one folder to the next, and the group for the 
+     * previous PATH can be considered as analyzed and can be displayed.
+     * 
+     * NOTE: this a close approximation for when all files in a folder have been processed, 
+     * but there's some room for error - files may go down the ingest pipleline  
+     * out of order or the events may not always arrive in the same order
+     * 
+     * @param groupKey 
+     */
+    private void checkForPathGroupChange(GroupKey<?> groupKey) {
```
rcordovano: In my opinion, the way this method is documented and named, the changing of the filed tracking the last updated path group is a side effect - everything is about checking but not abut changing, e.g. "Checks if the given path is different from the last updated path group."  I suggest that we rename this method to something like updateLastUpdatedPathGroup. Or perhaps the field lastUpdatedPathGroup is actually something more like currentPathGroup and the method could be updateCurrentPathGroup? Does that express the concept better?

raman-bt:
Closing this PR. The changes in this PR haven been merged to another branch, and pulled in with the following PR:

[PR 4446](https://github.com/sleuthkit/autopsy/pull/4446)

PR #4455, opened 2019-01-10
4622 dedupe os accounts

wschaeferB:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractIE.java
```diff
@@ -22,6 +22,7 @@
  */
 package org.sleuthkit.autopsy.recentactivity;
 
+import com.sun.org.apache.bcel.internal.generic.L2D;
```
APriestman: Might as well remove this

PR #4467, opened 2019-01-16
4632 detect android os files

wschaeferB:
Builds on 4630 in PR #4466 as well as requiring sleuthkit/sleuthkit#1413


wschaeferB:
PR closed while it is re-worked to meet desired structure

PR #4470, opened 2019-01-18
4632 4630 4629 detect os specific files

wschaeferB:
New PR for 4629, 4630, and 4632 taking into account the desired split approach.

Still builds on #4456 and requires https://github.com/sleuthkit/sleuthkit/pull/1413 to build


wschaeferB:
replaced by PR #4470

PR #4478, opened 2019-01-23
4656 - Add OCR language pack configuration

dannysmyda:


Core/src/org/sleuthkit/autopsy/coreutils/PlatformUtil.java
```diff
@@ -126,6 +128,26 @@ public static String getUserPythonModulesPath() {
     public static String getOcrLanguagePacksPath() {
         return getUserDirectory().getAbsolutePath() + File.separator + OCR_LANGUAGE_SUBDIRECTORY;
     }
+    
+    /**
+     * Get the names of the language packs installed at the user directory.
+     * 
+     * @return List of language packs base names
+     */
+    public static List<String> getOcrLanguagePacks() {
```
rcordovano: @dannysmyda and @wschaeferB, this does not strike me as something that needs to be added to exported public API. It's only current client is TikaTextExtractor. Why not put it there?

rcordovano: Never mind.

dannysmyda: It's used elsewhere to populate the dropdown menu.

PR #4482, opened 2019-01-24
Now using single-frame thumbnails for videos.

dgrove727:


Core/src/org/sleuthkit/autopsy/coreutils/VideoUtils.java
```diff
@@ -148,48 +146,42 @@ static BufferedImage generateVideoThumbnail(AbstractFile file, int iconSize) {
                 LOGGER.log(Level.WARNING, "Error getting fps or total frames for {0}", ImageUtils.getContentPathSafe(file)); //NON-NLS
                 return null;
             }
-            double milliseconds = 1000 * (totalFrames / fps); //total milliseconds
-
-            double timestamp = Math.min(milliseconds, 500); //default time to check for is 500ms, unless the files is extremely small
+            double duration = 1000 * (totalFrames / fps); //total milliseconds
 
-            int framkeskip = Double.valueOf(Math.floor((milliseconds - timestamp) / (THUMB_COLUMNS * THUMB_ROWS))).intValue();
+            int timestamp = (int) Math.min(duration * 0.9, 5000); //default time to check for is 5 seconds, unless the files is extremely small
```
esaunders: What's the significance of the 0.9 magic number here?
Also, might want to more the Thread.interrupted() block up to take care of the Codacy issue.

dgrove727: I have updated this to document the rationale behind the 0.9, as well as making it a class-level constant. I also moved the 'interrupted()' call up. Hopefully Codacy will be fine with it now.

Core/src/org/sleuthkit/autopsy/coreutils/VideoUtils.java
```diff
@@ -148,48 +160,77 @@ static BufferedImage generateVideoThumbnail(AbstractFile file, int iconSize) {
...
-                    if (imageMatrix.channels() == 3) {
-                        for (int k = 0; k < data.length; k += 3) {
-                            byte temp = data[k];
-                            data[k] = data[k + 2];
-                            data[k + 2] = temp;
-                        }
-                    }
-
-                    bufferedImage.getRaster().setDataElements(imageMatrix.cols() * x, imageMatrix.rows() * y, imageMatrix.cols(), imageMatrix.rows(), data);
+            // Convert the matrix that contains the frame to a buffered image.
```
esaunders: You might consider checking the height and width of the imageMatrix here. Otherwise we end up with log messages and stack traces related to bad thumbnail dimensions.

dgrove727: I've added a call to 'empty()'. If true, we simply return null.

PR #4488, opened 2019-01-28
4680 Suppress inserts in image_galley_groups_seen table

dannysmyda:


dannysmyda:
@rcordovano Testing this might be more trouble than it is worth.

PR #4490, opened 2019-01-28
4632 4630 4629 detect os specific files

wschaeferB:
re-places PR #4470


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractOs.java
```diff
@@ -0,0 +1,264 @@
...
+        LINUX_GENTOO(Bundle.ExtractOs_gentooLinuxOs_label(), Bundle.ExtractOs_gentooLinuxVolume_label(), Arrays.asList(LINUX_GENTOO_PATH)),
+        LINUX_UNITED_LINUX(Bundle.ExtractOs_unitedLinuxOs_label(), Bundle.ExtractOs_unitedLinuxVolume_label(), Arrays.asList(LINUX_UNITED_LINUX_PATH)),
+        LINUX_UBUNTU(Bundle.ExtractOs_ubuntuLinuxOs_label(), Bundle.ExtractOs_ubuntuLinuxVolume_label(), Arrays.asList(LINUX_UBUNTU_PATH));
+
+        private final String osInfoLabel;
+        private final String dsUsageLabel;
+        private final List<String> filePaths;
+
+        /**
+         * An constructs a value for an OS_TYPE enum
```
APriestman: typo

Core/src/org/sleuthkit/autopsy/casemodule/datasourceSummary/DataSourceSummaryPanel.java
```diff
@@ -118,11 +119,11 @@ private String getOSName(DataSource selectedDataSource) {
...
                     //get the datasource id from the OSInfo's first artifact if it has artifacts
                     if (!osInfo.getArtifacts().isEmpty() && osInfo.getArtifacts().get(0).getDataSource().getId() == selectedDataSource.getId()) {
-                        osName = osInfo.getOSName();
-                        //if this OSInfo object has a name use it otherwise keep checking OSInfo objects
                         if (!osName.isEmpty()) {
-                            break;
+                            osName += ", ";
                         }
+                        osName += osInfo.getOSName();
+                        //if this OSInfo object has a name use it otherwise keep checking OSInfo objects
```
APriestman: Maybe remove this old comment and add something up a few lines like "If this OSInfo object has a name add it to the list of OS names for this data source" (assuming that's accurate)

wschaeferB: I will remove the comment for now, this is adjusted again in the stories I am currently working on so that it no longer uses the OSInfo class or OSUtility, so a revised comment would be extremely short lived.

wschaeferB:
Latest changes to consolidate gson versions appears to of had side effect of causing a additional version of sqlite-jdbc to appear

PR #4495, opened 2019-01-29
ensure path variable in diff script is always defined before use

wschaeferB:


wschaeferB:
Replaced by: https://github.com/sleuthkit/autopsy/pull/4529

PR #4506, opened 2019-01-31
4692 fix hxd editor bugs

dannysmyda:


Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerHex.java
```diff
@@ -359,36 +361,71 @@ private void goToOffsetTextFieldActionPerformed(java.awt.event.ActionEvent evt)
     }//GEN-LAST:event_goToOffsetTextFieldActionPerformed
 
     @NbBundle.Messages({"DataContentViewerHex.launchError=Unable to launch HxD Editor. "
-                        + "Please set-up the HdX install location in Tools -> Options -> External Viewer"})
+                        + "Please set-up the HxD install location in Tools -> Options -> External Viewer",
```
APriestman: I don't think the dash should be there

dannysmyda:
@rcordovano This is good to go now.

PR #4510, opened 2019-02-01
Optimization and Bug fix - Make Tika use AbstractFile mimetype instead of recomputing

dannysmyda:
This is esp critical if we override the mimetype ourselves. We also gain a noticeable performance boost.


Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -145,6 +147,19 @@
 
     public TikaTextExtractor(Content content) {
         this.content = content;
+        if(content instanceof AbstractFile) {
+            //Override the detector in Tika and use the already computed
+            //mimetype. This also saves on unneccessary file reads.
+            parser = new AutoDetectParser(new Detector() {
+                @Override
+                public MediaType detect(InputStream in, Metadata mtdt) throws IOException {
+                    return MediaType.parse(AbstractFile.class.cast(content).getMIMEType());
```
esaunders: What if the mime type hasn't been calculated yet?

PR #4511, opened 2019-02-01
4367 - upgrade to tika 1.20

dannysmyda:


dannysmyda:
We need Travis to be able to perform a full build before we can mark this as complete. Right now they are both hung up on another issue.

PR #4513, opened 2019-02-01
4696 - custom artifact handling

dannysmyda:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -65,101 +90,48 @@ public static String getEmailAddressAttrString() {
...
+            return eamArtifacts;
+        }
+        
         try {
-            BlackboardArtifact artifactForInstance = null;
-            if (BlackboardArtifact.ARTIFACT_TYPE.TSK_INTERESTING_ARTIFACT_HIT.getTypeID() == artifact.getArtifactTypeID()) {
-                // Get the associated artifactForInstance
-                BlackboardAttribute attribute = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ASSOCIATED_ARTIFACT));
-                if (attribute != null) {
-                    artifactForInstance = Case.getCurrentCaseThrows().getSleuthkitCase().getBlackboardArtifact(attribute.getValueLong());
```
APriestman: Did we decide that we don't need to process interesting item artifacts since the underlying artifact should have also come this code?

dannysmyda: I think so, though @bcarrier should be able to confirm.

PR #4522, opened 2019-02-05
4484 restructure tools menu

dannysmyda:


dannysmyda:
@rcordovano This required Bundle changes that also exist in the Japanese files, what is the policy for changes to this file?

Core/src/org/sleuthkit/autopsy/commonpropertiessearch/CommonAttributeSearchAction.java
```diff
@@ -151,7 +151,7 @@ protected void done() {
     }
 
     @NbBundle.Messages({
-        "CommonAttributeSearchAction.getName.text=Find Common Properties"})
+        "CommonAttributeSearchAction.getName.text=Find Related Cases"})
```
wschaeferB: I think we want the dialog over at line 94 of CommonAttributePanel.java to have the same name as the action which opens it.

PR #4527, opened 2019-02-07
1184 chrome cache

raman-bt:
This PR is for Review/Test ONLY
(DO NOT merge till we've finalized the new artifact name)

Needs this [PR ](https://github.com/sleuthkit/sleuthkit/pull/1427) for testing.

Can be tested using the XP image, which has Chrome cache files.

### DIFF
--- a/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
+++ b/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
@@ -27,6 +27,7 @@ import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -46,7 +47,6 @@ import org.sleuthkit.autopsy.ingest.IngestModule.IngestModuleException;
 import org.sleuthkit.autopsy.ingest.IngestServices;
 import org.sleuthkit.autopsy.ingest.ModuleContentEvent;
 import org.sleuthkit.autopsy.ingest.ModuleDataEvent;
-import org.sleuthkit.autopsy.modules.filetypeid.FileTypeDetector;
 import org.sleuthkit.datamodel.AbstractFile;
 import org.sleuthkit.datamodel.BlackboardArtifact;
 import org.sleuthkit.datamodel.BlackboardArtifact.ARTIFACT_TYPE;
@@ -69,7 +69,7 @@ final class ChromeCacheExtractor {
     private final static int INDEXFILE_HDR_SIZE = 92*4;
     private final static int DATAFILE_HDR_SIZE = 8192;
     
-    private final Logger logger = Logger.getLogger(this.getClass().getName());
+    private final static Logger logger = Logger.getLogger(ChromeCacheExtractor.class.getName());
     
     private static final String VERSION_NUMBER = "1.0.0";
     private final String moduleName;
@@ -79,10 +79,9 @@ final class ChromeCacheExtractor {
      
     private final Content dataSource;
     private final IngestJobContext context;
-    private IngestServices services = IngestServices.getInstance();
+    private final IngestServices services = IngestServices.getInstance();
     private Case currentCase;
     private FileManager fileManager;
-    private FileTypeDetector fileTypeDetector;
  
     
     private final Map<String, CacheFileCopy> filesTable = new HashMap<>();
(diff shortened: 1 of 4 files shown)


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -0,0 +1,1156 @@
...
+     * @return 
+     */
+    private String getRelOutputFolderName() {
+        return relOutputFolderName;
+    }
+    
+    /**
+     * Extracts the data from Chrome cache
+     */
+    void getCache() {
```
kellykelly3: would it make sense to break getCache into multiple functions?  Maybe put the code in the nested loop into its own function?

raman-bt: Yes, this method could and should be simplified and broken  up a little.

I realize that this method is going to need some refactoring in order to handle multiple user profiles ( and hence caches) in a data source.  

Therefore Id like to defer this and address as it as part of that.

kellykelly3: OK.

rcordovano: @raman-bt and @kellykelly3, what is the status of this PR? I see that there are still Codacy issues that can be easily remedied, is this otherwise ready to go?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -0,0 +1,1265 @@
...
+        public ByteBuffer getByteBuffer() {
+            return byteBuffer;
+        }
+        AbstractFile getAbstractFile() {
+            return abstractFile;
+        } 
+    }
+
+    ChromeCacheExtractor(Content dataSource, IngestJobContext context ) {
+        moduleName = NbBundle.getMessage(ChromeCacheExtractor.class, "ChromeCacheExtractor.moduleName");
```
kellykelly3: Should use the @message annotation instead of manually modifying Bundles.properties.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -0,0 +1,1265 @@
...
+     */
+    void moduleInit() throws IngestModuleException {
+        
+        try {
+            currentCase = Case.getCurrentCaseThrows();
+            fileManager = currentCase.getServices().getFileManager();
+             
+            // Create an output folder to save any derived files
+            absOutputFolderName = RAImageIngestModule.getRAOutputPath(currentCase, moduleName);
+            relOutputFolderName = RAImageIngestModule.getRelModuleOutputPath() + File.separator + moduleName;
```
kellykelly3: Consider using Path class for building Paths

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -0,0 +1,1265 @@
...
+            try {
+                entry.getValue().getFileCopy().getChannel().close();
+                entry.getValue().getFileCopy().close();
+               
+                File tmpFile = new File(tempFilePathname);
+                if (!tmpFile.delete()) {
+                    tmpFile.deleteOnExit();
+                }
+            } catch (IOException ex) {
+                logger.log(Level.SEVERE, String.format("Failed to delete cache file copy %s", tempFilePathname), ex);
```
kellykelly3: If we are unable to delete a temp file is that a SEVERE error?  Maybe this should be information or a warning?

rcordovano: I would call this a SEVERE error. Here is the way I think about the error levels:

INFO - Something normal/expected happened and we are logging it to records the flow of processing, establishing context for other messages. For example, the ingest job started and ingest job finished messages. 
WARNING - An error occurred but it is a result of the input, not a problem with the system. For example, it appears that a file is corrupted. Or library code experienced an error processing a file.
SEVERE - Something is wrong with the system or the code. For example, a column is missing from the case database or a case database query fails, i.e., many TskCoreExceptions. Or the RegRipper executable laid down by the installer is missing.

In this case, I would say that an inability to delete a temp file that we wrote says something is wrong with the system.

raman-bt: LOL, I would have made it a Warning, were I not aware of Richard's sentiments on this subject, I've made that mistake a few times.

That said, I do agree, not being able to delete a temp file shouldn't  be a  SEVERE error, its does not have the same import and implications on working of Autopsy, as does not being able to find RegRipper, or being unable to open a database, or not finding an expected  column in a table.  

I often wish we had additional error levels like MAJOR and MINOR , in addition to SEVERE.  SEVERE, IMHO, should be reserved for "cannot proceed further, going to shut down now...".  This error would be a MINOR,

rcordovano: Please hold that thought, we will continue to work towards a standard offline. Please call it whichever way you two want here.

PR #4531, opened 2019-02-08
4715 vcard tsk contact artifacts

dgrove727:


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/VcardParser.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Is the supplied content a vCard file?
+     * 
+     * @param content The content to check.
+     * 
+     * @return True if the supplied content is a vCard file; otherwise false.
+     */
+    static boolean isVcardFile(Content content) {
+        try {
+            byte[] t = new byte[VCARD_HEADER.length()];
```
APriestman: This could be moved to inside the if block

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -299,6 +309,69 @@ private ProcessResult processMBox(AbstractFile abstractFile) {
...
+        File file = new File(fileName);
+
+        long freeSpace = services.getFreeDiskSpace();
+        if ((freeSpace != IngestMonitor.DISK_FREE_SPACE_UNKNOWN) && (abstractFile.getSize() >= freeSpace)) {
+            logger.log(Level.WARNING, String.format("Not enough disk space to write file '%s' (id=%d) to disk.",
+                    abstractFile.getName(), abstractFile.getId())); //NON-NLS
+            IngestMessage msg = IngestMessage.createErrorMessage(EmailParserModuleFactory.getModuleName(), EmailParserModuleFactory.getModuleName(),
+                    Bundle.ThunderbirdMboxFileIngestModule_errorMessage_outOfDiskSpace(abstractFile.getName(), abstractFile.getId()));
+            services.postMessage(msg);
+            return ProcessResult.OK;
```
APriestman: I realize you're copying the other methods in this class, but do you know the reasoning behind returning OK after failures like this instead of ERROR?

dgrove727: My guess is if you have a really large file that's too big to be written, we don't necessarily want to stop ingest over that. Sort of the same logic we use when returning 'OK' for files that fail to be parsed correctly. It's a problematic file for ingest, but not necessarily an application error. At least that's my interpretation.

rcordovano:
Use custom attribute types for distinguishing phone number types instead of new standard attributes.

PR #4535, opened 2019-02-13
4683 Initial version of portable case

APriestman:


Core/src/org/sleuthkit/autopsy/report/CreatePortableCaseModule.java
```diff
@@ -0,0 +1,459 @@
...
+            skCase = currentCase.createPortableCase(caseName, caseFolder);
+        } catch (TskCoreException ex) {
+            handleError("Error creating case " + caseName + " in folder " + caseFolder.toString(),
+                Bundle.CreatePortableCaseModule_createCase_errorCreatingCase(), ex, progressPanel);  
+            return;
+        }
+        
+        // Create the base folder for the copied files
+        copiedFilesFolder = Paths.get(caseFolder.toString(), FILE_FOLDER_NAME).toFile();
+        copiedFilesFolder.mkdir();
```
eugene7646: Shouldn't we be checking return argument here? mkdir() returns true/false.

PR #4537, opened 2019-02-13
1190 basic edge module

kellykelly3:
This code is for Jira 1190 to create a module to support edge that will run the the nirsoft tools dumping the output files from WebCacheV01.dat to a temporary file.


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+    ExtractEdge() throws NoCurrentCaseException{
+        moduleName = NbBundle.getMessage(Chrome.class, "ExtractEdge.moduleName");
+        moduleTempResultsDir = RAImageIngestModule.getRATempPath(Case.getCurrentCaseThrows(), EDGE) + File.separator + "results";
+    }
+    
+    @Override
+    void process(Content dataSource, IngestJobContext context) {
+        this.dataSource = dataSource;
+        this.context = context;
+        dataFound = false;
```
rcordovano: I see that this dataFound field is yet another protected field in the super class. For this story, let's do the partial fix of adding a protected setter for the super class field and then this new code can use that instead of writing directly to the field.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+        this.getBookmark(); // Not implemented yet
+    }
+    
+    void processWebCache(){
+        Path path = Paths.get("ESEDatabaseView", ESE_TOOL_NAME);
+        ESE_TOOL_FILE =  InstalledFileLocator.getDefault().locate(path.toString(), ExtractEdge.class.getPackage().getName(), false); //NON-NLS
+        if (ESE_TOOL_FILE == null) {
+            this.addErrorMessage(
+                    NbBundle.getMessage(this.getClass(), "ExtractEdge.process.errMsg.unableFindESEViewer", this.getName()));
+            logger.log(Level.SEVERE, "Error finding ESEDatabaseViewer program "); //NON-NLS
```
rcordovano: I'm being a bit nit-picky here, but if I read in the log that there was an error I am going to expect to see a stack trace for an exception. As a small refinement, I would prefer to just have a log message like  "Failed to find..."

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+
+        // get WebCacheV01.dat files
+        org.sleuthkit.autopsy.casemodule.services.FileManager fileManager = currentCase.getServices().getFileManager();
+        List<AbstractFile> webCachFiles;
+        try {
+            webCachFiles = fileManager.findFiles(dataSource, EDGE_WEBCACHE_NAME); //NON-NLS
+        } catch (TskCoreException ex) {
+            this.addErrorMessage(NbBundle.getMessage(this.getClass(), "ExtractEdge.process.errMsg.errGettingWebCacheFiles",
+                    this.getName()));
+            logger.log(Level.WARNING, "Error fetching 'index.data' files for Internet Explorer history."); //NON-NLS
```
rcordovano: If an exception is thrown by the FileManager something is very wrong, so this should be Level.SEVERE and not Level.WARNING. Also, we should almost never throw away an exception like this - it makes it much, much harder to debug the problem. If you pass the exception as the last argument to logging method a stack trace will be logged and we will have a better idea of what went wrong with the findFiles call.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+            this.addErrorMessage(
+                    NbBundle.getMessage(this.getClass(), "ExtractEdge.process.errMsg.unableFindESEViewer", this.getName()));
+            logger.log(Level.SEVERE, "Error finding ESEDatabaseViewer program "); //NON-NLS
+        }
+        
+         final String esedumper = ESE_TOOL_FILE.getAbsolutePath();
+
+        // get WebCacheV01.dat files
+        org.sleuthkit.autopsy.casemodule.services.FileManager fileManager = currentCase.getServices().getFileManager();
+        List<AbstractFile> webCachFiles;
```
rcordovano: webCacheFiles would be a better variable name here.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+        } catch (TskCoreException ex) {
+            this.addErrorMessage(NbBundle.getMessage(this.getClass(), "ExtractEdge.process.errMsg.errGettingWebCacheFiles",
+                    this.getName()));
+            logger.log(Level.WARNING, "Error fetching 'index.data' files for Internet Explorer history."); //NON-NLS
+            return;
+        }
+
+        if (webCachFiles.isEmpty()) {
+            String msg = NbBundle.getMessage(this.getClass(), "ExtractEdge.process.errMsg.noWebCachFiles");
+            logger.log(Level.INFO, msg);
```
rcordovano: Log messages should never be localized. The most likely consumer of them is someone on the Basis dev team and we speak English; English is also a de facto lingua franca, so there is no up side in making logs the primarily English-speaking Basis devs cannot read.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+            return;
+        }
+
+        dataFound = true;   
+        
+        if(!PlatformUtil.isWindowsOS()){
+            logger.log(Level.WARNING, "Edge data found, unable to parse on non-windows system."); //NON-NLS
+            return;
+        }
```
rcordovano: It seems like this should happen up front somewhere, perhaps as the very first line of the calling process() method.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+        
+        String temps;
+        String indexFileName;
```
rcordovano: These local variables are only used in the for loop; it would be better in terms of "locality of reference" to define these inside of the loop. Note that there is no penalty for doing so - the compiler will do what it needs to do regarding space efficiency of local variables, so we need not concern ourselves with that and can follow the best practice of locality of reference.

Also, what is temps? I'm thinking tempFilePath or something similar would be better.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+            indexFileName = "WebCacheV01" + Integer.toString((int) indexFile.getId()) + ".dat";
+            temps = RAImageIngestModule.getRATempPath(currentCase, EDGE) + File.separator + indexFileName; //NON-NLS
+            File datFile = new File(temps);
+            if (context.dataSourceIngestIsCancelled()) {
+                break;
+            }
+            try {
+                ContentUtils.writeToFile(indexFile, datFile, context::dataSourceIngestIsCancelled);
+            } catch (IOException e) {
+                logger.log(Level.WARNING, "Error while trying to write index.dat file " + datFile.getAbsolutePath(), e); //NON-NLS
```
rcordovano: If we fail to write to disk here, this is a SEVERE error.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+                continue;
+            }
+            
+            File resultsDir = new File(moduleTempResultsDir + Integer.toString((int) indexFile.getId()));
+            resultsDir.mkdirs();
+            executeDumper(esedumper, datFile.getAbsolutePath(), "webcache", resultsDir.getAbsolutePath());
+             
+            this.getHistory(); // Not implemented yet
+            this.getCookie(); // Not implemented yet
+            this.getDownload(); // Not implemented yet
```
rcordovano: Are these supposed to be called from the process() method? If not, we will need to be careful with the cancellation checks to ensure the recent activity module stays responsive to cancellation with cancellation checks after each call.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+        commandLine.add(outputDir + "\\" + inputFilePrefix + "_*.csv");
+         
+        ProcessBuilder processBuilder = new ProcessBuilder(commandLine);
+        processBuilder.redirectOutput(new File(outputFileFullPath));
+        processBuilder.redirectError(new File(errFileFullPath));
+
+        try{
+            ExecUtil.execute(processBuilder, new DataSourceIngestModuleProcessTerminator(context));
+        }catch(IOException ex){
+            logger.log(Level.SEVERE, "Unable to execute ESEDatabaseView to process Edge file." , ex); //NON-NLS
```
rcordovano: Which Edge file?

Personally, I think it is better to throw exceptions than have boolean or error code return values; the general policy is lower level code throws exceptions for handling by higher level code. I would just propagate the IOException here - it is cleaner than having multiple return statements fro true/false.

Also, be sure to format (ALT+SHIFT+F) before check in. For example, the formatter will add space between the reserved words and the "curly" braces of this try-catch.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+import org.sleuthkit.autopsy.coreutils.PlatformUtil;
+import org.sleuthkit.autopsy.datamodel.ContentUtils;
+import org.sleuthkit.autopsy.ingest.DataSourceIngestModuleProcessTerminator;
+import org.sleuthkit.autopsy.ingest.IngestJobContext;
+import org.sleuthkit.autopsy.ingest.IngestServices;
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.Content;
+import org.sleuthkit.datamodel.TskCoreException;
+
+public class ExtractEdge extends Extract{
```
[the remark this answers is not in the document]
kellykelly3: The other browser classes are private (the public in ExtractEdge was probably from habit), but they are not final.  It does make sense to make this class final however that will make it inconsistent with the others.  Is that ok?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: Brian's requirement in the Jira was that we log that there are Edge files, but we cannot process them.
I can put the check in the beginning of process, but we need to continue far enough to determine if there are edge files.  I honestly didn't know if what I did here would be enough to fufill the requirement and now that I type this I realize that it might not be.  I am not sure the effect of the dataFound boolean on the gui, if any.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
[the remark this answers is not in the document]
rcordovano: Definitely. The code in the RecentActivity NBM is not model code so there is really no precedent for us to adhere to in this case. In fact, over time, I would like to improve the code in this module as an exercise in technical debt payoff, so making all brand new code adhere to more rigorous standards is a good thing.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: That was lazy copy & paste.  Sorry about that.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
rcordovano: How about doing the query for the files at the start of the process() method? You can then do an early exit if you get no files, can get the presence of the files logged still do an early exit if the platform is not Windows and after that an early exit if the executable cannot be located (assuming that the same executable will be used by all of the  sub-methods; the location could be a local variable passed as an argument to each of the processXXX() methods by the process() method).

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: Very true.  Maybe a future task is to review the other classes, the similar error is a warning in ExtractIE.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: I realize that I am in general missing the cancellation checks.  Maybe we can talk about when\where it makes sense to put them.  

Those calls are in the processWebCache because I don't want to have to open and dump WebCacheV01 multiple times, just once for everything that needs the tables from that file.  I am guesses the cancellation should be checked before each of those calls.

rcordovano: Sure, we can talk about cancellation checks. In general, you want to do them frequently.

If you take my suggestion of fetching and dumping the files up front and passing them as arguments to the processXXX() methods I thin that you might have a more readable and easier to work with structuring of the code...

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: I agree with that too, but I noticed that the examples I was looking at didn't.   It will make me feel better to change it.  :-)

rcordovano: Me, too! :)

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+    private Content dataSource;
+    private IngestJobContext context;
+    
+    private static String ESE_TOOL_NAME = "ESEDatabaseView.exe";
+    private static File ESE_TOOL_FILE;
+    private static String EDGE_WEBCACHE_NAME = "WebCacheV01.dat";
+    private static String EDGE = "Edge";
+
+    ExtractEdge() throws NoCurrentCaseException{
+        moduleName = NbBundle.getMessage(Chrome.class, "ExtractEdge.moduleName");
```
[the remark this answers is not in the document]
kellykelly3: You are correct.  The funny thing is that I don't even use the moduleName in my class.  I should have investigated if the super class used it and if there was a real need to set it.  The superclass does have a get method.  I will overload it and return the value using the Bundle accessor.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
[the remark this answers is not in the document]
kellykelly3: After cleaning up the getModule() name in the other files as part of my next pull request, there is a lot that could be cleaned up.  :-)  I was going for this initial pass to just look the same, now you've given me the clear to be more thoughtful.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Bundle.properties
```diff
@@ -105,3 +105,8 @@ SearchEngineURLQueryAnalyzer.toString=Name\: {0}\n\
 SearchEngineURLQueryAnalyzer.parentModuleName.noSpace=RecentActivity
 SearchEngineURLQueryAnalyzer.parentModuleName=Recent Activity
 UsbDeviceIdMapper.parseAndLookup.text=Product\: {0}
+ExtractEdge.moduleName=Microsoft Edge
+ExtractEdge.process.errMsg.unableFindESEViewer=Unable to find ESEDatabaseViewer
+ExtractEdge.process.errMsg.errGettingWebCacheFiles=Error retrieving Edge file
+ExtractEdge.process.errMsg.noWebCachFiles=No Edge WebCache file found
+ExtractEdge.process.errMsg.errWriteFile={0}\: Error while trying to write file\:{1}
```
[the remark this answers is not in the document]
kellykelly3: I noticed that in case.java the message annotation was before each function.  Is there a reason for that?  If there is one that is repeatedly used can I define them at the top of the file?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,197 @@
...
+        this.processWebCache();
+        
+         // Bookmarks come from spartan.edb different file
+        this.getBookmark(); // Not implemented yet
+    }
+    
+    void processWebCache(){
+        Path path = Paths.get("ESEDatabaseView", ESE_TOOL_NAME);
+        ESE_TOOL_FILE =  InstalledFileLocator.getDefault().locate(path.toString(), ExtractEdge.class.getPackage().getName(), false); //NON-NLS
+        if (ESE_TOOL_FILE == null) {
```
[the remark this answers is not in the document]
kellykelly3: I missed that string in path when I was cleaning up.  I have moved it to a constant and I will go through and add the NON-NLS as needed.

Yes, if the tool cannot be located we are done. 

I agree, I am reorganizing and writing some little functions.  I do too wish that process allowed throwing exceptions, it would make the code in process a little cleaner.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: I moved it.  You'll have to let me know what you think of the flow.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: I put them in, please let me know what you think.  Too much?  Not enough?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
rcordovano: The flow looks good!

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,256 @@
...
+
+        if (context.dataSourceIngestIsCancelled()) {
+            return;
+        }
+
+        // Bookmarks come from spartan.edb different file
+        this.getBookmark(); // Not implemented yet
+    }
+
+    void processWebCache(String eseDumperPath, List<AbstractFile> webCachFiles) throws IOException {
```
rcordovano: Looks like a typo: webCachFiles => webCacheFiles.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Bundle.properties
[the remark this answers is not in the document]
rcordovano: Yes, I think that you can. It certainly would be sensible to do so.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
rcordovano: Looking good. There is one addition I would make, see other comment,

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
rcordovano: Yes, this entire NBM could use a good scrubbing. It is quite old and our ways of doing things has evolved.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,256 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.Content;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Extract the bookmarks, cookies, downloads and history from the Microsoft Edge
+ * files
+ *
+ * @author kelly
+ */
```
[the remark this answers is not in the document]
kellykelly3: done

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,256 @@
...
+final class ExtractEdge extends Extract {
+
+    private static final Logger logger = Logger.getLogger(ExtractIE.class.getName());
```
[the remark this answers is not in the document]
kellykelly3: done

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
kellykelly3: At least when I spell things wrong I'm consistent.  :-)

Fixed.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractEdge.java
```diff
@@ -0,0 +1,256 @@
...
+
+        // No edge files found 
+        if (webCacheFiles == null && spartanFiles == null) {
+            return;
+        }
+
+        dataFound = true;
+
+        if (!PlatformUtil.isWindowsOS()) {
+            logger.log(Level.INFO, "Microsoft Edge files found, unable to parse on Non-Windows system"); //NON-NLS
```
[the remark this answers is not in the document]
kellykelly3: Changed.

We should add to the technical debt list to go through all of Recent Activity to make the usage of the log level more consistent.

PR #4540, opened 2019-02-13
4673 Command line reports 

eugene7646:


Core/src/org/sleuthkit/autopsy/modules/case_uco/CaseUcoFormatExporter.java
```diff
@@ -0,0 +1,391 @@
...
+import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.autopsy.coreutils.MessageNotifyUtil;
+import org.sleuthkit.autopsy.datamodel.ContentUtils;
+import org.sleuthkit.autopsy.ingest.IngestManager;
+import org.sleuthkit.autopsy.report.ReportProgressPanel;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskData;
+
+public class CaseUcoFormatExporter {
```
dgrove727: I realize a vast majority of the code here is essentially copy/paste, but somehow the Javadocs got left out off of the class and methods within. I would at least put these back in.

dgrove727: I also notice a few Codacy things that are probably easy to address.

Core/src/org/sleuthkit/autopsy/commandline/CommandLineIngestManager.java
```diff
@@ -0,0 +1,602 @@
...
+                }
+
+                // read options panel configuration
+                String rootOutputDir = UserPreferences.getCommandLineModeResultsFolder();
+                if (!(new File(rootOutputDir).exists())) {
+                    LOGGER.log(Level.SEVERE, "The output directory doesn't exist {0}", rootOutputDir);
+                    System.out.println("The output directory doesn't exist " + rootOutputDir);
+                    return;
+                }
+                rootOutputDirectory = Paths.get(rootOutputDir);
```
rcordovano: Given that we write the case name and data source path to the log and to the console, shouldn't we also write the output directory path as well?

Core/src/org/sleuthkit/autopsy/commandline/CommandLineIngestManager.java
```diff
@@ -0,0 +1,602 @@
...
+
+                // open case
+                Case caseForJob;
+                try {
+                    caseForJob = openCase(baseCaseName);
+                } catch (CaseActionException ex) {
+                    LOGGER.log(Level.SEVERE, "Error creating or opening case " + baseCaseName, ex);
```
rcordovano: Up to this point, we are writing to both the log and to System.out. After this we are just writing to the log. If we are not going to continue writing to both, it seems that we should at least have some sort of outcome message written to the console. Thoughts?

If we were going to continue to log everything in both places, one approach would be to write some private helper methods to write to both the log and the console.

eugene7646: I have added the System.out messages to all logging in the JobProcessingTask.run() method. My thought process was that System.out are there to help the user debug initial configuration and input parameter errors. I think for debugging "bigger" issues, like errors during ingest or in DSPs, we should continue to rely on application logs, so I think limiting System.out  logging to only the main JobProcessingTask.run() method is sufficient. @rcordovano Agree?

rcordovano: Sure. I do think we should add some sort of completion message to the console when the task is done.

eugene7646: Oh yes, I have duplicated all logging in JobProcessingTask.run(), to have both application log and System.out, including task completion. I just haven't pushed up my changes yet.

Core/src/org/sleuthkit/autopsy/commandline/CommandLineIngestManager.java
```diff
@@ -0,0 +1,602 @@
...
+
+    public void start() {
+        JobProcessingTask jobProcessingTask = new JobProcessingTask();
+        jobProcessingTaskFuture = jobProcessingExecutor.submit(jobProcessingTask);
+    }
+
+    public void stop() {
+
+        jobProcessingTaskFuture.cancel(true);
+        jobProcessingExecutor.shutdown();
```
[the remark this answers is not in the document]
rcordovano: Okay, so we are going to go with an exceptions firewall instead, for now.

PR #4550, opened 2019-02-21
1197 safari history

kellykelly3:
Added support for getting history from the safari history.db file.  Moved create attribute functions from ExtractEdge to Extract for use in ExtractSafari.


kellykelly3:
Changes will be reviewed in another pull request.

PR #4563, opened 2019-02-25
4696 add custom artifact handling

dannysmyda:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -65,101 +119,61 @@ public static String getEmailAddressAttrString() {
...
-            BlackboardArtifact artifactForInstance = null;
-            if (BlackboardArtifact.ARTIFACT_TYPE.TSK_INTERESTING_ARTIFACT_HIT.getTypeID() == artifact.getArtifactTypeID()) {
-                // Get the associated artifactForInstance
-                BlackboardAttribute attribute = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ASSOCIATED_ARTIFACT));
-                if (attribute != null) {
-                    artifactForInstance = Case.getCurrentCaseThrows().getSleuthkitCase().getBlackboardArtifact(attribute.getValueLong());
+            int artifactTypeID = artifact.getArtifactTypeID();
+            if (artifactTypeID == ARTIFACT_TYPE.TSK_KEYWORD_HIT.getTypeID()) {
+                BlackboardAttribute setNameAttr = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SET_NAME));
+                if (setNameAttr != null
```
rcordovano: Please add a comment to explain the surprising use of the TSK_SET_NAME ATTRIBUTE, thanks!

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -65,101 +119,61 @@ public static String getEmailAddressAttrString() {
...
-                        || artifactTypeID == ARTIFACT_TYPE.TSK_MESSAGE.getTypeID()) {
-
-                    String value = null;
-                    if (null != artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER))) {
-                        value = artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER)).getValueString();
-                    } else if (null != artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM))) {
-                        value = artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM)).getValueString();
-                    } else if (null != artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO))) {
-                        value = artifactForInstance.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO)).getValueString();
+            for (BlackboardAttribute attribute : artifact.getAttributes()) {
```
rcordovano: I think we need a conditional here - if we have processed a keyword hit artifact we do not need to continue.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -65,101 +119,61 @@ public static String getEmailAddressAttrString() {
...
+                Integer customAttributeTypeId = attribute.getAttributeType().getTypeID();
```
rcordovano: I thin the use of "custom" for the attribute type id variable is not quite accurate.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
```diff
@@ -65,101 +119,61 @@ public static String getEmailAddressAttrString() {
...
-                            if (inst != null) {
-                                eamArtifacts.add(inst);
-                            }
-                        }
+
+                    if (attributeName.startsWith("TSK_PHONE_NUMBER_")) {
+                        addCorrelationAttributeToList(eamArtifacts, artifact,
+                                attribute,
+                                CorrelationAttributeInstance.PHONE_TYPE_ID
+                        );
```
rcordovano: @bcarrier, are we going to add this sort of "secret handshake" on custom attribute names to the module developer documentation?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeNormalizer.java
```diff
@@ -164,6 +165,9 @@ private static String normalizeEmail(String data) throws CorrelationAttributeNor
     private static String normalizePhone(String data) throws CorrelationAttributeNormalizationException {
         if (data.matches("\\+?[0-9()\\-\\s]+")) {
             String phoneNumber = data.replaceAll("[^0-9\\+]", "");
+            if (phoneNumber.length() <= MIN_PHONE_LENGTH) {
+                throw new CorrelationAttributeNormalizationException(String.format("Data was expected to be a valid phone number: %s", data));
+            }
```
APriestman: With the phone number validation move, I'm now getting stack traces in the log every time it finds a short or otherwise invalid phone number. I believe we were previously just not recording them - they can be valid but are not interesting for correlation.

dannysmyda: @rcordovano Should this logic be moved back over? I don't see why this shouldn't be here (to enforce this check across all modules). My first reaction is that we probably shouldn't log these things so heavily, who cares if its not valid?

dannysmyda: @rcordovano Do we study production logs heavily, that's the only way I can see this being useful (i.e we find email formats that are actually real and common in production)?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamArtifactUtil.java
rcordovano: We will leave this as is for now. A story will be created for defining how custom attributes should be defined to work with the CR.

rcordovano:
Work "stashed" as 4696-Add-Custom-Artifact-Handling branch.

PR #4564, opened 2019-02-25
1200 safari bookmarks

kellykelly3:
Moved a copy of the plist jar into RecentActivity and added support for Safari Bookmarks.


kellykelly3:
Changes will be review in PR
https://github.com/sleuthkit/autopsy/pull/4572

PR #4565, opened 2019-02-25
1198 safari downloads

kellykelly3:
I am 90% sure this code for downloads is the correct code.  I need to test with a Downloads.plist with more than one file in it which I do not currently have.


kellykelly3:
Changes will be review in PR
https://github.com/sleuthkit/autopsy/pull/4572

PR #4566, opened 2019-02-26
3921 html content viewer

dgrove727:


Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public List<String> getSupportedMIMETypes() {
+        return Arrays.asList(SUPPORTED_MIMETYPES);
+    }
+
+    @Override
+    public void setFile(AbstractFile file) {
+        WindowManager.getDefault().getMainWindow().setCursor(Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR));
+        abstractFile = file;
```
wschaeferB: variable named abstractFile does not exist, which prevents pr from building

PR #4567, opened 2019-02-26
4750 Add new layout file type for externally defined files

APriestman:


Core/src/org/sleuthkit/autopsy/datamodel/LayoutFileNode.java
```diff
@@ -30,6 +30,7 @@
 import org.sleuthkit.autopsy.actions.AddContentTagAction;
 import org.sleuthkit.autopsy.actions.DeleteFileContentTagAction;
 import org.sleuthkit.autopsy.coreutils.ContextMenuExtensionPoint;
+import static org.sleuthkit.autopsy.datamodel.FileNode.getIconForFileType;
```
esaunders: Codacy complains about unused import

PR #4585, opened 2019-03-04
Removed the plist jar from recentActivity, it should be using the one…

kellykelly3:
… from corelib now.

This is working correctly now for me.  I tested that both my Safari code work as well as the code for the PListViewer.


RecentActivity/ivy.xml
```diff
@@ -7,6 +7,6 @@
       
     </configurations>
 	<dependencies>
-		<dependency conf="recent-activity->default" org="com.googlecode.plist" name="dd-plist" rev="1.20"/>
+<!--		<dependency conf="recent-activity->default" org="com.googlecode.plist" name="dd-plist" rev="1.20"/> -->
```
rcordovano: Can just remove this instead of commenting it out?

PR #4586, opened 2019-03-04
4755 open in external viewer shortcut

wschaeferB:


ImageGallery/src/org/sleuthkit/autopsy/imagegallery/gui/drawableviews/Bundle.properties-MERGED
```diff
@@ -1,4 +1,4 @@
-DrawableTileBase.externalViewerAction.text=Open in External Viewer
+DrawableTileBase.externalViewerAction.text=Open in External Viewer (Ctrl+E)
```
millmanorama: does this shortcut actually work from imagegallery?

wschaeferB: It does not, I had attempted to remove those text changes from image gallery as a result of it not working in image gallery, (I added them before testing it there) looks like i missed some.

millmanorama: That is too bad, It might not be that much work to make it work there also...

wschaeferB:
I'll take a look, I honestly only tried to get it working where we already
had the bookmark shortcut working.

ImageGallery/src/org/sleuthkit/autopsy/imagegallery/gui/drawableviews/Bundle.properties-MERGED
wschaeferB: ll take a look, I honestly only tried to get it working where we already had the bookmark shortcut working.

wschaeferB: Alright got it sorted out and working in image gallery

millmanorama: awesome!

millmanorama:
Maybe I missed something, but I just tested this in Image Gallery and it didn't work, neither in the communications visualization tool...  It did work in timeline!

PR #4593, opened 2019-03-06
1192 manually created events

millmanorama:
depends on https://github.com/sleuthkit/sleuthkit/pull/1454


Core/src/org/sleuthkit/autopsy/timeline/actions/AddManualEvent.java
```diff
@@ -0,0 +1,299 @@
...
+     */
+    public AddManualEvent(TimeLineController controller, Long epochMillis) {
+        super(Bundle.AddManualEvent_text());
+        this.controller = controller;
+        setGraphic(new ImageView(ADD_EVENT_IMAGE));
+        setLongText(Bundle.AddManualEvent_longText());
+
+        setEventHandler(actionEvent -> {
+            //shoe the dialog and if it completed normally add the event.
+            new EventCreationDialog(controller, epochMillis).showAndWait().ifPresent(this::addEvent);
```
kellykelly3: This might be an issue in EventCreationDialog, but when the dialog pops up its not confined to the bounds of the main window.  In the particular case of my setup the dialog pops up on my other screen, far away from the main window.   This also means that I can put the main window on top of the popup dialog thus hiding the the popup causing the main window to appear "stuck".

Core/src/org/sleuthkit/autopsy/timeline/actions/AddManualEvent.java
```diff
@@ -0,0 +1,299 @@
...
+    private static class EventCreationDialog extends Dialog<ManualEventInfo> {
+
+        /** Custom DialogPane defined below. */
+        private final EventCreationDialogPane eventCreationDialogPane;
+
+        EventCreationDialog(TimeLineController controller, Long epochMillis) {
+            this.eventCreationDialogPane = new EventCreationDialogPane(controller, epochMillis);
+            setTitle("Add Event");
+            setDialogPane(eventCreationDialogPane);
+
```
kellykelly3: In reference to my above comment you should be able to put code here, before the dialog is shown.  JDialog has a function setLocationRelativeTo(), I know this is not a JDialog, but I would assume there is something similar.

Core/src/org/sleuthkit/autopsy/timeline/actions/AddManualEvent.java
kellykelly3: I also noticed that I was allowed to popup more than one add event dialog.

Core/src/org/sleuthkit/autopsy/timeline/TimeLineController.java
```diff
@@ -116,7 +118,8 @@
 
     private static final ReadOnlyObjectWrapper<TimeZone> timeZone = new ReadOnlyObjectWrapper<>(TimeZone.getDefault());
```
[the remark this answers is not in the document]
millmanorama: Even though netbeans compain, we have the convention of using lowercase for logger, since it is not a 'simple' constant.  Similarly timeZone will never refer to another object, but it's wrapped value changes.

rcordovano:
Jonathan, please merge in timeline-event-mgr and resolve the conflict, thanks!

millmanorama:
> Jonathan, please merge in timeline-event-mgr and resolve the conflict, thanks!

done

rcordovano:
@kellykelly3, I need you to mark this with your approval, thanks.

@millmanorama, please address the merge conflicts, thanks!

rcordovano:
@millmanorama, it looks like the Codacy issues can be resolved, at least the ones about potential final fields.

PR #4595, opened 2019-03-06
1208 zoneidentifier

kellykelly3:
Added support for finding :Zone.Identifier Alternate Data Stream files


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+        } catch (TskCoreException ex) {
+            addErrorMessage(Bundle.ExtractZone_process_errMsg_find());
+            LOG.log(Level.SEVERE, "Unable to find zone identifier files, exception thrown. ", ex);
+        }
+
+        if (zoneFiles == null || zoneFiles.isEmpty()) {
+            return;
+        }
+        
+        ArrayList<Long> knownPathIDs = null;
```
millmanorama: As @bcarrier suggested, I would make knownPathIDs a set.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+        try {
+            knownPathIDs = getPathIDsForType(BlackboardArtifact.ARTIFACT_TYPE.TSK_WEB_DOWNLOAD);
+        } catch (TskCoreException ex) {
+            addErrorMessage(Bundle.ExtractZone_process_errMsg());
+            LOG.log(Level.SEVERE, "Failed to build PathIDs List for TSK_WEB_DOWNLOAD", ex);
+        }
+
+        if (knownPathIDs == null) {
+            return;
```
millmanorama: the only way knownPathIDs can be null is if getPathIDsForType throws an exception, so I would put the return in the try block, and skip the null check.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+
+        try {
+            zoneInfo = new ZoneIdentifierInfo(zoneFile);
+        } catch (IOException ex) {
+            String message = String.format("Unable to parse temporary File for %s", zoneFile.getName()); //NON-NLS
+            LOG.log(Level.WARNING, message, ex);
+        }
+
+        if (zoneInfo == null) {
+            return;
```
millmanorama: again see comment about putting the return in the catch block.  This is not a functional thing just stylistic so maybe @bcarrier or @rcordovano can way in on what they prefer.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+     * Creates a list of PathIDs for the given Artifact type.
+     *
+     * @param type BlackboardArtifact.ARTIFACT_TYPE
+     *
+     * @return A list of PathIDs
+     *
+     * @throws TskCoreException
+     */
+    private ArrayList<Long> getPathIDsForType(BlackboardArtifact.ARTIFACT_TYPE type) throws TskCoreException {
+        ArrayList<Long> idList = new ArrayList();
```
millmanorama: add <> to "new ArrayList()" to avoid warnings

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+         * into a HashMap.
+         *
+         * @param zoneFile The ZoneIdentifier file
+         *
+         * @throws FileNotFoundException
+         * @throws IOException
+         */
+        ZoneIdentifierInfo(AbstractFile zoneFile) throws FileNotFoundException, IOException {
+            String line;
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(new ReadContentInputStream(zoneFile)))) {
```
millmanorama: I would consider using the [Properties ](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-)class and avoiding this low level stream reading code.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+     * @param zoneInfo     Zone Indentifer file wrapper object
+     *
+     * @return TSK_DOWNLOAD_SOURCE object for given parameters
+     */
+    private BlackboardArtifact createDownloadSourceArtifact(AbstractFile downloadFile, ZoneIdentifierInfo zoneInfo) {
+
+        Collection<BlackboardAttribute> bbattributes = new ArrayList<>();
+
+        bbattributes.add(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_URL,
+                RecentActivityExtracterModuleFactory.getModuleName(),
```
millmanorama: I would use some static imports and bbattributes.addAll(asLists(....)) to remove alot of the noise here.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+                (zoneInfo.getURL() != null) ? zoneInfo.getURL() : "")); //NON-NLS
+
+        bbattributes.add(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DOMAIN,
+                RecentActivityExtracterModuleFactory.getModuleName(),
+                (zoneInfo.getURL() != null) ? NetworkUtils.extractDomain(zoneInfo.getURL()) : "")); //NON-NLS
```
millmanorama: apache StringUtils.defaultString() does exactly this coersion of null to an empty string, and I find it easier to read.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+     * @param zoneInfo ZoneIdentifierInfo file wrapper object
+     *
+     * @return BlackboardArifact for the given parameters
+     */
+    private BlackboardArtifact createDownloadArtifact(AbstractFile zoneFile, ZoneIdentifierInfo zoneInfo) {
+
+        Collection<BlackboardAttribute> bbattributes = createDownloadAttributes(
+                null, null,
+                zoneInfo.getURL(), null,
+                (zoneInfo.getURL() != null ? NetworkUtils.extractDomain(zoneInfo.getURL()) : ""),
```
millmanorama: again, maybe StringUtils.defaultString()

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+    /**
+     * Find a list of all the file that end in <i>:Zone.Identifier<i>.
+     *
+     * @param dataSource Content
+     *
+     * @return A list of zone identifier files
+     *
+     * @throws TskCoreException
+     */
+    private List<AbstractFile> findZoneFiles(Content dataSource) throws TskCoreException {
```
millmanorama: Personally, I would inline this method, since it is only used in one place and can be a one liner

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+            String message = String.format("Unable to parse temporary File for %s", zoneFile.getName()); //NON-NLS
+            LOG.log(Level.WARNING, message, ex);
+        }
+
+        if (zoneInfo == null) {
+            return;
+        }
+
+        AbstractFile downloadFile = getDownloadFile(dataSource, zoneFile);
+        ArrayList<Long> knownPathIDs = getPathIDsForType(BlackboardArtifact.ARTIFACT_TYPE.TSK_WEB_DOWNLOAD);
```
millmanorama: I think there is some cut and paste cleanup to do .  You shouldn't do this query for every file when the results will be the same.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+                }
+            }
+        }
+
+        /**
+         * Get the integer zone id
+         *
+         * @return interger zone id or -1 if unknown
+         */
+        int getZoneId() {
```
millmanorama: ZoneIdentifierInfo is a private inner class so this point is debatable, but codacy suggests all the accessor methods of ZoneIdentifierInfo should also be private.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,398 @@
...
+
+    /**
+     * Wrapper class for information in the :ZoneIdentifier file. The
+     * Zone.Identifier file has a simple format of <i>key<i>=<i>value<i>. There
+     * are four known keys: ZoneId, ReferrerUrl, HostUrl, and
+     * LastWriterPackageFamilyName. Not all browsers will put all values in the
+     * file, in fact most will only supply the ZoneId. Only Edge supplies the
+     * LastWriterPackageFamilyName.
+     */
+    private final class ZoneIdentifierInfo {
```
millmanorama: this should probably be a static class

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/RAImageIngestModule.java
```diff
@@ -87,6 +88,7 @@ public void startUp(IngestJobContext context) throws IngestModuleException {
         extractors.add(registry); // this should run after quicker modules like the browser modules and needs to run before the DataSourceUsageAnalyzer
         extractors.add(osExtract); // this needs to run before the DataSourceUsageAnalyzer
         extractors.add(dataSourceAnalyzer); //this needs to run after ExtractRegistry and ExtractOs
+        extractors.add(zoneInfo); // this needs to run after the web browser modules
```
millmanorama: See comment above about reducing noise.  Similarly, here, I would use addAll(Arrays.asList(.....))  and inline the local variables.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
kellykelly3: I originally had it in the catch, but I thought this was a bit more explicit.   I guess moving it does save the check.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
kellykelly3: I do prefer to make it clear what's going on, I am afraid if someone was to change the code so that null was a possibility, is return would be lost in the catch causing null pointer exceptions later.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/RAImageIngestModule.java
kellykelly3: I think I will leave this one as is.  Having the individual calls, especially with the comments, makes it clear that there is reason to the ordering of the adds.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
kellykelly3: I would think that addAll(asLists()) would be less efficient then just adding them.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,387 @@
...
+    })
+
+    @Override
+    void process(Content dataSource, IngestJobContext context, DataSourceIngestModuleProgress progressBar) {
+
+        progressBar.progress(Bundle.ExtractZone_progress_Msg());
+
+        List<AbstractFile> zoneFiles = null;
+        try {
+            zoneFiles = currentCase.getServices().getFileManager().findFiles(dataSource, ZONE_IDENIFIER_FILE)
```
millmanorama: this is so small, but the semi colon is on the next line from the statement it ends

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Bundle.properties-MERGED
```diff
@@ -48,8 +47,14 @@ ExtractSafari_Error_Getting_History=An error occurred while processing Safari hi
 ExtractSafari_Error_Parsing_Bookmark=An error occured while processing Safari Bookmark files
 ExtractSafari_Error_Parsing_Cookies=An error occured while processing Safari Cookies files
 ExtractSafari_Module_Name=Safari
+ExtractZone_process_errMsg=An error occured processing ':Zone.Indentifier' files.
+ExtractZone_process_errMsg_find=A failure occured while searching for :Zone.Indentifier files.
+ExtractZone_progress_Msg=Extracting :Zone.Identifer files
+INTENET_ZONE=Internet Zone
```
raman-bt: Typo? INTERNET?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,386 @@
...
+ * Extract the <i>:Zone.Indentifier<i> alternate data stream files.  A file with
+ * a <i>:Zone.Indentifier<i> extention contains information about the similarly 
+ * named (with out zone identifer extention) downloaded file.
+ */
+final class ExtractZoneIdentifier extends Extract {
+
+    private static final Logger LOG = Logger.getLogger(ExtractEdge.class.getName());
+
+    private static final String ZONE_IDENIFIER_FILE = "%:Zone.Identifier"; //NON-NLS
+    private static final String ZONE_IDENIFIER = ":Zone.Identifier"; //NON-NLS
```
raman-bt: Typo? missing 'T' in both lines above.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,386 @@
...
+import static org.sleuthkit.datamodel.BlackboardAttribute.ATTRIBUTE_TYPE.TSK_URL;
+import org.sleuthkit.datamodel.Content;
+import org.sleuthkit.datamodel.ReadContentInputStream;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Extract the <i>:Zone.Indentifier<i> alternate data stream files.  A file with
+ * a <i>:Zone.Indentifier<i> extention contains information about the similarly 
+ * named (with out zone identifer extention) downloaded file.
+ */
```
raman-bt: typo? extension?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,386 @@
...
+
+    /**
+     * Process a single Zone Identifier file.
+     *
+     * @param context           IngetJobContext
+     * @param dataSource        Content
+     * @param zoneFile          Zone Indentifier file
+     * @param sourceArtifacts   List for TSK_DOWNLOAD_SOURCE artifacts
+     * @param downloadArtifacts List for TSK_WEB_DOWNLOAD aritfacts
+     */
```
raman-bt: add @throws

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,386 @@
...
+        }
+
+        AbstractFile downloadFile = getDownloadFile(dataSource, zoneFile);
+
+        if (downloadFile != null) {
+            // Only create a new TSK_WEB_DOWNLOAD artifact if one does not exist for downloadFile
+            if (!knownPathIDs.contains(downloadFile.getDataSourceObjectId())) {
+                // The zone identifier file is the parent of this artifact 
+                // because it is the file we parsed to get the data
+                BlackboardArtifact downloadbba = createDownloadArtifact(zoneFile, zoneInfo);
```
raman-bt: CamelCase naming makes it a little easier to read.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractZoneIdentifier.java
```diff
@@ -0,0 +1,386 @@
...
+                if (downloadbba != null) {
+                    downloadArtifacts.add(downloadbba);
+                }
+            }
+            
+            // check if download has a child TSK_DOWNLOAD_SOURCE artifact, if not create one
+            if (downloadFile.getArtifactsCount(TSK_DOWNLOAD_SOURCE) == 0) {
+                BlackboardArtifact sourcebba = createDownloadSourceArtifact(downloadFile, zoneInfo);
```
raman-bt: same comment as above re naming.

rcordovano:
@kellykelly3, please note that this code no longer auto-merges.

PR #4604, opened 2019-03-11
4786: Handle incorrectly written HTML files.

raman-bt:


Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
```diff
@@ -56,6 +57,9 @@
      * 
      * @return The text content of the file.
      */
+    @NbBundle.Messages({
+       "HtmlViewer_file_error=This file is missing or unreadable.",
+    })
```
wschaeferB: The Core/src/org/sleuthkit/autopsy/contentviewers/Bundle.properties-MERGED file which is changed with the addition of the two messages in this file should also be included in the PR

PR #4606, opened 2019-03-13
4792 redo other occurrences viewer

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -754,57 +734,139 @@ private void populateTable(Node node) {
...
-    private void setColumnWidthToText(int columnIndex, String text) {
-        TableColumn column = otherCasesTable.getColumnModel().getColumn(columnIndex);
-        FontMetrics fontMetrics = otherCasesTable.getFontMetrics(otherCasesTable.getFont());
-        int stringWidth = fontMetrics.stringWidth(text);
-        column.setMinWidth(stringWidth + CELL_TEXT_WIDTH_PADDING);
+    private String makeDataSourceString(String caseUUID, String deviceId, String dataSourceName) {
+        return caseUUID + deviceId + dataSourceName;
+    }
+
+    private void updateCaseSelection() {
```
dannysmyda: Missing header doc

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
```diff
@@ -754,57 +734,139 @@ private void populateTable(Node node) {
...
+                    }
+                }
+            }
+        }
+        if (dataSourcesTable.getRowCount() > 0) {
+            dataSourcesTable.setRowSelectionInterval(0, 0);
+        }
+    }
+
+    private void updateDataSourceSelection() {
```
dannysmyda: Missing header doc

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
dannysmyda: Also could you consider a rename of the function? What it seems like its doing is updating the data sources list on a new case selection, which isn't immediately clear from the name updateCaseSelection()

Core/src/org/sleuthkit/autopsy/centralrepository/contentviewer/DataContentViewerOtherCases.java
dannysmyda: Same rename idea as above but for files and data sources.

PR #4609, opened 2019-03-14
1213 communications snapshot report

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -505,6 +522,16 @@ public void actionPerformed(ActionEvent evt) {
 
         jSeparator2.setOrientation(SwingConstants.VERTICAL);
 
+        snapshotButton.setIcon(new ImageIcon(getClass().getResource("/org/sleuthkit/autopsy/report/images/image.png"))); // NOI18N
+        snapshotButton.setText(NbBundle.getMessage(VisualizationPanel.class, "VisualizationPanel.snapshotButton.text_1")); // NOI18N
+        snapshotButton.addActionListener(new ActionListener() {
```
millmanorama: I would use a method reference here.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -648,6 +681,14 @@ private void clearVizButtonActionPerformed(ActionEvent evt) {//GEN-FIRST:event_c
         setCursor(Cursor.getDefaultCursor());
     }//GEN-LAST:event_clearVizButtonActionPerformed
 
+    private void snapshotButtonActionPerformed(ActionEvent evt) {//GEN-FIRST:event_snapshotButtonActionPerformed
+        try{
+            handleSnapshotEvent();
```
millmanorama: I would inline handleSnapshotEvent(), or alternatively break it and the related code out into another more focussed class.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -648,6 +681,14 @@ private void clearVizButtonActionPerformed(ActionEvent evt) {//GEN-FIRST:event_c
...
+        } catch (NoCurrentCaseException | IOException ex) {
+            // TODO something here
```
millmanorama: resolve this TODO please

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -674,7 +715,116 @@ private void fitGraph() {
...
+        final String defaultReportName = FileUtil.escapeFileName(currentCase.getDisplayName() + " " + new SimpleDateFormat("MMddyyyyHHmmss").format(generationDate)); //NON_NLS
+
+        final JTextField text = new JTextField(50);
+        final JPanel panel = new JPanel(new GridLayout(2, 1));
+        panel.add(new JLabel(Bundle.VisualizationPane_fileName_prompt()));
+        panel.add(text);
+
+        text.setText(defaultReportName);
+
+        int result = JOptionPane.showConfirmDialog(graphComponent, panel,
```
millmanorama: can you use JOptionPane.showInoutDialog() here, and eliminate the  GUI config code?

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -674,7 +715,116 @@ private void fitGraph() {
...
+                Bundle.VisualizationPanel_action_dialogs_title(), JOptionPane.OK_CANCEL_OPTION);
+
+        List<CommunicationsFilter.SubFilter> filters = currentFilter.getAndFilters();
```
millmanorama: filters is unused.  looks like it is left over from cut and paste.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -674,7 +715,116 @@ private void fitGraph() {
...
+
+        if (result == JOptionPane.OK_OPTION) {
+            String enteredReportName = text.getText();
+            String reportName = StringUtils.defaultIfBlank(enteredReportName, defaultReportName);
+            Path reportPath = Paths.get(currentCase.getReportDirectory(), reportName);
+            if (!Files.exists(reportPath)) {
```
millmanorama: I would switch the if-else and drop the !

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -674,7 +715,116 @@ private void fitGraph() {
...
+        "# {0} -  report path",
+        "VisualizationPane_Report_Success=Report Successfully create at:\n{0}",
+        "VisualizationPane_Report_OK_Button=OK",
+        "VisualizationPane_Open_Report=Open Report",})
+    private void createReport(Case currentCase, String reportName) throws IOException {
+        
+        // Create the report.
+        Path reportFolderPath = Paths.get(currentCase.getReportDirectory(), reportName, Bundle.VisualizationPane_reportName()); //NON_NLS
+        BufferedImage image = mxCellRenderer.createBufferedImage(graph, null, graph.getView().getScale(), Color.WHITE, true, null);
+        Path reportPath = (new CommSnapShotReportWriter(currentCase, reportFolderPath, reportName, new Date(), image, currentFilter)).writeReport();
```
millmanorama: There is a confusing extra set of parentheses on line 791

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
```diff
@@ -674,7 +715,116 @@ private void fitGraph() {
...
+        
+        // Report success to the user and offer to open the report.
+        String message = Bundle.VisualizationPane_Report_Success(reportPath.toAbsolutePath());
+        String[] buttons = {Bundle.VisualizationPane_Report_OK_Button(),
```
millmanorama: i would swap the order of the buttons to match timeline

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
```diff
@@ -0,0 +1,126 @@
...
+        ImageIO.write(image, "png", getReportFolderPath().resolve("snapshot.png").toFile()); //NON-NLS
+
+        //make a map of context objects to resolve template paramaters against
+        HashMap<String, Object> snapShotContext = new HashMap<>();
+        snapShotContext.put("reportTitle", getReportName()); //NON-NLS
+
+        List<SubFilter> filters = filter.getAndFilters();
+
+        for (SubFilter f : filters) {
+            if (f.getClass().getName().equals(DateRangeFilter.class.getName())) {
```
millmanorama: I think using instanceOf is the usual idiom for this, or am I confused?

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
```diff
@@ -0,0 +1,126 @@
...
+
+                if (startDate > 0) {
+
+                    snapShotContext.put("startTime", formatter.format(new Date((Instant.ofEpochSecond(startDate)).toEpochMilli()))); //NON-NLS
+                }
+
+                if (endDate > 0) {
+                    snapShotContext.put("endTime", formatter.format(new Date((Instant.ofEpochSecond(endDate)).toEpochMilli()))); //NON-NLS
+                }
+            } else if (f.getClass().getName().equals(AccountTypeFilter.class.getName())) {
```
millmanorama: again, I think using instanceOf is the usual idiom for this.

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
```diff
@@ -0,0 +1,126 @@
...
+                        snapShotContext.put("twitterSelected", "checked"); //NON-NLS
+                    } else if (type == Account.Type.INSTAGRAM) {
+                        snapShotContext.put("instagramSelected", "checked"); //NON-NLS
+                    } else if (type == Account.Type.WHATSAPP) {
+                        snapShotContext.put("whatsAppSelected", "checked"); //NON-NLS
+                    } else if (type == Account.Type.WEBSITE) {
+                        snapShotContext.put("websiteSelected", "checked"); //NON-NLS
+                    }
+                }
+            }
```
millmanorama: what about the device filter?

Core/src/org/sleuthkit/autopsy/communications/snapshot/comm_snapshot_template.html
```diff
@@ -0,0 +1,27 @@
+<html>
+    <head>
+        <title>Communications Snapshot: {{reportTitle}}</title>
+        <link rel="stylesheet" type="text/css" href="index.css" />
+        <link rel="icon" type="image/ico" href="favicon.ico" />
+    </head>
+    <body><div id="header">Communications Snapshot</div>
+        <div id="content">
+            <img id="snapshot" src="snapshot.png" alt="Timeline Snapshot" >
```
millmanorama: the alt text needs to be chagnges

Core/src/org/sleuthkit/autopsy/report/uisnapshot/UiSnapShotReportWriter.java
```diff
@@ -0,0 +1,226 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.HashMap;
+import org.apache.commons.lang3.StringUtils;
+import org.sleuthkit.autopsy.casemodule.Case;
+import org.sleuthkit.autopsy.ingest.IngestManager;
+import org.sleuthkit.autopsy.report.ReportBranding;
+
+/**
+ * Generate and write the Timeline snapshot report to disk.
```
millmanorama: Timeline is left over in comment from before refactoring

Core/src/org/sleuthkit/autopsy/report/uisnapshot/UiSnapShotReportWriter.java
```diff
@@ -0,0 +1,226 @@
...
+            // use default Autopsy icon if custom icon is not set
+            try (InputStream faviconStream = UiSnapShotReportWriter.class.getResourceAsStream("/org/sleuthkit/autopsy/report/images/favicon.ico")) { //NON-NLS
+                Files.copy(faviconStream, reportFolderPath.resolve("favicon.ico")); //NON-NLS
+            }
+        } else {
+            Files.copy(Files.newInputStream(Paths.get(agencyLogoPath)), reportFolderPath.resolve("favicon.ico")); //NON-NLS           
+        }
+
+        //copy report summary icon
+        try (InputStream summaryStream = UiSnapShotReportWriter.class.getResourceAsStream("/org/sleuthkit/autopsy/report/images/summary.png")) { //NON-NLS
```
millmanorama: maybe we could make a helper method to avoid so much boilerplate/noise ?

kellykelly3:
1. I will play and see what it makes sense to do when the user leaves the field blank.  Maybe I will play with the wording a bit, I think anymore text in that dialog will make it clunky.  I will figure something out.
2. I can do that.
3. I went back and forth with the headers, I will just dump the word Filter all together and move everything over.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
kellykelly3: Generally I would too, but this code is generated from the ide and is consistent with the rest of the code.  I am reluctant to change just this one.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
kellykelly3: There is no simple way to get both the value of a text field and determine which of the buttons was clicked.  If you want the behavior of assuming that empty means default instead of cancel this is the simplest way to achieve that functionality.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
kellykelly3: I think they add extra clarity by grouping the creations of the object from the call. 

Like 2*4+4 is fine, but (2*4)+4 is a little easier read.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
kellykelly3: I originally had it that way, but it adds a bit of confusion below because they when the user clicks OK the value of result would be equal to NO_OPTION.  I went back and forth on this.  I kind of felt that they already looked so different that it didn't really matter.

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
kellykelly3: instanceOf is very expensive, I believe this is the recommended way to do this.

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
kellykelly3: We talked yesterday at the standup and Brian didn't think the report needed anything more than the picture.  At this point i had already done the other filters so it was decided to leave as is.  Also the device is in the picture.

Core/src/org/sleuthkit/autopsy/communications/VisualizationPanel.java
millmanorama: ok, makes sense

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
millmanorama: If Brian is cool with it, then so be it.

kellykelly3: As a result of a conversation with Richard, I have sent a message to both Richard and Brian to confirm whether or not the table is even needed.

millmanorama: My two cents is that is is necessary to understand the image.  Otherwise it could easily be missinterpreted to show the abscence of communications/accounts, when really they are just filtered out.  Maybe Brian figures that context can be provided by analysts in other ways, but I feel negligent to not include it in the report.  Users can always use just the image anyways...

rcordovano:
@kellykelly3, please note that this PR has merge conflicts that must be resolved before I can accept it, thanks!

Core/src/org/sleuthkit/autopsy/communications/snapshot/CommSnapShotReportWriter.java
millmanorama: This is still weird to me.  My quick research indicated that instanceOf is quite fast, and that comparing by class name is potentially dangerous (although probably not in our use case).  I wouldn't worry about performance here anyways cause it is not a huge list or we are checking or something that is done very often.  @rcordovano do you have an opinion

PR #4614, opened 2019-03-15
1216 drop immutablesets

millmanorama:
remove usage of ImmutableSets and defensive copies to improve performance


Core/src/org/sleuthkit/autopsy/timeline/ui/detailview/datamodel/EventCluster.java
```diff
@@ -100,9 +99,9 @@ public static EventCluster merge(EventCluster cluster1, EventCluster cluster2) {
 
         Interval spanningInterval = IntervalUtils.span(cluster1.span, cluster2.span);
 
-        Set<Long> idsUnion = union(cluster1.getEventIDs(), cluster2.getEventIDs()).immutableCopy();
-        Set<Long> hashHitsUnion = union(cluster1.getEventIDsWithHashHits(), cluster2.getEventIDsWithHashHits()).immutableCopy();
-        Set<Long> taggedUnion = union(cluster1.getEventIDsWithTags(), cluster2.getEventIDsWithTags()).immutableCopy();
+        Set<Long> idsUnion = Sets.union(cluster2.getEventIDs(), cluster2.getEventIDs());
+        Set<Long> hashHitsUnion = Sets.union(cluster2.getEventIDsWithHashHits(), cluster2.getEventIDsWithHashHits());
+        Set<Long> taggedUnion = Sets.union(cluster2.getEventIDsWithTags(), cluster2.getEventIDsWithTags());
```
rcordovano: This does not look right. We appear to be computing the union of a set with itself. I suspect the change of "cluster1" to "cluster2" for the first argument to the union calls is incorrect.

PR #4626, opened 2019-03-20
1215 use simpler query from sleuthkit to determine datasources for the fil…

millmanorama:
…ters.


Core/src/org/sleuthkit/autopsy/timeline/FilteredEventsModel.java
```diff
@@ -340,14 +342,14 @@ synchronized public EventTypeZoomLevel getEventTypeZoom() {
         return getZoomState().getTypeZoomLevel();
     }
 
-    /**
+    /** S
```
kellykelly3: Missing description

PR #4639, opened 2019-03-21
added comments, renamed. No logic changes except moving of an 'if'

bcarrier:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -388,104 +394,113 @@ void getCache(AbstractFile indexAbstractFile) {
         
         // Only process the first payload data segment in each entry
         //  first data segement has the HTTP headers, 2nd is the payload
-        for (int j = 1; j < dataEntries.size() && j < 2; j++) {
-            CacheData data = dataEntries.get(j);
-            String dataFilename = data.getAddress().getFilename();
-            Optional<AbstractFile> dataFile = this.findCacheFile(dataFilename, cachePath);
+        CacheData dataSegment = dataEntries.get(1);
```
raman-bt: I'm getting an IndexOutOfBoundsException at this line on my test data. 

I think it's possible some entries may have no payload at all ?

raman-bt: This check before line 397 does the trick:
` if (dataEntries.size() < 2) {
            return derivedFiles;
        }`

bcarrier: Good catch. Just fixed that.

PR #4647, opened 2019-03-22
4578 Improved case deletion

rcordovano:
- There is new "Cases" tab as a sibling to the "Auto Ingest Jobs" tab and the "Auto Ingest Nodes" tab of the cluster administration dashboard.
- All three requested use cases have been implemented:
--Delete the auto ingest job manifests and corresponding data sources, while leaving the manifest file coordination service nodes and the rest of the case intact. The use case is freeing auto ingest input directory space while retaining the option to restore the data sources, effectively restoring the case.
-- Delete the manifest file coordination service nodes and the output for a case, while leaving the auto ingest job manifests and corresponding data sources intact. The use case is auto ingest reprocessing of a case with a clean slate without having to restore the manifests and data sources.
-- Delete everything.
-- Manifest file paths are added to a auto-ingest-job-manifests.txt file in the case directory root as the auto ingest jobs are executed. This file is used to guide deleting of case input; however there is support for legacy cases in the form of sifting through the complete set of manifest file znodes and matching on case name.
- There is a dedicated auto ingest dashboard log that is used by the "Cases" tab. This is modeled on the auto ingest system log as a separate and verbose record of case deletion.
- The "Cases" tab is a tabular NetBeans view with "type to find", sortable /hideable columns, and context menu items.
-- Columns include case name, create date, last access date, case directory, and columns to indicate whether or not the following case components have been fully deleted: manifests and data sources, manifest file znodes, case database, text index and Solr core.properties file, and case directory.
-- Context menu actions are open case, open case auto ingest log, delete case input, delete case out put, delete case input and output, and (display node) properties.
-- Multi-select is supported.
- Deletion happens in a background thread pool with four threads.
- Progress is reported via the progress bar in the lower right hand corner of the main application window. 
- Cancellation is supported. Whether or not it is advisable depends on circumstances. Interrupting case directory deletion is, unfortunately, not currently supported, although it could and perhaps should be, given that this is the lengthiest part of case deletion by a huge margin.
- A race condition between case deletion and multiple input directory scanners has been eliminated. Another race condition technically remains, since the scanners read manifest files before they obtain locks. I have mitigated this by making scanners check for manifest file deletion after they obtain a lock and by having the case deletion code do a few retries when deleting manifest files. To really eliminate this, we would need to implement an input directory locking protocol.


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/AutoIngestJobsNode.java
```diff
@@ -35,6 +36,13 @@
 import org.openide.util.NbBundle.Messages;
 import org.sleuthkit.autopsy.datamodel.NodeProperty;
 import org.sleuthkit.autopsy.experimental.autoingest.AutoIngestJob.Stage;
+import org.sleuthkit.autopsy.experimental.autoingest.AutoIngestAdminActions;
+import org.sleuthkit.autopsy.experimental.autoingest.AutoIngestDashboard;
+import org.sleuthkit.autopsy.experimental.autoingest.AutoIngestMonitor;
+import org.sleuthkit.autopsy.experimental.autoingest.AutoIngestNodeRefreshEvents;
+import org.sleuthkit.autopsy.experimental.autoingest.Bundle;
+import org.sleuthkit.autopsy.experimental.autoingest.Manifest;
+import org.sleuthkit.autopsy.experimental.autoingest.PrioritizationAction;
```
wschaeferB: was looking around in the code for the name of the _aiaa file and noticed these added imports appear to be unnecessary as they are all from the same package.

APriestman: I believe codacy calls those out

rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+            }
+            setDeletedItemFlag(caseNodeData, CaseNodeData.DeletedFlags.CASE_DB);
+        }
+    }
+
+    /**
+     * Attempts to delete the text index for a multi-user case.
+     *
+     * @param caseNodeData      The coordination service node data for the case.
+     * @param metadata          The case mnetadata.
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+     * @param metadata          The case metadata.
+     * @param progressIndicator A progress indicator.
+     * @param logger            A logger.
+     *
+     * @return True if one or more errors occurred (see log for details), false
+     *         otherwise.
+     *
+     * @throws InterruptedException If the thread this code is running in is
+     *                              interrupted while blocked, i.e., if
+     *                              cancellation of the opersation is detected
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+     * @param metadata          The case metadata.
+     * @param progressIndicator A progress indicator.
+     *
+     * @throws CaseActionException  If there were one or more errors deleting
+     *                              the case. The exception will have a
+     *                              user-friendly message and may be a wrapper
+     *                              for a lower-level exception.
+     * @throws InterruptedException If the thread this code is running in is
+     *                              interrupted while blocked, i.e., if
+     *                              cancellation of the opersation is detected
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+            logger.log(Level.INFO, String.format("Deleting text index for %s", caseNodeData.getDisplayName(), caseNodeData.getName(), caseNodeData.getDirectory())); //NON-NLS
+            deleteTextIndex(metadata, progressIndicator);
+            setDeletedItemFlag(caseNodeData, CaseNodeData.DeletedFlags.TEXT_INDEX);
+        }
+    }
+
+    /**
+     * Attempts to delete the text index for a case.
+     *
+     * @param metadata          The case mnetadata.
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+        for (KeywordSearchService searchService : Lookup.getDefault().lookupAll(KeywordSearchService.class)) {
+            searchService.deleteTextIndex(metadata);
+        }
+    }
+
+    /**
+     * Attempts to delete the case directory for a multi-user case.
+     *
+     * @param caseNodeData      The coordination service node data for the case.
+     * @param metadata          The case mnetadata.
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+            logger.log(Level.INFO, String.format("Deleting case directory for %s", caseNodeData.getDisplayName(), caseNodeData.getName(), caseNodeData.getDirectory())); //NON-NLS
+            deleteCaseDirectory(metadata, progressIndicator);
+            setDeletedItemFlag(caseNodeData, CaseNodeData.DeletedFlags.CASE_DIR);
+        }
+    }
+
+    /**
+     * Attempts to delete the case directory for a case.
+     *
+     * @param metadata          The case mnetadata.
```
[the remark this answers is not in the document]
rcordovano: Thanks, fixed in next commit.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2688,6 +2531,379 @@ private String getOrCreateSubdirectory(String subDirectoryName) {
...
+     * Attempts to remove a case from the recent cases menu if the main
+     * application window is present.
+     *
+     * @param metadata          The case metadata.
+     * @param progressIndicator A progress indicator.
+     */
+    @Messages({
+        "Case.progressMessage.removingCaseFromRecentCases=Removing case from Recent Cases menu..."
+    })
+    private static void deleteFromRecentCases(CaseMetadata metadata, ProgressIndicator progressIndicator) {
```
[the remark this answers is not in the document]
rcordovano: Yes, good catch. Thanks. Fixed in next commit.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+                     */
+                    progress.progress(Bundle.DeleteCaseTask_progress_acquiringManifestLocks());
+                    logger.log(Level.INFO, String.format("Acquiring exclusive manifest file locks for %s", caseNodeData.getDisplayName()));
+                    try {
+                        if (!acquireManifestFileLocks()) {
+                            logger.log(Level.INFO, String.format("Could not delete %s because at least one manifest file lock was already held by another host", caseNodeData.getDisplayName()));
+                            return;
+                        }
+                    } catch (IOException | CoordinationServiceException ex) {
+                        logger.log(Level.SEVERE, String.format("Could not delete %s because an error occurred acquiring the manifest file locks", caseNodeData.getDisplayName()), ex);
```
APriestman: Do the manifest locks need to be released here?

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+                        return;
+                    } catch (InterruptedException ex) {
+                        logger.log(Level.WARNING, String.format("Deletion of %s cancelled while incomplete", caseNodeData.getDisplayName()), ex);
```
APriestman: Do the manifest locks need to be released here?

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+                    logger.log(Level.WARNING, String.format("Deletion of %s cancelled while incomplete", caseNodeData.getDisplayName()));
+                    releaseManifestFileLocks();
+                    return;
+                }
+
+                if (deleteOption == DeleteOptions.DELETE_OUTPUT || deleteOption == DeleteOptions.DELETE_ALL) {
+                    try {
+                        deleteManifestFileNodes();
+                    } catch (InterruptedException ex) {
+                        logger.log(Level.WARNING, String.format("Deletion of %s cancelled while incomplete", caseNodeData.getDisplayName()), ex);
```
APriestman: Should the manifest file locks be released here?

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+                        return;
+                    }
+                }
+
+                releaseManifestFileLocks();
+
+            } catch (CoordinationServiceException ex) {
+                logger.log(Level.SEVERE, String.format("Could not delete %s because an error occurred acquiring the case directory lock", caseNodeData.getDisplayName()), ex);
+                return;
```
APriestman: Should the manifest file locks be released here?

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+            }
+
+            if (Thread.currentThread().isInterrupted()) {
+                logger.log(Level.WARNING, String.format("Deletion of %s cancelled while incomplete", caseNodeData.getDisplayName()));
```
APriestman: Should the manifest file locks be released here?

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+                            try {
+                                manifest = parser.parse(manifestFilePath);
+                                break;
+                            } catch (ManifestFileParser.ManifestFileParserException ex) {
+                                logger.log(Level.WARNING, String.format("Error parsing manifest file %s for %s", manifestFilePath, caseNodeData.getDisplayName()), ex);
+                            }
+                        }
+                    }
+                    if (manifest != null) {
+                        if (deleteDataSources(manifest, dataSources)) {
```
APriestman: I believe this should have an else that sets allInputDeleted to false

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -0,0 +1,850 @@
...
+     *                    database.
+     *
+     * @return True if all of the data source files werre deleted, false
+     *         otherwise.
+     */
+    @NbBundle.Messages({
+        "# {0} - data source name", "# {1} - device id", "DeleteCaseTask.progress.deletingDataSource=Deleting data source {0} with device id {1}...",})
+    private boolean deleteDataSources(Manifest manifest, List<DataSource> dataSources) {
+        final String dataSourceFileName = manifest.getDataSourceFileName();
+        final String dataSourceDeviceId = manifest.getDeviceId();
```
APriestman: Isn't the device ID optional in the manifest file?

wschaeferB: It looks like while DeviceID is optional for the manifest file, it is required to construct the Manifest object so we I think its probably generated after reading the file before constructing the object

APriestman: Yes just confirmed that if there is no devId field in the manifest it will auto-generate one.

APriestman: Continuing discussion in JIRA-4838

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
rcordovano: See discussion in Jira.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
rcordovano: See discussion in Jira.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
rcordovano: See discussion in Jira.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
rcordovano: See discussion in Jira.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
rcordovano: Thanks, corrected flag setting and removal of directory input deletion are in the next commit.

PR #4658, opened 2019-03-26
4860 gstreamer 0 length duration

dannysmyda:
This PR fixes the 00:00:00 duration bug that is present on all video and audio files when they are first clicked on. It also contains general code clean up that prepares it for follow on stories. A side effect of the refactor also fixes JIRA-4861 and an unreported NPE. Some parameters and imports were removed because either a) They are no longer needed or b) were not being used in the first place. Some of these diffs are due to formatting.


rcordovano:
@dannysmyda, there are three Codacy issues that sound like they could be addressed. Thanks!

Core/src/org/sleuthkit/autopsy/contentviewers/MediaViewImagePanel.java
```diff
@@ -172,9 +172,8 @@ private void showErrorNode(String errorMessage, AbstractFile file) {
      * Show the contents of the given AbstractFile as a visual image.
      *
      * @param file image file to show
-     * @param dims dimension of the parent window (ignored)
      */
-    void showImageFx(final AbstractFile file, final Dimension dims) {
```
wschaeferB: removing this unnecessary parameter makes the import for java.awt.Dimension also unnecessary.

Core/src/org/sleuthkit/autopsy/contentviewers/MediaFileViewer.java
```diff
@@ -120,14 +120,11 @@ public void setFile(AbstractFile file) {
             }
 
             lastFile = file;
-
-            final Dimension dims = MediaFileViewer.this.getSize();
```
wschaeferB: removing this unnecessary parameter makes the import for java.awt.Dimension also unnecessary.

dannysmyda:
This commit fixed the bug reported by Will. In this we also fix the Panel and Form files so that they can be manipulated in Netbeans. An audio bar has now been added. State changes have been handled correctly to (hopefully) completely remove the possibility for deadlock. Preroll frame management has been improved. The duration and position off by one bugs have been fixed. Audio bar getting stuck has been fixed. Codacy fixes have been addressed in previous commit. New commits for new fixes will be updated accordingly.

dannysmyda:
@wschaeferB I realize now that the application viewer disappears and doesn't display the gstreamer not initialized message when it is not on the path (since I decided to make the init of MediaPlayer throw an exception rather than maintain state that it was not initialized). I'll fix this on Monday. It should not affect the meaningful testing portion of this PR.

wschaeferB:
There appears to be a resizing issue that did not exist in the current release candidate.  

That is if you resize the height of the content viewers to be smaller while viewing a picture, and then switch to viewing a video, it attempts to display the content of the video content viewer at a size that would be appropriate for it prior to the resize.  I think the visibility of this issue is related to us now displaying a video preview.  See jira for steps to replicate

PR #4663, opened 2019-03-27
4886 Make Case.updateCaseNodeData more robust, update MERGED file

rcordovano:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2187,15 +2187,29 @@ private void createCaseNodeData(ProgressIndicator progressIndicator) throws Case
     private void updateCaseNodeData(ProgressIndicator progressIndicator) throws CaseActionException {
```
raman-bt: Could add param for ProgressIndicator  to the header.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -2187,15 +2187,29 @@ private void createCaseNodeData(ProgressIndicator progressIndicator) throws Case
...
-                } catch (CoordinationServiceException | InterruptedException | IOException ex) {
-                    throw new CaseActionException(Bundle.Case_exceptionMessage_couldNotUpdateCaseNodeData(ex.getLocalizedMessage()), ex);
+                } else {
+                    /*
+                     * This is a "legacy" case with no data stored in its case
+                     * directory coordination service node yet, or the node is
+                     * empty due to some error, so create the coordination
+                     * service node data from the case metadata.
+                     */
+                    nodeData = new CaseNodeData(metadata);
```
raman-bt: Does the LastAccessDate need to be set for nodeData here?

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
rcordovano: @raman-bt, thanks I definitely need to do that.

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
rcordovano: @raman-bt, the CaseNodeData constructor that takes a CaseMetaData object to make node data from scratch sets the last accessed date.

PR #4667, opened 2019-03-29
1150 limit plaso to 2 workers

millmanorama:


millmanorama:
@kellykelly3 you can test this by invoking an ingest with the plaso module selected and verifying that there are not more than 3 ( 2 workers plus 1 message queue) log2timeline.exe processes.

rcordovano:
@millmanorama, please address merge conflicts, thanks!

PR #4687, opened 2019-04-05
4893 data src delete fix

rcordovano:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteCaseTask.java
```diff
@@ -606,26 +606,47 @@ private boolean deleteDataSources(Manifest manifest, List<DataSource> dataSource
          * set, report file, archive file, etc.). In this case, just the file
          * referenced by the manifest will be deleted.
          */
-        boolean allFilesDeleted = true;
         Set<Path> filesToDelete = new HashSet<>();
-        for (DataSource dataSource : dataSources) {
+        int index = 0;
```
APriestman: Does index ever get updated?

rcordovano: No, the increment was omitted and testing on just the XP case in the small hours of the night was a bad idea! :)

Good catch, thanks - fixed.

PR #4696, opened 2019-04-09
1228 communications poc

kellykelly3:
A Couple of notes:
- MessageNode & MessageFactory are in this PR, but are unused, they are skeletons for the next steps
- For POC, the RelationshipBrowser only appears in the AccountBrowser. 
- The MessageBrowser object is still created in the AccountBrowser.  The MessageBrowser object is passed on line 105 to the constructor for the ProxyLookup, I haven't dug into what that is doing yet so I thought it was for the best to leave it.


Core/src/org/sleuthkit/autopsy/communications/RelaionshipSetNodeFactory.java
```diff
@@ -1,7 +1,20 @@
 /*
- * To change this license header, choose License Headers in Project Properties.
- * To change this template file, choose Tools | Templates
- * and open the template in the editor.
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2017-2019 Basis Technology Corp.
+ * Contact: carrier <at> sleuthkit <dot> org
+ *
```
wschaeferB: This class may be deleted.

Core/src/org/sleuthkit/autopsy/communications/AccountsBrowser.java
```diff
@@ -31,11 +34,13 @@
 import org.openide.explorer.ExplorerUtils;
 import org.openide.nodes.AbstractNode;
 import org.openide.nodes.Children;
+import org.openide.nodes.Node;
 import org.openide.util.Lookup;
 import org.openide.util.lookup.ProxyLookup;
 import org.sleuthkit.autopsy.casemodule.Case;
 import org.sleuthkit.autopsy.casemodule.NoCurrentCaseException;
 import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.datamodel.AccountDeviceInstance;
```
wschaeferB: The new imports appear to be unused.

Core/src/org/sleuthkit/autopsy/communications/MessageNode.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ *
+ * 
+ */
+final class MessageNode extends BlackboardArtifactNode {
+    private static final Logger logger = Logger.getLogger(RelationshipNode.class.getName());
+    
+    MessageNode(BlackboardArtifact artifact) {
+        super(artifact);
+        // Grabbed this from RelationshipNode, does always work even with internalization?
```
wschaeferB: I don't know the answer to this question.  But the comment asking a question should be removed / replaced before this code goes into develop / release.  As this PR is an intermediate step on its own branch it is probably fine for now.

Core/src/org/sleuthkit/autopsy/communications/AccountsBrowser.java
kellykelly3: Got it in the next PR.

Core/src/org/sleuthkit/autopsy/communications/MessageNode.java
kellykelly3: I removed it.  I will ask the question in a jira.

PR #4704, opened 2019-04-10
1229 selectioninfo implementation

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/MessageNode.java
```diff
@@ -18,23 +18,147 @@
...
+        }
+
+        sheetSet.put(new NodeProperty<>("Type", "Type", "Type", getDisplayName())); //NON-NLS
+        
+        BlackboardArtifact.ARTIFACT_TYPE fromID = BlackboardArtifact.ARTIFACT_TYPE.fromID(artifact.getArtifactTypeID());
+        if (null != fromID) {
+            //Consider refactoring this to reduce boilerplate
+            switch (fromID) {
+                case TSK_EMAIL_MSG:
+                    sheetSet.put(new NodeProperty<>("From", "From", "From",
```
wschaeferB: These NodeProperty Strings should probably be declared using @Messages so that they can be localized when necessary since they are user facing column names.

kellykelly3: Of the three strings you supply to NodeProperty, the first string is an identifier that should not be internationalized, the second some is a display name for the property\column which I have moved the to messages, the third string is a description that we do not show any where so I just pass an empty string.

PR #4722, opened 2019-04-16
Tl filter fixes

millmanorama:
this is part of the fixes for an unreported bug where adding filters resulted in duplicated filters in the GUI

please take https://github.com/sleuthkit/autopsy/pull/4721 first


rcordovano:
@millmanorama, please merge in the main line to reduce diffs, thanks.

Core/src/org/sleuthkit/autopsy/timeline/ui/filtering/datamodel/RootFilterState.java
```diff
@@ -260,17 +249,32 @@ public boolean isSelected() {
...
     @Override
-    public ReadOnlyBooleanProperty selectedProperty() {
+    public BooleanProperty selectedProperty() {
         return ALWAYS_TRUE;
     }
 
     @Override
     public void setDisabled(Boolean act) {
-        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
+//        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
kellykelly3: For both setDisabled and setSelected, I would removed the commented out line and leave a comment about why you are overriding the functions, but they do nothing.

PR #4723, opened 2019-04-16
1225 reduce list view memory

millmanorama:
please take https://github.com/sleuthkit/autopsy/pull/4721 first


rcordovano:
@millmanorama, please merge in the main line to reduce diffs, thanks.

PR #4732, opened 2019-04-17
Update NEWS for release 4.11.0

rcordovano:


markmckinnon:
Check the first line of the file is says 4.10.0 instead of 4.11.0 for version.

PR #4743, opened 2019-04-20
Add Initial Version of README_Linux.md

bghost:
With the recent change in Oracle Java SE Licensings (as of April 16, 2019), it
seems useful to pursue alternatives to satisfy the Java dependency.
Additionally, it is difficult to find comprehensive documentation about
installing and using Autopsy on Linux. This is an initial attempt to address
these issues.


gdicristofaro:
Hello @djds ,

We recently updated our mac/linux build instructions [here](https://github.com/sleuthkit/autopsy/blob/develop/Running_Linux_OSX.md), and it seems that the landscape for the jdk has changed since this PR was made.  Is there anything is this PR that you feel we should look into more for integration?  Thanks,

Greg

PR #4749, opened 2019-04-24
1232 thumbnail view

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
```diff
@@ -77,18 +78,61 @@ protected Sheet createSheet() {
...
-        sheetSet.put(new NodeProperty<>("url", Bundle.ContactNode_URL(), "",
-                getAttributeDisplayString(artifact, TSK_URL))); //NON-NLS
+        // Sorting the attributes by type so that the duplicates can be removed
+        // and they can be grouped by type for display.
+        try {
+            HashMap<String, BlackboardAttribute> phoneNumList = new HashMap<>();
+            HashMap<String, BlackboardAttribute> emailList = new HashMap<>();
+            HashMap<String, BlackboardAttribute> nameList = new HashMap<>();
+            HashMap<String, BlackboardAttribute> otherList = new HashMap<>();
+            for (BlackboardAttribute bba : artifact.getAttributes()) {
```
esaunders: I'm guessing the use of "contains" against the attribute type string is so that we catch all TSK_PHONE_*, TSK_EMAIL_* etc. attributes. If so, it might be worth a comment as it threw me for a loop and I had to go digging to figure it out.
As an aside, does this open us up to the possibility that this might break if a new attribute type is added that starts with one of these strings e.g. "TSK_NAME_PLACE"?

Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
```diff
@@ -77,18 +78,61 @@ protected Sheet createSheet() {
...
+                    emailList.put(bba.getDisplayString(), bba);
+                } else if (bba.getAttributeType().getTypeName().contains("TSK_NAME")) {
+                    nameList.put(bba.getDisplayString(), bba);
+                } else {
+                    otherList.put(bba.getDisplayString(), bba);
+                }
+            }
+            String propertyID = BlackboardAttribute.ATTRIBUTE_TYPE.TSK_NAME.getLabel();
+            int count = 0;
+            for (BlackboardAttribute bba : nameList.values()) {
```
esaunders: The following 3 "for" loops all look eerily similar differing only in the propertyID used and the collection they operate on. Perhaps these can be consolidated into a single method that does the work (taking the propertyID and the collection as parameters)?

Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
kellykelly3: I can add to the existing comment.  Would the code be clearer if I used startsWith instead of contains?  

Why do you believe that TSK_NAME_PLACE would break the existing code?  It would be added to the list of contact names.   Depending on the meaning of TSK_NAME_PLACE in the context of a contact you might not get it placed it in the best location of the list, but nothing would be "broken".

Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
kellykelly3: I thought about that.  I guess it would make it a bit cleaner.

Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
esaunders: I don't think using startsWith would make a whole lot of difference.

As you point out, it would depend on the meaning of the new fictitious TSK_NAME_PLACE attribute but assuming it has nothing to do with people/relationships would it be correct for it to show up in the list? (Disclaimer: I'm not familiar enough with the feature or functionality to know what I'm talking about).
The alternative would be to compare the attributes to the specific list of types that we care about but perhaps this is too limiting and perhaps one reason for using the attribute name was to allow for the possibility that third party modules could create new types of artifacts and attributes that  would then magically show up in the CVT. Again, see disclaimer above.

Core/src/org/sleuthkit/autopsy/communications/ThumbnailChildren.java
```diff
@@ -0,0 +1,127 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.Content;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Factory for creating thumbnail children nodes.
+ *
+ * Given the current way that DataResultViewerThumbnail works this class must
+ * extend Children.Keys not ChildNodeFactory. When a ChildNodeFactory is used
```
esaunders: Do you really mean ChildNodeFactory here or did you mean ChildFactory?

Core/src/org/sleuthkit/autopsy/communications/ContactNode.java
kellykelly3: The attribute TSK_NAME_PLACE would only show up if it was added as an attribute of the contact in which case it must have some meaning in relationship to the contact artifact therefore is valid to show.  Whether it shows up with the names or at the bottom of the list is not a big deal, the intent is the same which is to show all of the attributes of a contact.   

Having the code be dynamic means that we don't need to revisiting if new TSK_NAME_X, TSK_PHONE_NUMBER_X or TSK_EMAIL attributes are added for contacts.

rcordovano:
@wschaeferB, please confirm with @kellykelly3 that she is ready for your review/test before proceeding, thanks!

PR #4752, opened 2019-04-25
4978 text content viewer

wschaeferB:


Core/src/org/sleuthkit/autopsy/corecomponentinterfaces/TextViewer.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Interface that TextViewers for the TextContentViewer must implement. These
+ * modules analyze an individual file that the user has selected and display
+ * results in some form of JPanel.
+ *
+ * TextViewer panels should handle their own vertical scrolling, the horizontal
+ * scrolling when under their panel's preferred size will be handled by the
+ * TextContentViewerPanel which contains them.
+ */
+public interface TextViewer {
```
dannysmyda: Given that this implementation duplicates a lot of the DataContentViewer interface, is there a reason we cannot extend the DataContentViewer interface here?

wschaeferB: It is my understanding that TextViewer cannot extend DataContentViewer or vice versa because it will screw up the lookups, (i.e. looking up the parent class will get all instances of its child classes).

I could make, and did consider making, a common parent interface which both DataContentViewer and TextViewer extended. Neither extension would have anything in it at this point, because neither would add anything to that interface at this time. The only reason that I didn't do this is because I don't really know what the TextViewer interface will look like in the end based on 4945 being kinda up in the air I was hesitant to tie it so closely to the DataContentViewer interface at this point.

PR #4758, opened 2019-04-30
4975 pdf attachment extractor

dannysmyda:
This PR adds support for extracting attachments from PDFs in the EmbeddedFileExtractor and disables Tika's embedded content extraction for mime-types that are supported by the EmbeddedFileExtractor. This is necessary to not duplicate text in the Solr index.


rcordovano:
@dannysmyda, the two Codacy items look easily resolvable to me. The class with only static methods should have a private constructor, for example.

PR #4766, opened 2019-05-02
Orphan znode deletion fixes

rcordovano:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/DeleteOrphanManifestNodesTask.java
```diff
@@ -68,36 +71,45 @@ public void run() {
...
-                        logger.log(Level.WARNING, String.format("Task cancelled while deleting %s znode for %s", manifestFilePath, caseName));  // NON-NLS
+                        logger.log(Level.WARNING, String.format("Task cancelled while deleting orphaned manifest file %s", manifestFilePath));  // NON-NLS
                         return;
                     }
+
+                    /*
+                     * Back to looking for orphans...
+                     */
+                    progress.progress(Bundle.DeleteOrphanManifestNodesTask_progress_lookingForOrphanedManifestFileZnodes());
+                    logger.log(Level.INFO, Bundle.DeleteOrphanManifestNodesTask_progress_lookingForOrphanedManifestFileZnodes());
```
eugene7646: We should not hold up the point release for this but I think we should remove this log entry. It just pollutes the log to be printing this every loop iteration, potentially hundreds of times. Same for case node logging. Unless it's done for debugging purposes.

rcordovano: @eugene7646, the message is not being logged on every loop iteration.  It only happens as a reset in the conditional block after a node has been deleted. That said, the "reset" is really only needed for the progress bar, so that it does not look to the user like the task is stuck on deleting a single node when it is actually inspecting other nodes.

eugene7646: I see. I guess it does appear after a node has been deleted. I had almost a hundred nodes deleted in my test scenario so I had a hundred of these messages in the log and they didn't provide any value. It's no big deal though.

PR #4780, opened 2019-05-08
4911 update childfactory impls

esaunders:


Core/src/org/sleuthkit/autopsy/datamodel/DeletedContent.java
```diff
@@ -386,25 +391,19 @@ public void update(Observable o, Object arg) {
             }
 
             @Override
-            protected void addNotify() {
+            protected void onAdd() {
```
rcordovano: Do we need to use the Template Method design pattern here? The superclass addNotify method seems to do nothing other than call the onAdd method...or is there more coming in a later PR?

PR #4781, opened 2019-05-08
4991 text extractor metadata api

dannysmyda:


Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -486,20 +513,20 @@ private static int getTimeout(long size) {
      */
     @Override
     public void setExtractionSettings(Lookup context) {
-        if (context != null) {
+        if (nonNull(context)) {
```
esaunders: This is probably more a matter of style but I don't see the benefit to using Objects::nonNull() over "!= null".
I don't think it adds clarity (in fact I think it is less clear), there is the (small) overhead of a static method call each time and there are more characters to type. The Java documentation states that nonNull "exists to be used as a Predicate, filter(Objects::nonNull)". I don't think the goal was to replace != and == with isNull and nonNull.

dannysmyda: Good point, I didn't think about why 'Objects' class would even include this method. I originally added it to experiment with a different side of null check readability, but I agree it is less clear. I'll take it out.

PR #4784, opened 2019-05-08
5010 make logical imager dsp stub with config panel

jkho:


rcordovano:
@jkho and @APriestman, the AppVeyor failure is due to assorted feature test failures and is unrelated to the content of this PR.

PR #4792, opened 2019-05-10
4961 add registry content viewer

wschaeferB:
Making PR now to start working on what I expect to be many Codacy complaints


wschaeferB:
closing PR till I finish clean up since codacy seems to be somewhat satisfied

wschaeferB:
Should be ready to review.

Core/src/org/sleuthkit/autopsy/contentviewers/WindowsRegistryViewer.java
```diff
@@ -0,0 +1,149 @@
...
+        if (content.getSize() == 0) {
+            return;
+        }
+
+        byte[] data = new byte[(int) content.getSize()];
+
+        try {
+            content.read(data, 0x0, content.getSize());
+        } catch (TskCoreException ex) {
+            logger.log(Level.WARNING, "Failed to read file content.", ex);
```
eugene7646: Shouldn't we exit here?

Core/src/org/sleuthkit/autopsy/contentviewers/WindowsRegistryViewer.java
```diff
@@ -0,0 +1,149 @@
...
+
+        if (file.getNameExtension().toLowerCase().startsWith(LOG_FILE_EXTENSION)) {
+            return false;
+        }
+        byte[] header = new byte[0x4000];
+
+        try {
+            file.read(header, 0x0, Math.min(0x4000, file.getSize()));
+        } catch (TskCoreException ex) {
+            logger.log(Level.WARNING, "Failed to read file content", ex);
```
eugene7646: I guess same question, shouldn't we exit if there was an exception while trying to read data from the file? Is there a point in trying to continue and to pass incomplete/gibberish data to other classes?

Core/src/org/sleuthkit/autopsy/rejview/HexView.java
```diff
@@ -0,0 +1,304 @@
...
+                int endByte = e.getDot();
+
+                if (startByte > endByte) {
+                    int t = endByte;
+                    endByte = startByte;
+                    startByte = t;
+                }
+
+                // the number of line endings before the start,end points
+                int startRows = (startByte - (startByte % bytesPerLine)) / (3 * bytesPerLine);
```
eugene7646: We should probably make the "3" into a constant, since we are using it in multiple places.

Core/src/org/sleuthkit/autopsy/contentviewers/WindowsRegistryViewer.java
wschaeferB: Returning there sounds like a reasonable thing to do if the reading of the data fails. It wasn't being done in the original code I was integrating, but that is no indication it shouldn't be done. I will make this change and test to ensure it works, unless I hear otherwise.

Core/src/org/sleuthkit/autopsy/contentviewers/WindowsRegistryViewer.java
wschaeferB: Same answer, I have no idea if there is a point.

Core/src/org/sleuthkit/autopsy/rejview/HexView.java
wschaeferB: Will do.

Core/src/org/sleuthkit/autopsy/contentviewers/WindowsRegistryViewer.java
eugene7646: I don't really know what the right answer is so you should probably seek the opinion of @rcordovano.

Core/src/org/sleuthkit/autopsy/rejview/RegeditExeValueFormatter.java
```diff
@@ -0,0 +1,104 @@
...
+            case REG_SZ: // empty case - intentional fall-through
+            case REG_EXPAND_SZ: {
+
+                String valString = val.getAsString();
+                if (valString.length() == 0) {
+                    sb.append(Bundle.RegeditExeValueFormatter_valueNotSet_text());
+                } else {
+                    sb.append(valString);
+                }
+                if (sb.length() > 48) {
```
eugene7646: I think we should also make constants out of these, I would have no idea what 45 or 48 is if I had to debug the code.

wschaeferB: Alright, the constants might be a bit vague in name like "MAX_VALUE_STRING_LEN" or something like that but I agree even vaguely named constants would be easier to understand than the raw numbers.

eugene7646:
I have finished the review and testing. As soon as the minor code review things are taken care of - the PR is good to go.

PR #4795, opened 2019-05-13
Data source tree paging

esaunders:


Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -169,10 +169,12 @@ private void removeListeners() {
...
-                    if (parentsChildren instanceof ContentChildren) {
-                        ((ContentChildren) parentsChildren).refreshChildren();
-                        parentsChildren.getNodesCount();
-                    }
+
+// TODO: How will this work with ChildFactory approach?
+//                    if (parentsChildren instanceof ContentChildren) {                     
+//                        ((ContentChildren) parentsChildren).refreshChildren();
+//                        parentsChildren.getNodesCount();
+//                    }
```
rcordovano: I'm not sure what to make of this, is this addressed later?

Core/src/org/sleuthkit/autopsy/datamodel/ContentChildren.java
```diff
@@ -33,17 +33,15 @@
...
 class ContentChildren extends AbstractContentChildren<Content> {
 
     private static final Logger logger = Logger.getLogger(ContentChildren.class.getName());
-    //private static final int MAX_CHILD_COUNT = 1000000;
 
     private final Content parent;
 
     ContentChildren(Content parent) {
-        super(); //initialize lazy behavior
+        super("content_" + Long.toString(parent.getId()));
```
rcordovano: I wonder if we should use canonical class names or something else a little more "formal" for these key strings. It seems that whoever adds new Children classes in the future might need to survey all of the key names to see which have been taken...

Core/src/org/sleuthkit/autopsy/datamodel/CreateSleuthkitNodeVisitor.java
```diff
@@ -0,0 +1,102 @@
...
+import org.sleuthkit.datamodel.SlackFile;
+import org.sleuthkit.datamodel.SleuthkitItemVisitor;
+import org.sleuthkit.datamodel.SleuthkitVisitableItem;
+import org.sleuthkit.datamodel.VirtualDirectory;
+import org.sleuthkit.datamodel.Volume;
+
+/**
+ * Creates appropriate Node for each sub-class of Content
+ */
+public class CreateSleuthkitNodeVisitor extends SleuthkitItemVisitor.Default<AbstractContentNode<? extends Content>> {
```
rcordovano: Is it required that this class be public?

Core/src/org/sleuthkit/autopsy/datamodel/DataSourcesNode.java
```diff
@@ -97,25 +98,24 @@ public DataSourcesNodeChildren(long dsObjId) {
...
                 if (eventType.equals(Case.Events.DATA_SOURCE_ADDED.toString())) {
-                    reloadKeys();
+                    refresh(true);
                 }
             }
         };
 
         @Override
-        protected void addNotify() {
+        protected void onAdd() {
```
rcordovano: It looks like you found a use for the onAdd and onRemove, after all.

Core/src/org/sleuthkit/autopsy/directorytree/DataResultFilterNode.java
```diff
@@ -320,17 +320,17 @@ private DataResultFilterChildren(Node arg, ExplorerManager sourceEm, boolean fil
...
+//            if (file != null) {
+//                if (filterKnown && (file.getKnown() == TskData.FileKnown.KNOWN)) {
+//                    // Filter out child nodes that represent known files
+//                    return new Node[]{};
+//                }
+//                if (filterSlack && file.getType().equals(TskData.TSK_DB_FILES_TYPE_ENUM.SLACK)) {
+//                    // Filter out child nodes that represent slack files
+//                    return new Node[]{};
+//                }
+//            }
```
rcordovano: Is this replaced in the next PR?

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
esaunders: Yes, this is addressed later.

Core/src/org/sleuthkit/autopsy/datamodel/CreateSleuthkitNodeVisitor.java
esaunders: Yes, it is used by ViewAssociatedContentAction in the directorytree package and it was previously public in it's previous location.

Core/src/org/sleuthkit/autopsy/directorytree/DataResultFilterNode.java
esaunders: Yes.

Core/src/org/sleuthkit/autopsy/datamodel/ContentChildren.java
esaunders: Yeah, I had considered doing something different but went with this because DataResultViewerTable uses the node name (Node.getName()) to communicate with the ChildFactory and AbstractContentNode has always set the name in it's constructor to "content_" + id.

PR #4802, opened 2019-05-15
5011 logical imager dsp adds data

jkho:


Core/src/org/sleuthkit/autopsy/casemodule/LogicalImagerDSProcessor.java
```diff
@@ -114,14 +125,59 @@ public boolean isPanelValid() {
      * @param callback        Callback that will be used by the background task
      *                        to return results.
      */
+    @Messages({
+        "# {0} - imageDirPath", "LogicalImagerDSProcessor.imageDirPathNotFound={0} not found.\nUSB drive has been ejected.",
+        "# {0} - directory", "LogicalImagerDSProcessor.failToCreateDirectory=Fail to create directory {0}",
```
APriestman: Should be "Failed" I think

jkho: updated

Core/src/org/sleuthkit/autopsy/casemodule/AddLogicalImageTask.java
```diff
@@ -0,0 +1,139 @@
...
+        List<Content> emptyDataSources = new ArrayList<>();
+        
+        try {
+            progressMonitor.setProgressText(Bundle.AddLogicalImageTask_copyingImageFromTo(src.toString(), dest.toString()));
+            FileUtils.copyDirectory(src, dest);
+            progressMonitor.setProgressText(Bundle.AddLogicalImageTask_doneCopying());
+        } catch (IOException ex) {
+            // Copy directory failed
+            String msg = Bundle.AddLogicalImageTask_failedToCopyDirectory(src.toString(), dest.toString());
+            logger.log(Level.SEVERE, msg);
```
APriestman: Forgot to put this earlier. Our log messages should never be from the bundles - they should always just be a text string. This helps us diagnose problems through the logs. There are a couple of these to fix.

jkho: fixed

APriestman: I think it's still not quite right. The messages added to errorList are displayed to the user and should be bundle messages, but the log messages should be strings, so you'll have two very similar strings in the bundle message and in the log message.

jkho: I think we should not log the IOException from copyDirectory. There are all kinds of possible IOExceptions while copying: USB drive ejected, out of disk space etc. These are *not* program logic errors that Autopsy developer should be debug.

The user will get the error message in the wizard dialog box. 

The exception in addReport should be logged and reported to user. TskCoreException is from our code/logic so it is not obvious to the user what caused this exception.

I'll fix them again

rcordovano: No, we need to log all exceptions. I can tell you from long experience that it can be very difficult to get error messages from the UI from a user, while it is often possible for the user to simply zip up the log files and send them to us. Also, without the stack traces and exceptions messages we lose valuable information. The coding guideline for Autopsy development is that we never simply swallow exceptions and ALL exceptions are logged, with stack traces. 

Guidelines for logging levels:
INFO - provide context for other log messages and duration info. For example, logging the start and finish of ingest jobs.
WARNING - an exception occurred, but the integrity of the case is not in question. Example, a corrupted file was encountered.
SEVERE - an exception occurred that makes the case incomplete. Example, failure to insert a record in the case database.

jkho:
I fixed most of Codacy issues. The long argument list is inherited from AddImageTask constructor, not touching it because many other DSProcessors are calling it.

rcordovano:
Thanks, @jkho, we can accept the long argument list. You still need to provide the class header documentation, please.

jkho:
I added the class header document. What is wrong with it?

rcordovano:
Looking at the file itself to see why Codacy didn't pick it up, I see that you are missing a single asterisk on line 37. Comments we want doxygen to process to generate documentation need to start with "/**" instead of just "/*".

PR #4803, opened 2019-05-16
log exceptions.  Don't limit text sizes

bcarrier:


Core/src/org/sleuthkit/autopsy/texttranslation/ui/TranslatedTextViewer.java
```diff
@@ -287,7 +296,7 @@ private String getFileText(Node node) throws IOException,
 
             //Correct for UTF-8
             byte[] resultInUTF8Bytes = result.getBytes("UTF8");
-            byte[] trimTo1MB = Arrays.copyOfRange(resultInUTF8Bytes, 0, MAX_SIZE_1MB / 1000);
+            byte[] trimTo1MB = Arrays.copyOfRange(resultInUTF8Bytes, 0, MAX_SIZE_1MB );
```
wschaeferB: I have concerns about the length of time it takes to translate this much text, which I am voicing in jira 4808 as there isn't a story specifically for this PR.

PR #4810, opened 2019-05-21
2409-case-database-data-src-delete

markmckinnon:
Intial commit of the database src delete code


Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.swing.AbstractAction;
+import org.openide.util.NbBundle;
+import org.sleuthkit.autopsy.casemodule.Case;
+import org.sleuthkit.autopsy.casemodule.NoCurrentCaseException;
+import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ *
+ * @author markm
```
rcordovano: This class needs a class documentation header and we don't use "author."

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public final class DeleteDataSourceAction extends AbstractAction {
+    private static final Logger logger = Logger.getLogger(DeleteDataSourceAction.class.getName());
+    private final Long selectDataSource;
```
rcordovano: The selectDataSource instance variable can be a primitive long. Also, I think you change the name of the variable to selectedDataSource in a later PR. If I am wrong, please make the change.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+    
+    @NbBundle.Messages({"DeleteDataSourceAction.name.text=Delete Data Source"})
+    public DeleteDataSourceAction(Long selectedDataSource) {
+        super(Bundle.DeleteDataSourceAction_name_text());
+        selectDataSource = selectedDataSource;
+
+    }
+    @NbBundle.Messages({"ErrorDeletingDataSource.name.text=Error Deleting Data Source"})
+    @Override
```
rcordovano: There is some funky formatting here - empty lines in the constructor and no separation of the constructor and the method following it.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+    public void actionPerformed(ActionEvent event) {
+        try {
+            //VersionNumber checkVersionNumber = Case.getCurrentCaseThrows().getSleuthkitCase().getDBSchemaVersion();
```
rcordovano: This commented out code should be removed.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+            Case.getCurrentCaseThrows().getSleuthkitCase().deleteDataSource(selectDataSource);
+        } catch (NoCurrentCaseException | TskCoreException e) {
+	    String msg = MessageFormat.format(Bundle.ErrorDeletingDataSource_name_text(), selectDataSource);
+            logger.log(Level.WARNING, msg, e);
```
rcordovano: We don't localize exception messages.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,58 @@
...
+            //throw new TskCoreException(msg, e);
+        }
```
rcordovano: The commented out throw needs to be removed.

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
```diff
@@ -200,6 +204,27 @@ public String getItemType() {
...
+            ResultSet schemaVersion = query.getResultSet();
+            while (schemaVersion.next()) {
+                int creationMajorVersion = schemaVersion.getInt("creationMajorVersion");
+                int creationMinorVersion = schemaVersion.getInt("creationMinorVersion");
+                if ((creationMajorVersion == 8 && creationMinorVersion >= 3) || creationMajorVersion > 8) {
+                    schemaVersion.close();
+                    return true;
+                }
+            }
+            schemaVersion.close();
```
rcordovano: SleuthkitCase.CaseDbQuery takes care of closing the ResultSet.

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
```diff
@@ -200,6 +204,27 @@ public String getItemType() {
         return getClass().getName();
     }
 
+    private Boolean checkSchemaVersion() {
+        String sqlStatement = "SELECT a.value creationMajorVersion, b.value creationMinorVersion FROM tsk_db_info_extended a, tsk_db_info_extended b " +
+                              " WHERE a.name = 'CREATION_SCHEMA_MAJOR_VERSION' and b.name = 'CREATION_SCHEMA_MINOR_VERSION';";
+        try (CaseDbQuery query = Case.getCurrentCaseThrows().getSleuthkitCase().executeQuery(sqlStatement);) {
```
rcordovano: Perhaps the use of "AS" would make this SQL easier to read?

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
```diff
@@ -113,6 +114,9 @@ private void removeListeners() {
         actionsList.add(new RunIngestModulesAction(Collections.<Content>singletonList(content)));
         actionsList.add(new NewWindowViewAction(
                 NbBundle.getMessage(this.getClass(), "ImageNode.getActions.viewInNewWin.text"), this));
+        if (checkSchemaVersion()) {
```
dannysmyda: Call performed on EDT.

dannysmyda: Call does not need to be repeated.

rcordovano: @dannysmyda, I'm not sure there is a better alternative to making this query in the EDT. Suggestions? Also, which call is repeated?

dannysmyda: checkSchemaVersion is repeated for all right clicks. I think it should be done once before it is visible.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: fixed

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: removed

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: fixed

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: Fixed.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: Added comment and removed author.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: removed

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
markmckinnon: Removed close.

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
markmckinnon: Changed to add AS to select

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
markmckinnon: removed Select from imagenode.  Calls to sk public api to get the creation schema version now.

rcordovano: Danny's point is that getActions will be called on every ImageNode that is constructed and each query will do a database query in the EDT. The database schema version will not change from call to call to getDBCreationSchemaVersion for a given case. So we should make sure that getDBCreationSchemaVersion saves the result in a private instance variable to avoid the unnecessary database round trips.

rcordovano: Never mind, already done!

Core/src/org/sleuthkit/autopsy/datamodel/ImageNode.java
```diff
@@ -200,6 +204,22 @@ public String getItemType() {
         return getClass().getName();
     }
 
+    private Boolean checkSchemaVersion() {
+        try {
+            int creationMajorVersion = Case.getCurrentCaseThrows().getSleuthkitCase().getDBCreationSchemaVersion().getMajor();
+            int creationMinorVersion = Case.getCurrentCaseThrows().getSleuthkitCase().getDBCreationSchemaVersion().getMinor();
+
```
rcordovano: The SleuthkitCase method names on line 209 and 210 need to be getDBSchemaCreationVersion. Also, it would be a bit tidier to call getDBSchemaCreationVersion just once, in my opinion.

PR #4815, opened 2019-05-23
1252 lazy initialization of the date pickers

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/FiltersPanel.java
```diff
@@ -94,6 +98,8 @@
      * and it should be refreshed (by reapplying the filters).
      */
     private boolean needsRefresh;
+   
+    Case openCase = null;
```
wschaeferB: Can be private

PR #4819, opened 2019-05-26
2410 delete data sources from the case text index

markmckinnon:


Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sleuthkit.autopsy.keywordsearchservice.KeywordSearchService;
+import org.sleuthkit.autopsy.keywordsearchservice.KeywordSearchServiceException;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Instances of this Action allow users to delete the specified data source.
+ */
+public final class DeleteDataSourceAction extends AbstractAction {
+    private static final Logger logger = Logger.getLogger(DeleteDataSourceAction.class.getName());
+    private final Long selectedDataSource;
```
rcordovano: It's a small point, but from a readability point of view, I would prefer that we call this dataSourceId. When I saw the field name in the constructor, without the context of the data type, I figured it was a DataSource or Content object.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -48,5 +51,13 @@ public void actionPerformed(ActionEvent event) {
...
         }
     }
-
+    private static void deleteDataSource(Long dataSourceId) throws KeywordSearchServiceException {
+        try {
+            KeywordSearchService kwsService = Lookup.getDefault().lookup(KeywordSearchService.class);
+            kwsService.deleteDataSource(dataSourceId);
+        } catch (KeywordSearchServiceException e) {
+            logger.log(Level.WARNING, "KWS Error", e);
+        }
```
rcordovano: We may have already added this in a later PR, in which case this comment can be ignored...

If the action fails, we should give the user some immediate feedback with a popup from MessageUtil, in addition to logging the error. To this end, the static method should propagate the KeywordSearchServiceException (as it erroneously says it does) so that actionPerformed can take responsibility for comprehensive error handling. I see this as a micro application of the guideline from our coding standard that says that lower level code should leave error handling policy to top level code.

Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
```diff
@@ -97,4 +97,13 @@ default public void close() throws IOException {
          */
     }    
 
+    /**
+     * Deletes the keyword search text for a specific data source.
+     *
+     * @param dataSourceId The data source id to be deleted from Solr.
```
rcordovano: Technically, this interface knows nothing about Solr.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
```diff
@@ -1506,6 +1529,17 @@ private void commit() throws SolrServerException {
...
         }
 
+        private void deleteDataSource(Long dsObjId) {
+            String dataSourceId = Long.toString(dsObjId);
+            String deleteQuery = "image_id:" + dataSourceId;
+            try {
+                // Get the first result. 
+                UpdateResponse updateResponse = solrCore.deleteByQuery(deleteQuery);
+            } catch (SolrServerException | IOException ex) {
+                logger.log(Level.SEVERE, "Error deleting content from Solr. Solr image id " + dataSourceId, ex); //NON-NLS
```
rcordovano: As above for the DeleteDataSourceAction, low level classes should propagate exceptions for handling by the top level classes. E.g., here, how does the Server know what the performAction needs to do in its context of a menu selection in the UI by a user? These exceptions should propagate to the public deleteDataSource method above, which should propagate all errors, and so on, up the chain. The exceptions can be wrapped in other exceptions as needed.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
```diff
@@ -194,6 +194,28 @@ public void tryConnect(String host, int port) throws KeywordSearchServiceExcepti
...
+    @Override
+    public void deleteDataSource(Long dataSourceId) throws KeywordSearchServiceException {
+        try {
+            KeywordSearch.getServer().deleteDataSource(dataSourceId);
+        } catch (NoOpenCoreException ex) {
+            logger.log(Level.WARNING, NbBundle.getMessage(SolrSearchService.class,
+                "SolrSearchService.deleteDataSource.exceptionMessage.noCurrentSolrCore"));
+            throw new KeywordSearchServiceException(NbBundle.getMessage(SolrSearchService.class,
+                    "SolrSearchService.deleteDataSource.exceptionMessage.noCurrentSolrCore"));
+        }
```
rcordovano: We don't localize exception messages. The reason, for better or worse, is that the one who ultimately ends up reading them is often one of us... 

We don't need to actually log the exception until we get to the end of its propagation. It can be confusing to have duplicate log entries for the same error.

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
```diff
@@ -48,5 +51,13 @@ public void actionPerformed(ActionEvent event) {
             logger.log(Level.WARNING, "Error Deleting Data source " + selectedDataSource, e);
         }
     }
-
+    private static void deleteDataSource(Long dataSourceId) throws KeywordSearchServiceException {
```
rcordovano: This method is not called by performAction. Perhaps a merge issue?

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: Changed

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: This is in a later PR 2415 I have confirmed it.

Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
markmckinnon: Changed comment to remove Solr from it.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
markmckinnon: Better handle exceptions

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
markmckinnon: Put it back in.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
markmckinnon: Removed log message

Core/src/org/sleuthkit/autopsy/actions/DeleteDataSourceAction.java
rcordovano: Cool, I'm content to wait for the user notification.

However, we are still catching the KeywordSearchServiceException in DeleteDataSourceAction.deleteDataSource instead of propagating it to performAction as we should and the exception specification for DeleteDataSourceAction.deleteDataSource states. 

Feel free to refactor the code in DeleteDataSourceAction.deleteDataSource into DeleteDataSourceAction.performAction and delete the DeleteDataSourceAction.deleteDataSource method. Or propagate the exception by removing the try-catch in  DeleteDataSourceAction.deleteDataSource.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
rcordovano: Perhaps this PR is missing a commit? The logging is still present, and the the exception message is still in a resource bundle for localization.

rcordovano:
I am going to merge this now, then make any last changes I want in a feature branch of my own.

PR #4823, opened 2019-05-29
4963: offload computation of S C & O columns to a background task, in…

raman-bt:
… order to speed up the loading of Data Results View table.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
+++ b/Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
@@ -81,17 +81,6 @@ final class MessageNode extends BlackboardArtifactNode {
 
         sheetSet.put(new NodeProperty<>("Type", Bundle.MessageNode_Node_Property_Type(), "", getDisplayName())); //NON-NLS
 
-        addScoreProperty(sheetSet, tags);
-
-        CorrelationAttributeInstance correlationAttribute = null;
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
-            correlationAttribute = getCorrelationAttributeInstance();
-        }
-        addCommentProperty(sheetSet, tags, correlationAttribute);
-
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
-            addCountProperty(sheetSet, correlationAttribute);
-        }
         final BlackboardArtifact artifact = getArtifact();
 
         BlackboardArtifact.ARTIFACT_TYPE fromID = BlackboardArtifact.ARTIFACT_TYPE.fromID(artifact.getArtifactTypeID());
--- a/Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
+++ b/Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
@@ -18,13 +18,10 @@
  */
 package org.sleuthkit.autopsy.communications.relationships;
 
-import java.util.List;
 import java.util.TimeZone;
 import java.util.logging.Level;
 import org.apache.commons.lang3.StringUtils;
 import org.openide.nodes.Sheet;
-import org.sleuthkit.autopsy.centralrepository.datamodel.CorrelationAttributeInstance;
-import org.sleuthkit.autopsy.core.UserPreferences;
 import org.openide.util.NbBundle.Messages;
 import org.sleuthkit.autopsy.coreutils.Logger;
 import org.sleuthkit.autopsy.datamodel.BlackboardArtifactNode;
@@ -40,7 +37,6 @@ import static org.sleuthkit.datamodel.BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHO
 import static org.sleuthkit.datamodel.BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO;
 import static org.sleuthkit.datamodel.BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SUBJECT;
 import static org.sleuthkit.datamodel.BlackboardAttribute.TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.DATETIME;
-import org.sleuthkit.datamodel.Tag;
 import org.sleuthkit.datamodel.TimeUtilities;
 import org.sleuthkit.datamodel.TskCoreException;
 import org.sleuthkit.autopsy.communications.Utils;
@@ -72,7 +68,6 @@ final class MessageNode extends BlackboardArtifactNode {
     @Override
     protected Sheet createSheet() {
         Sheet sheet = super.createSheet();
-        List<Tag> tags = getAllTagsFromDatabase();
         Sheet.Set sheetSet = sheet.get(Sheet.PROPERTIES);
         if (sheetSet == null) {
             sheetSet = Sheet.createPropertiesSet();
--- a/Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
+++ b/Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
@@ -82,7 +82,8 @@ public abstract class AbstractAbstractFileNode<T extends AbstractFile> extends A
     private static final Set<Case.Events> CASE_EVENTS_OF_INTEREST = EnumSet.of(Case.Events.CURRENT_CASE,
             Case.Events.CONTENT_TAG_ADDED, Case.Events.CONTENT_TAG_DELETED, Case.Events.CR_COMMENT_CHANGED);
 
-    private static final ExecutorService translationPool;
+    // pool to run long running translation and getSCO tasks in backgound
+    private static final ExecutorService translationSCOPool;
     private static final Integer MAX_POOL_SIZE = 10;
 
     /**
@@ -101,7 +102,7 @@ public abstract class AbstractAbstractFileNode<T extends AbstractFile> extends A
         }
 
         if (UserPreferences.displayTranslatedFileNames()) {
-            AbstractAbstractFileNode.translationPool.submit(new TranslationTask(
+            AbstractAbstractFileNode.translationSCOPool.submit(new TranslationTask(
                     new WeakReference<>(this), weakPcl));
         }
 
@@ -113,8 +114,8 @@ public abstract class AbstractAbstractFileNode<T extends AbstractFile> extends A
     static {
         //Initialize this pool only once! This will be used by every instance of AAFN
         //to do their heavy duty SCO column and translation updates.
-        translationPool = Executors.newFixedThreadPool(MAX_POOL_SIZE,
-                new ThreadFactoryBuilder().setNameFormat("translation-task-thread-%d").build());
+        translationSCOPool = Executors.newFixedThreadPool(MAX_POOL_SIZE,
+                new ThreadFactoryBuilder().setNameFormat("translation-and-sco-task-thread-%d").build());
     }
 
     /**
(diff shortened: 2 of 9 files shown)


Core/src/org/sleuthkit/autopsy/datamodel/GetSCOTask.java
```diff
@@ -0,0 +1,71 @@
...
+    private final PropertyChangeListener listener;
+
+    public GetSCOTask(WeakReference<AbstractAbstractFileNode<?>> weakContentRef, PropertyChangeListener listener) {
+        this.weakNodeRef = weakContentRef;
+        this.listener = listener;
+    }
+
+    @Override
+    public void run() {
+        AbstractAbstractFileNode<?> fileNode = weakNodeRef.get();
```
wschaeferB: I think this will have trouble with the Results section of the tree and the BlackboardArtifactNodes

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -368,18 +327,16 @@ public String toString() {
...
-        properties.add(new NodeProperty<>(SCORE.toString(), SCORE.toString(), scoreAndDescription.getRight(), scoreAndDescription.getLeft()));
-        DataResultViewerTable.HasCommentStatus comment = getCommentProperty(tags, attribute);
-        properties.add(new NodeProperty<>(COMMENT.toString(), COMMENT.toString(), NO_DESCR, comment));
-        if (!UserPreferences.hideCentralRepoCommentsAndOccurrences()) {
-            Pair<Long, String> countAndDescription = getCountPropertyAndDescription(attribute);
-            properties.add(new NodeProperty<>(OCCURRENCES.toString(), OCCURRENCES.toString(), countAndDescription.getRight(), countAndDescription.getLeft()));
-        }
+        // Create place holders for S C O 
+        properties.add(new NodeProperty<>(SCORE.toString(), SCORE.toString(), NO_DESCR, ""));
+        properties.add(new NodeProperty<>(COMMENT.toString(), COMMENT.toString(), NO_DESCR, ""));
```
wschaeferB: Should we have a description like "Value loading" or something to that extent instead of "no description" as the temporary description for these columns.  I ask this because some of many cells will appear to be blank even after they load for these columns and the description is how we were giving the user a little information about what the cell value displayed means. "no description" doesn't really tell them anything.

Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
```diff
@@ -81,17 +81,6 @@ protected Sheet createSheet() {
 
         sheetSet.put(new NodeProperty<>("Type", Bundle.MessageNode_Node_Property_Type(), "", getDisplayName())); //NON-NLS
 
-        addScoreProperty(sheetSet, tags);
-
-        CorrelationAttributeInstance correlationAttribute = null;
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
```
wschaeferB: Class now has unused imports of 
import org.sleuthkit.autopsy.centralrepository.datamodel.CorrelationAttributeInstance;
import org.sleuthkit.autopsy.core.UserPreferences;

Core/src/org/sleuthkit/autopsy/datamodel/AbstractContentNode.java
```diff
@@ -50,6 +59,37 @@
...
+     * populate this node.
+     */
+    static final ExecutorService backgroundTasksPool;
+    static final Integer MAX_POOL_SIZE = 10;
+    
+    /**
+     * Default no description string
+     */
+    @NbBundle.Messages("AbstractContentNode.nodescription=no description")
+    protected static final String NO_DESCR = Bundle.AbstractContentNode_nodescription();
```
wschaeferB: Causes the NO_DESCR fields in the AbstractAbstractFileNode and BlackboardArtifactNode to hide this field, we should probably remove those two.

Core/src/org/sleuthkit/autopsy/datamodel/GetSCOTask.java
```diff
@@ -0,0 +1,73 @@
...
+ * Background task to get Score, Comment and Occurrences values for an 
+ * Abstract content node.
+ * 
+ */
+class GetSCOTask implements Runnable {
+
+    private final WeakReference<AbstractContentNode<?>> weakNodeRef;
+    private final PropertyChangeListener listener;
+
+    public GetSCOTask(WeakReference<AbstractContentNode<?>> weakContentRef, PropertyChangeListener listener) {
```
wschaeferB: public in non-public class, can have public removed

Core/src/org/sleuthkit/autopsy/datamodel/AbstractContentNode.java
```diff
@@ -50,6 +59,37 @@
     T content;
     private static final Logger logger = Logger.getLogger(AbstractContentNode.class.getName());
 
+    /**
+     * A pool of background tasks to run any long computation needed to 
+     * populate this node.
+     */
+    static final ExecutorService backgroundTasksPool;
+    static final Integer MAX_POOL_SIZE = 10;
```
wschaeferB: MAX_POOL_SIZE can be private

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -673,17 +690,19 @@ protected final void addScoreProperty(Sheet.Set sheetSet, List<Tag> tags) {
...
     @NbBundle.Messages({"BlackboardArtifactNode.createSheet.count.name=O",
         "BlackboardArtifactNode.createSheet.count.displayName=O",
         "BlackboardArtifactNode.createSheet.count.noCentralRepo.description=Central repository was not enabled when this column was populated",
         "BlackboardArtifactNode.createSheet.count.hashLookupNotRun.description=Hash lookup had not been run on this artifact's associated file when the column was populated",
         "# {0} - occuranceCount",
         "BlackboardArtifactNode.createSheet.count.description=There were {0} datasource(s) found with occurances of the correlation value"})
 
-    protected final void addCountProperty(Sheet.Set sheetSet, CorrelationAttributeInstance attribute) {
+    @Override
+    protected Pair<Long, String> getCountPropertyAndDescription(CorrelationAttributeInstance attribute) {
```
wschaeferB: public api change

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -628,8 +643,10 @@ protected final void addCommentProperty(Sheet.Set sheetSet, List<Tag> tags, Corr
         "BlackboardArtifactNode.createSheet.taggedItem.description=Result or associated file has been tagged.",
         "BlackboardArtifactNode.createSheet.notableTaggedItem.description=Result or associated file tagged with notable tag.",
         "BlackboardArtifactNode.createSheet.noScore.description=No score"})
-    protected final void addScoreProperty(Sheet.Set sheetSet, List<Tag> tags) {
-        Score score = Score.NO_SCORE;
+    
+    @Override
+    protected Pair<DataResultViewerTable.Score, String> getScorePropertyAndDescription(List<Tag> tags) {
```
wschaeferB: public api change

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -581,16 +596,17 @@ protected final CorrelationAttributeInstance getCorrelationAttributeInstance() {
...
      * @param attribute the correlation attribute associated with this
      *                  artifact's associated file, null if central repo is not
      *                  enabled
+     * @return comment property
      */
     @NbBundle.Messages({"BlackboardArtifactNode.createSheet.comment.name=C",
         "BlackboardArtifactNode.createSheet.comment.displayName=C"})
-    protected final void addCommentProperty(Sheet.Set sheetSet, List<Tag> tags, CorrelationAttributeInstance attribute) {
+    @Override
+    protected DataResultViewerTable.HasCommentStatus getCommentProperty(List<Tag> tags, CorrelationAttributeInstance attribute) {
```
wschaeferB: public api change

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -351,17 +366,15 @@ protected Sheet createSheet() {
...
-        }
-        addCommentProperty(sheetSet, tags, correlationAttribute);
-
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
-            addCountProperty(sheetSet, correlationAttribute);
-        }
+        // Create place holders for S C O 
+        sheetSet.put(new NodeProperty<>(Bundle.BlackboardArtifactNode_createSheet_score_name(), Bundle.BlackboardArtifactNode_createSheet_score_displayName(), NO_DESCR, ""));
+        sheetSet.put(new NodeProperty<>(Bundle.BlackboardArtifactNode_createSheet_comment_name(), Bundle.BlackboardArtifactNode_createSheet_comment_displayName(), NO_DESCR, ""));
+        sheetSet.put(new NodeProperty<>(Bundle.BlackboardArtifactNode_createSheet_count_name(), Bundle.BlackboardArtifactNode_createSheet_count_displayName(), NO_DESCR, ""));
```
wschaeferB: We should only add the count property to sheet set when UserPreferences.hideCentralRepoCommentsAndOccurrences() == false so that the column does not exist when its contents would be always be empty.

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -368,18 +327,16 @@ public String toString() {
...
+        properties.add(new NodeProperty<>(OCCURRENCES.toString(), OCCURRENCES.toString(), NO_DESCR, ""));
```
wschaeferB: We should only add the occureence property to sheet set when UserPreferences.hideCentralRepoCommentsAndOccurrences() == false so that the column does not exist when its contents would be always be empty.

Core/src/org/sleuthkit/autopsy/datamodel/AbstractContentNode.java
```diff
@@ -240,4 +280,79 @@ public int getContentChildrenCount() {
...
+        //setSheet() will notify Netbeans to update this node in the UI.
+        this.setSheet(visibleSheet);
+    }
+    
+    /**
+     * Reads and returns a list of all tags associated with this content node.
+     * 
+     * @return list of tags associated with the node.
+     */ 
+    abstract protected List<Tag> getAllTagsFromDatabase();
```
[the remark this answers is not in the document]
raman-bt: I went back and forth on this in my head several times - regarding pro and cons of defining abstract methods vs null/dummy implementation in the base class.  For exactly the same reason, I actually originally went with having null  methods in the base class so I wouldn't have to have duplicate them in the said classes - but then Codacy complained and suggested base class methods should be abstract,  so I switched it.  If you really feel strongly about it, I can go back to it.   Probably the 'pure' approach is to have have abstract methods in the base class, and define  an intermediate class with null implementations that both VolumeNode & ImageNode derive from, but I'm trying to avoid introducing yet another class in the hierarchy...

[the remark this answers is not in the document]
wschaeferB: I definitely do not feel strongly one way or another. If codacy wants it this way that is certainly a benefit.

PR #4824, opened 2019-05-30
1250 message threader

kellykelly3:


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessage.java
```diff
@@ -80,8 +83,34 @@ String getSubject() {
...
         if (subject != null) {
             this.subject = subject;
+            if(subject.matches("^[R|r][E|e].*?:.*")) {
+                this.simplifiedSubject = subject.replaceAll("[R|r][E|e].*?:", "").trim();
+                isReplySubject = true;
+            } else {
+                this.simplifiedSubject = subject;
+            }
+        } else {
+            this.simplifiedSubject = null;
```
bcarrier: We should set this to empty string, like subject is.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+ * This threader is based heavely off of the algorithum found at
+ * <a href="https://www.jwz.org/doc/threading.html">
+ * "message threading." by Jamie Zawinski</a>
+ *
+ */
+final class EmailMessageThreader {
+
+    private int bogus_id_count = 0;
+
+    public void threadMessages(List<EmailMessage> emailMessages, String threadIDPrefix) {
```
bcarrier: Seems like this should all be in a constructor since the class isn't useful without this data being passed in.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+        }).filter((container) -> (container.hasChildren())).forEachOrdered((container) -> {
+            printContainerSet(container.getChildren(), prefix + "\t");
+        });
+    }
+
+    /**
+     * The container object is used to wrap and email message and track the
+     * messages parent and child messages.
+     */
+    final class Container {
```
bcarrier: I think there should be "Email" somewhere in the name of this class to make it more clear it represents a single Email.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+
+            // Check the id_table for an existing Container for message-id
+            Container container = id_table.get(messageID);
+
+            // An existing container for message-id was found
+            if (container != null) {
+                // If the existing Container has a message already assocated with it
+                // emailMessage maybe a duplicate, so we don't lose the existance of 
+                // the duplicate message assign it a bogus message-id
+                if (container.hasMessage()) {
```
bcarrier: In what scenario would we have a Container in the map without a message?

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+    /**
+     * Loops throught message's list of references, creating objects as needed
+     * and setting up the parent child relationships amoung the messages.
+     *
+     * @param message   The current email messags
+     * @param container Container object for message
+     * @param id_table  Hashtable of known message-id\container pairs
+     */
+    void processMessageReferences(EmailMessage message, Container container, Map<String, Container> id_table) {
+        List<String> referenceList = message.getReferences();
```
bcarrier: The API docs for getReferences() seem to be incorrect (before this PR). They say that we return empty string.  For consistency, I think getReferences should return an empty list if there are no references instead of returning null.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
bcarrier: Just saw it from below.  We should add a line here that it could be empty if we've seen a reference to this ID, but haven't seen the actual message yet.

kellykelly3: When the initial list of message ids is created we go through the message and then its list of reference's.  When we process the reference list we only have the ID, not the actual message (its assumed that we will probably come across it later).  If we don't ever come across that message, the container would exist without the message.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: I took a lot of the naming conventions straight from the algorithm docs so as to not confuse myself\remember I named it something different.  

If you would like Email in the name I can change it.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+        }
+
+        // No references, nothing to do
+        if (referenceList == null) {
+            return;
+        }
+
+        Container parent_ref = null;
+        Container ref;
+
```
bcarrier: We should add a comment that referenceList is sorted by parent order.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+                && (parent_ref == container
+                || container.isChild(parent_ref))) {
+            parent_ref = null;
+        }
+
+        // If container already has a parent, the parent was assumed based on
+        // the list of references from another message.  parent_ref will be
+        // the real parent of container so throw away the old parent and set a
+        // new one.
+        if (container.hasParent()) {
```
bcarrier: We should only do this if parent_ref != null, right?

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+        containerSet.removeAll(containersToRemove);
+    }
+
+    /**
+     * Recursively work through the list of parent's children removing empy
+     * containers.
+     *
+     * @param parent
+     */
+    boolean pruneChildren(Container parent) {
```
bcarrier: Shoudl return what the true/false return value mean.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
bcarrier: Or, make it static since it doesn't really save any state.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+            }
+        });
+
+        containerSet.removeAll(containersToRemove);
+    }
+
+    /**
+     * Recursively work through the list of parent's children removing empy
+     * containers.
+     *
```
bcarrier: I think we shoudl add to this description something to the effect of "if this passed in container does not have a message associated with it, it will get removed and its children will be assigned to its parent."

At least I think that is what is going on here.

kellykelly3: That is what is happening, I will put more comments.  It look me awhile to get that bit right.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: No, it took me a bit to wrap my head around this.  I will explain\update the comment.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Now that the emails are grouped together by references\message ID take
+     * another pass through and group together messages with the same simplified
+     * subject.
```
bcarrier: For me, it would have been useful to have a line in here to the effect of "the same thread could currently be in spread out in different groups because some of the messages are missing.  Try to regroup them based on subject."

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+            subject_table.put(rootSubject, newParent);
+
+            finalSet.add(newParent);
+        }
+        return finalSet;
+    }
+
+    /**
+     * Creates a Hashtable of Container and subjects. There will be one
+     * Container subject pair for each unique subject.
```
bcarrier: We should document here what happens when there are groups with the same subject.  Which one get saved?

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: I checked the original code that I based my off of an there is not a check of parent_ref, here are the comments from that code:
```
      // If it has a parent already, that's there because we saw this message
      // in a references field, and presumed a parent based on the other
      // entries in that field.  Now that we have the actual message, we can
      // be more definitive, so throw away the old parent and use this new one.
      // Find this container in the parent's child-list, and unlink it.
      //
      // Note that this could cause this message to now have no parent, if it
      // has no references field, but some message referred to it as the
      // non-first element of its references.  (Which would have been some
      // kind of lie...)
```

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+
+    private int bogus_id_count = 0;
+
+    public void threadMessages(List<EmailMessage> emailMessages, String threadIDPrefix) {
+        HashMap<String, Container> id_table = createIDTable(emailMessages);
+        Set<Container> rootSet = getRootSet(id_table);
+
+        pruneEmptyContainers(rootSet);
+
+        Set<Container> finalSet = groupBySubject(rootSet);
```
bcarrier: For consistency, is it accurate to call 'finalSet' as 'finalRootSet'?

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: At this point in the code it doesn't matter which one gets saved. But it would be the first one we find.  This is just a list of unique subjects.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+                    (!reSubject && (tableContainer.getMessage() != null && tableContainer.getMessage().isReplySubject()))) { //current doesn't have RE in it, use current instead of the one in the table
+                subject_table.put(subject, rootSetContainer);
+            }
+
+        }
+
+        return subject_table;
+    }
+    
+    private void assignThreadIDs(Set<Container> containerSet, String IDPrefix) {
```
bcarrier: Add JavaDoc. 

is 'containerSet' at this point the same thing that we've been calling a 'rootSet'?

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+         * @param message
+         */
+        Container(EmailMessage message) {
+            this.message = message;
+        }
+
+        /**
+         * Returns the EmailMessage object
+         *
+         * @return
```
bcarrier: Should document that this can return null.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+
+            return false;
+        }
+
+        /**
+         * Clears the Containers list of children.
+         *
+         */
+        void clearChildren() {
+            children.clear();
```
bcarrier: SHould check if children is not null

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+        /**
+         * Removes the given container from the list of children.
+         *
+         * @param child - the container to remove from the children list
+         *
+         * @return - True if the given container successfully removed from the
+         *         list of childrent
+         */
+        boolean removeChild(Container child) {
+            return children.remove(child);
```
bcarrier: Check children.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: Yes.  I believe so.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+         * @return True if the child list is not null or empty
+         */
+        boolean hasChildren() {
+            return children != null && !children.isEmpty();
+        }
+
+        /**
+         * Returns the list of children of this container.
+         *
+         * @return The child list.
```
bcarrier: Should note that null can be returned.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -410,7 +410,8 @@ static String getRelModuleOutputPath() throws NoCurrentCaseException {
     private void processEmails(List<EmailMessage> emails, AbstractFile abstractFile) {
         List<AbstractFile> derivedFiles = new ArrayList<>();
         
-       
+        EmailMessageThreader threader = new EmailMessageThreader();
+        threader.threadMessages(emails, String.format("%d", abstractFile.getId()));
```
bcarrier: Add try/catch.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
```diff
@@ -0,0 +1,674 @@
...
+                child.clearChildren();
+
+            }
+        }
+
+        parent.addChildren(add);
+        parent.removeChildren(remove);
+
+        if (!parent.hasMessage() && grandParent != null) {
+            children.forEach((child) -> {
```
bcarrier: A comment here would have helped, to the effect of "we are going to remove this node. We will reset the children now and the other references will get updated by the caller".

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: Let me clarify, its kind of assumed to be in parent order and it might be.  If we wanted to get really fancy with the algorithm the threads should be sorted by date at the very end.    I think saying that this list could be in a particular order is misleading.

The final step of the algorithm which is optional, is to sort the threads by date to put them in true order.  The documents suggest that it could also be done by the caller\client of this code.

I will document that the threads will not be in the correct time order.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: See above comment.

thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/EmailMessageThreader.java
kellykelly3: I put a big comment before the add.

rcordovano:
@kellykelly3, are any of these Codacy issues new with your code? Most of them look like quick fixes.

kellykelly3:
> @kellykelly3, are any of these Codacy issues new with your code? Most of them look like quick fixes.

Most of them are.  I don't know that I can reduce the complexity of the functions easily.  But I will take care of the rest.

kellykelly3:
@rcordovano I have addressed all of the "easy" codacy issues.   I left the catch Exception around the call to my threader as a broad catch anything that might go wrong so that we will always create the artifacts.

PR #4843, opened 2019-06-05
5035 make ui to configure logical image

jkho:


jkho:
Don't comment yet. I have to fix the codacy errors.

jkho:
I fixed as much as I can with Codacy. Please review.

Core/src/org/sleuthkit/autopsy/configurelogicalimager/LogicalImagerRule.java
```diff
@@ -0,0 +1,254 @@
...
+            this.shouldAlert = shouldAlert;
+            return this;
+        }
+        
+        public Builder shouldSave(boolean shouldSave) {
+            this.shouldSave = shouldSave;
+            return this;
+        }
+        
+        public Builder name(String name) {
```
APriestman: I agree with codacy that it is odd that this method and others have the same name as your field. Is there a reason not to name them "setName()" etc?

jkho: I don't agree. At least in the platforms and clouds code, that is how we do builders.
Just extra typing :(

jkho:
@APriestman I fixed Codacy exception problem.

jkho:
@APriestman So I resolved one conflict in this submit https://github.com/sleuthkit/autopsy/pull/4843/commits/404a7ce40de2dd56a613b95ca4f5004bb730d7ff which shows only one file change.
But now a lot of files are pulled from the release-4.12.0 branch into my branch. My branch must be really out of date with the release branch.

APriestman:
It seems like a lot of these classes and methods could be package scope instead of public.
Also there are still files unrelated to your PR included

jkho:
Let me fix them.

jkho:
@APriestman  I reverted 2 extra files. Please review.

PR #4844, opened 2019-06-05
5065 export image tags

dannysmyda:


dannysmyda:
The last Codacy suggestion is not valid, those fields cannot be made final. This PR is ready to go.

PR #4848, opened 2019-06-06
5130 Export CSV button and context menu

APriestman:


Core/src/org/sleuthkit/autopsy/directorytree/ExportCSVAction.java
```diff
@@ -0,0 +1,347 @@
...
+        
+        /**
+         * Convert list of values to a comma separated string.
+         * 
+         * @param values Values to convert
+         * 
+         * @return values as CSV
+         */
+        private String listToCSV(List<String> values) {
+            return "\"" + String.join("\",\"", values) + "\"\n";
```
raman-bt: What happens if a value has a comma in it?

PR #4851, opened 2019-06-06
Fixed bug which was allocating too much memory to byte buffers (5096)

eugene7646:


Core/src/org/sleuthkit/autopsy/texttranslation/ui/TranslatedTextViewer.java
```diff
@@ -343,7 +343,7 @@ private String extractText(AbstractFile source, boolean ocrEnabled) throws IOExc
             }
 
             // The trim is on here because HTML files were observed with nearly 1MB of white space at the end
```
APriestman: Comment is no longer needed

eugene7646: Good point! Done.

PR #4855, opened 2019-06-07
5122 hang in translation viewer

dannysmyda:
This PR caps original text extraction to 50KB. The warning limit was removed and will be added back next in JIRA-5094 with more specific limits per translation service (ex. for Bing Translator: "Up to the first 5KB text translated").


dannysmyda:
Bumped it down to 25K. Changing component direction (for languages like arabic) takes awhile for 50K worth of text.

PR #4859, opened 2019-06-07
5055 occurrences column for artifacts

wschaeferB:


Core/src/org/sleuthkit/autopsy/datamodel/GetSCOTask.java
```diff
@@ -42,28 +45,58 @@
...
+                if (bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_ENCRYPTION_DETECTED.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_ENCRYPTION_SUSPECTED.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_INTERESTING_FILE_HIT.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_METADATA_EXIF.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_KEYWORD_HIT.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_OBJECT_DETECTED.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_EXT_MISMATCH_DETECTED.getTypeID()
+                        || bbArtifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_HASHSET_HIT.getTypeID()) {
+                    occurrencesAttribute = fileAttribute;
```
raman-bt: Minor optional comment: It might it be more readable to define a set of these artifact types and then check if the given artifact type is in that set.

wschaeferB: I personally prefer it this way as opposed to building a set in the constructor (or each time the method is executed). Another option would be a static array and loop through it with explicit break when condition met, but if no one has a strong preference I personally find the medium sized conditional block easier to read since the conditions it is checking are right there.

PR #4887, opened 2019-06-12
1256 message thread ui

kellykelly3:


kellykelly3:
Handled some of the codacy issues. Codacy really wanted some of the vars defined by the gui builder to be private to the initComponent function, it also doesn't like the length of that function.  Not sure what can be done about that.

rcordovano:
I don't understand. The variables in the generated code are usually
private. Has something been done to change that?

kellykelly3:
> I don't understand. The variables in the generated code are usually private. Has something been done to change that?

The generated code variable are private, private to the class.  The codacy suggestion is to declare them in initComponents because they are not used outside of that function.  It's weird.  I have never seen that before.

PR #4907, opened 2019-06-17
2955 dont index nsrl

wschaeferB:


raman-bt:
I did notice this PR is targeted to 'develop' - are we merging to develop directly?

PR #4908, opened 2019-06-17
5209 sco column broken

wschaeferB:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -366,7 +367,7 @@ protected Sheet createSheet() {
         // Create place holders for S C O 
         sheetSet.put(new NodeProperty<>(Bundle.BlackboardArtifactNode_createSheet_score_name(), Bundle.BlackboardArtifactNode_createSheet_score_displayName(), VALUE_LOADING, ""));
         sheetSet.put(new NodeProperty<>(Bundle.BlackboardArtifactNode_createSheet_comment_name(), Bundle.BlackboardArtifactNode_createSheet_comment_displayName(), VALUE_LOADING, ""));
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
+        if (EamDbUtil.allowUseOfCentralRepository() && UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
```
raman-bt: I believe this should use EamDb.isEnabled() instead.

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -327,7 +328,7 @@ public String toString() {
         // Create place holders for S C O 
         properties.add(new NodeProperty<>(SCORE.toString(), SCORE.toString(), VALUE_LOADING, ""));
         properties.add(new NodeProperty<>(COMMENT.toString(), COMMENT.toString(), VALUE_LOADING, ""));
-        if (UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
+        if (EamDbUtil.allowUseOfCentralRepository() && UserPreferences.hideCentralRepoCommentsAndOccurrences() == false) {
```
raman-bt: I believe this should use EamDb.isEnabled() instead.

PR #4915, opened 2019-06-18
5202 export tags saving bug

dannysmyda:
All images/gifs that can be tagged in Autopsy should now be exportable.


dannysmyda:
This PR also includes some refactoring work.

dannysmyda:
> My one comment on this is we use "png" and ".png" at so many places. I think it would be useful to define static finals for those strings instead of sprinkling them all over. More than anything it would call out that some things are always saved as a PNG.

Good point, I'll change that. Thanks!

PR #4924, opened 2019-06-20
1247 - Added back the ImmutableSets to fix hang

kellykelly3:


kellykelly3:
I am going to close this pull request because I think I am able to narrow down to the list that is causing the hang and am trying to figure out why.  I will create a new PR with the updated changes.

PR #4925, opened 2019-06-20
5230 add version number to logical imager config file

jkho:


Core/src/org/sleuthkit/autopsy/logicalimager/configuration/LogicalImagerConfig.java
```diff
@@ -28,6 +28,12 @@
  */
 class LogicalImagerConfig {
 
+    static private final String DEFAULT_VERSION = "1.0";
```
APriestman: Would it be more accurate to name this "CURRENT_VERSION"?

PR #4937, opened 2019-06-25
5224 clean up tsk logical imager.exe packaging

jkho:


nbproject/project.properties
```diff
@@ -6,8 +6,8 @@ app.name=${branding.token}
 ### if left unset, version will default to today's date
 app.version=4.12.0
 ### build.type must be one of: DEVELOPMENT, RELEASE
-build.type=RELEASE
-#build.type=DEVELOPMENT
+#build.type=RELEASE
+build.type=DEVELOPMENT
```
rcordovano: This change needs to be removed. We need this branch to be using the DEVELOPMENT build.type.

Core/src/org/sleuthkit/autopsy/logicalimager/configuration/ConfigureLogicalImager.java
```diff
@@ -41,18 +43,20 @@
...
         id = "org.sleuthkit.autopsy.configurelogicalimager.ConfigureLogicalImager"
 )
-@ActionRegistration(
-        displayName = "#CTL_ConfigureLogicalImager"
-)
+@ActionRegistration(displayName = "#CTL_ConfigureLogicalImager", lazy = false)
 @ActionReference(path = "Menu/Tools", position = 2000, separatorBefore = 1999)
 @Messages("CTL_ConfigureLogicalImager=Configure Logical Imager")
-public final class ConfigureLogicalImager implements ActionListener {
+public final class ConfigureLogicalImager extends CallableSystemAction {
```
rcordovano: This class needs to have action in the name. Let's call it  CreateLogicalImagerAction.

Core/src/org/sleuthkit/autopsy/logicalimager/configuration/ConfigureLogicalImager.java
```diff
@@ -41,18 +43,20 @@
...
+
+    private static final String DISPLAY_NAME = Bundle.CTL_ConfigureLogicalImager();
```
rcordovano: I think Will has changed the display name for this action so there should not be anything to do, but please coordinate with him.

Core/src/org/sleuthkit/autopsy/logicalimager/configuration/ConfigureLogicalImager.java
jkho: @wschaeferB  PR will address this.

nbproject/project.properties
jkho: done

bcarrier:
This should also delete the .exe from the repo, right?

PR #4949, opened 2019-06-26
5240 Change OpenCvLoader behavior

dannysmyda:


CoreLibs/src/org/sleuthkit/autopsy/corelibs/OpenCvLoader.java
```diff
@@ -18,43 +18,31 @@
...
-        if (!OPEN_CV_LOADED) {
-             //exception should never be null if the open cv isn't loaded but just in case
-            if (exception != null) {
-                throw exception;
-            } else {
-                throw new UnsatisfiedLinkError("OpenCV native library failed to load");
-            }
-
-        }
+    public static boolean isOpenCvLoaded() {
```
bcarrier: I think this needs to continue to throw an exception. Otherwise it is backward incompatible.

dannysmyda: From Richard: "I think that in this case we can just modify this exported public API. I can’t imagine anybody is using it but us. It is cool with me if you want to remove the exception specification and the parameter documentation. The proper thing to do would be to deprecate the method and make a new version of it, but it doesn’t seem worth it to me."

rcordovano: @dannysmyda, please go ahead and do the deprecation and make a new method so that we preserve this exported public API after all, thanks!

CoreLibs/src/org/sleuthkit/autopsy/corelibs/OpenCvLoader.java
```diff
@@ -45,6 +54,7 @@
      *   Note that the stack trace in the exception can be confusing because it refers to a
      *   past invocation. 
      */
+    @Deprecated
```
rcordovano: In addition to the deprecated annotation, we also need to add @deprecated to the comment with words top the effect of "Use hasOpenCvLoaded instead."

CoreLibs/src/org/sleuthkit/autopsy/corelibs/OpenCvLoader.java
```diff
@@ -57,4 +67,13 @@ public static boolean isOpenCvLoaded() throws UnsatisfiedLinkError {
...
         return OPEN_CV_LOADED;
     }
+    
+    /**
+     * Return whether OpenCV library has been loaded.
+     * 
+     * @return true if OpenCV library was loaded, false if not.
+     */
+    public static boolean hasOpenCvLoaded() {
+        return OPEN_CV_LOADED;
```
rcordovano: I know that the OPEN_CV_LOADED identifier is not new, but while we are in here we should put it in standard varibale name camel case, since this is not a constant.

PR #4966, opened 2019-06-27
5265 logical imager dsp should not be enabled for network drives

jkho:


Core/src/org/sleuthkit/autopsy/logicalimager/dsp/LogicalImagerPanel.java
```diff
@@ -482,6 +482,9 @@ private void refreshButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN
         int i = 0;
         for (File root : roots) {
             String description = FileSystemView.getFileSystemView().getSystemTypeDescription(root);
+            if ("Network Drive".equalsIgnoreCase(description)) { // NON-NLS
```
APriestman: Will this work on a non-English version of windows?

jkho: I don't know. I can't find a solution that doesn't test for "Network Drive".

jkho: @APriestman I found this https://gist.github.com/digulla/31eed31c7ead29ffc7a30aaf87131def which spawn a process with "CMD /c net use Z:" to test for network drive.

Is this solution OK? Requires CMD and net available on the target system (which should be available on all Windows system)

Core/src/org/sleuthkit/autopsy/logicalimager/dsp/DriveListUtils.java
```diff
@@ -48,4 +54,36 @@ public static String humanReadableByteCount(long bytes, boolean si) {
...
+            String line;
+            try (BufferedReader in = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
+                while ((line = in.readLine()) != null) {
+                    consoleOutput.append(line).append("\r\n");
+                }
+            }
+            
+            int rc = p.waitFor();
+//            System.out.println(consoleOutput);
+//            System.out.println("rc=" + rc);
```
APriestman: Please clean up the commented out lines

PR #4987, opened 2019-06-28
5239 prevent duplicate named rules being created

wschaeferB:


Core/src/org/sleuthkit/autopsy/logicalimager/configuration/Bundle.properties-MERGED
```diff
@@ -37,6 +37,9 @@ ConfigVisualPanel2.editConfiguration=Configure imager
 ConfigVisualPanel2.editRuleError=Edit rule error
 ConfigVisualPanel2.editRuleSet=Edit Rule
 ConfigVisualPanel2.newRule.name=New Rule
+# {0} - ruleName
+ConfigVisualPanel2.newRuleError.duplicateName=A rule with named "{0}" already exists please choose a different name
```
APriestman: Maybe change to "A rule with name {0} already exists. Please enter a different rule name"

PR #5005, opened 2019-07-02
5297 codacy error cleanup

wschaeferB:


Core/src/org/sleuthkit/autopsy/textextractors/SqliteTextExtractor.java
```diff
@@ -204,7 +204,7 @@ private void fillBuffer(String val) {
          */
         @Override
         public int read(char[] cbuf, int off, int len) throws IOException {
-            buf = cbuf;
+            buf = cbuf.clone();
```
rcordovano: Is the caller expecting this method to fill the cbuf array, but we are now filling a clone of it? Frankly, I don't understand why buf is a instance variable anyway - it is only referenced in this single read method...

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/FileExportRuleSet.java
```diff
@@ -337,12 +351,12 @@ boolean conditionsAreEqual(Rule that) {
                 return false;
             }
             this.fileSizeConditions.sort(null);
-            that.fileSizeConditions.sort(null);
+            that.sortFileSizeConditions();
```
rcordovano: It might be a tiny bit more readable to call this.sortFileSizeConditions as well. Ditto on line 359.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/FileExportRuleSet.java
```diff
@@ -376,13 +390,18 @@ public int compareTo(Rule that) {
         List<Long> evaluate(long dataSourceId) throws ExportRulesException {
             try {
                 SleuthkitCase db = Case.getCurrentCaseThrows().getSleuthkitCase();
+                ResultSet resultSet = null;
                 try (SleuthkitCase.CaseDbQuery queryResult = db.executeQuery(getQuery(dataSourceId))) {
-                    ResultSet resultSet = queryResult.getResultSet();
+                    resultSet = queryResult.getResultSet();
```
rcordovano: Is it feasible to include the ResultSet in the try so it is closed via AutoCloseable?

Core/src/org/sleuthkit/autopsy/textextractors/SqliteTextExtractor.java
wschaeferB: Apparently yes we need buf to equal cbuf not a copy of it. So I will undo this change, and leave it as a codacy error.

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/FileExportRuleSet.java
wschaeferB: good call

Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/FileExportRuleSet.java
wschaeferB: Apparently it is when putting each as its own line of code in the parens, thank you for helping me learn this.

wschaeferB:
After looking at the remaining clone methods and the usage of the arrays they are cloning I feel pretty confident I didn't make the same mistake in changing any of them.

PR #5013, opened 2019-07-05
1282, 1283 - modified filterPane layout and fixed chart tool tips

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/Bundle.properties
```diff
@@ -22,15 +22,13 @@ CVTTopComponent.browseVisualizeTabPane.AccessibleContext.accessibleName=Visualiz
 CVTTopComponent.vizPanel.TabConstraints.tabTitle_1=Visualize
 VisualizationPanel.fitGraphButton.text=
 VisualizationPanel.jTextArea1.text=Right-click an account in the Browse Accounts table, and select 'Visualize' to begin.
-VisualizationPanel.zoomLabel.text=100%
-VisualizationPanel.jLabel2.text=Zoom:
-VisualizationPanel.fitZoomButton.toolTipText=fit visualization
+VisualizationPanel.fitZoomButton.toolTipText=Fit Visualization
 VisualizationPanel.fitZoomButton.text=
-VisualizationPanel.zoomActualButton.toolTipText=reset zoom
```
raman-bt: I think the tool tip text should read like a sentence, i.e. only the first word should begin with upper case. So i think the tool tips should be:
Reset zoom
Fit visualization
Zoom in
Click to go forward

and so on...

I'd created this bug because some of the tool tips started with lower case.

kellykelly3: I didn't love how it looked with just the first letter capitalized.  They are not full sentences.

I looked at what other tools do and I found that like netBean capitalized like a title.  So I went with a capitalization like a title.  
I think I should change "Click to Go Forward" to be just "Go Forward"  That will be more consistent with the title style.

rcordovano:
@kellykelly3 and @raman-bt, the NetBeans tool tips proposed as an exemplar don't really apply to what we are looking at here, since they are supplying the menu text for toolbar buttons. In the example of what we are discussing that @kellykelly3 gave me, namely the "Fit Visualization" button tool tip, we should provide a short sentence that adds some explanation for a novice user rather than repeating the button name, i.e., "Fits visualization to available space." (or whatever it does). We should make such tool tips as concise as we can without making them cryptic and we should use sentence case.

PR #5024, opened 2019-07-11
1289 Cleanup of the postArtifact code in RecentActivity

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -361,10 +361,16 @@ private void processCacheIndexFile(AbstractFile indexAbstractFile) {
...
 
         context.addFilesToJob(derivedFiles);
-        
-        services.fireModuleDataEvent(new ModuleDataEvent(moduleName, BlackboardArtifact.ARTIFACT_TYPE.TSK_DOWNLOAD_SOURCE, !sourceArtifacts.isEmpty() ? sourceArtifacts : null));
-        services.fireModuleDataEvent(new ModuleDataEvent(moduleName, BlackboardArtifact.ARTIFACT_TYPE.TSK_WEB_CACHE, !webCacheArtifacts.isEmpty() ? webCacheArtifacts : null));
 
+        Blackboard blackboard = currentCase.getSleuthkitCase().getBlackboard();
+        
+        try {
+            blackboard.postArtifacts(sourceArtifacts, moduleName);
```
bcarrier: for consistency, shoud this call the one in Extract?

kellykelly3: Yes, but ChromeCacheExtractor seems to be the only class in RecentActivity that does not subclass Extract. I decided to  I just focused on the problem at hand.  And write a Jira to come back and fix later.

I can fix it now if you would like.

PR #5026, opened 2019-07-11
reorganized multiuser setup.  Minor changes elsewhere

bcarrier:


docs/doxygen-user/multi-user/installMultiUser.dox
```diff
@@ -0,0 +1,40 @@
...
+
+Step 5: \ref install_solr
+
+Step 6: \ref install_activemq
+
+Step 7: \ref multiuser_install_clients
+
+
+
+When you setup the above services, write down the addresses, user names, and passwords or each so that you can configure each of the client systems afterwards.
```
APriestman: Typo - "and passwords or each"

docs/doxygen-user/multi-user/installSystems.dox
```diff
@@ -0,0 +1,42 @@
...
+- Network storage
+
+You can run each of these on their own dedicated VM, but that is not necessary.  We recommend:
+
+- Server 1: PostgreSQL and ActiveMQ 
+ - Both of these are low overhead services. 
+- Server 2: Apache Solr and Shared Storage (if you are using Windows File Sharing).
+ - Apache Solr uses a lot of memory, so we keep it by itself. 
+ - The benefit of running the shared storage on the same computer as Solr is performance. Solr is much faster when it can write to local storage instead of over the network.
+ - If you are using a different form of shared storage, then we still recommend keeping Solr on its own.
```
APriestman: This section doesn't seem to flow very well. It might be better as a single bullet point? Like:
Apache Solr uses a lot of memory, so we recommend keeping it by itself. The exception is if you are using Windows File Sharing, since putting the shared storage on the same VM as Solr will increase performance. Solr is much faster when it can write to local storage instead of over the network. If you are using a different form of shared storage, then we recommend keeping Solr on its own.

docs/doxygen-user/multi-user/installMultiUser.dox
```diff
@@ -0,0 +1,40 @@
...
+Each Autopsy client will then use those shared resources instead of the embedded versions that are used for single-user cases.  
+
+\image html multi-user-network.png
+
+
+\section multiuser_install_install Cluster Installation and Configuration
+
+Let's now step through the process of setting up an Autopsy cluster. 
+
+Step 1: \ref multiuser_system
```
APriestman: I personally would prefer the link to go to install_multiuser_systems_page instead of the first section. The way it is makes me think that there's more data above where the link went to, but it's just the page title. There are a few like this.

docs/doxygen-user/multi-user/installSolr.dox
```diff
@@ -0,0 +1,145 @@
...
+
+Solr's embedded Zookeeper is also used as a coordination service for Autopsy. 
+
+\section install_solr_prereq Prerequisites
+
+We use Bitnami Solr, which packages Solr as a Windows service. 
+
+You will need:
+- A 64-bit version of the Java Runtime Environment (JRE) from http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html. 
+- The Apache Solr 4.10.3-0 installation package.  This is no longer available from its original source, but you can find it on our site: https://sourceforge.net/projects/autopsy/files/CollaborativeServices/Solr.
```
APriestman: Based on a recent forum question, I might add a note that you must get this version of solr and can't use a newer version.

docs/doxygen-user/multi-user/installMultiUser.dox
bcarrier: Fixed.

docs/doxygen-user/multi-user/installSystems.dox
bcarrier: Added a line above the suggested configurations.

docs/doxygen-user/multi-user/installMultiUser.dox
bcarrier: Fixed and renamed the anchor names to consistently mention they were pages.

docs/doxygen-user/multi-user/installSolr.dox
bcarrier: Added.

docs/doxygen-user/multi-user/installPostgres.dox
```diff
@@ -0,0 +1,119 @@
+/*! \page install_postgresql_page Install and Configure PostgreSQL
+
+
+\section install_post_over Overview
+
+In a multi-user case, a central PostgreSQL database server is used instead of the embedded SQLite databases.  
+
+A new database will be created for each case and the database will be stored on the local drive.
```
esaunders: With the introduction of the Central Repository it may no longer be a good idea to store databases in the default location (system drive). We might want to redo this section to encourage users to choose a location where databases will be stored.

docs/doxygen-user/multi-user/installPostgres.dox
```diff
@@ -0,0 +1,119 @@
...
+<br><br>
+Add the line highlighted in yellow below, formatted with spaces between the entries, adjusting the IP address to an appropriate value as described above.
+<br><br>
+\image html postgresqlinstall4.PNG
+<br>
+<br><br>
+If you intend to use PostgreSQL from machines on a different subnet, you need an entry in the _pg_hba.conf_ file for each subnet. 
+<br><br>
+
+3. Uncomment the following entires in the configuration file located at <i>C:\\Program Files\\PostgreSQL\\9.5\\data\\postgresql.conf</i> by removing the leading "#", and change their values <i>"off"</i> as shown below.
```
esaunders: References to this location assume that databases are being stored in the default location. If we expand on the PG installation steps above to suggest that the user pick an alternative location for database storage we will need to update references to the location of the configuration files.

docs/doxygen-user/multi-user/installPostgres.dox
```diff
@@ -0,0 +1,119 @@
...
+<br><br>
+\image html postgresqlinstall7.PNG
+<br><br>
+
+
+\section install_post_test Testing
+
+You can verify that PostgreSQL is running by using either the _pgAdmin_ tool or the _psql_ tool to connect to the database server from another machine on the network.
+
+Common problems are from:
```
esaunders: Common problems are caused by:
- Firewall blocking the port (default: 5432) on the PostgreSQL server.
- Incorrectly configured database user account or incorrect credentials.
- Incorrectly configured IP address range in pg_hba.conf file.

docs/doxygen-user/multi-user/installPostgres.dox
```diff
@@ -0,0 +1,119 @@
...
+- Firewall blocking the ports on the PostgreSQL server
+- Incorrect credentials 
+- Wrong IP address range in the configuration file
+
+
+\section install_post_backup Backing Up
+
+The database are stored on the local drive (not shared storage).  So, you should backup that directory periodically. 
+
+The default location is XYZ.
```
esaunders: The default location is C:\Program Files\PostgreSQL\9.5\data but see earlier comments about how this might not be the best idea.

docs/doxygen-user/multi-user/installSolr.dox
```diff
@@ -0,0 +1,141 @@
+/*! \page install_solr_page Install and Configure Solr
+
+\section install_solr_overview Overview
+
+Autopsy uses Apache Solr to store keyword text indexes.  A central server is needed in a multi-user cluster to maintain and search the indexes.  
+
+A new Solr core is created for each case and the core/index is stored in the case folder (not on the local drive of the Solr server).
```
esaunders: I would propose changing this to "A new text index is created for each case and is stored in the case folder on shared storage (not on the local drive of the Solr server).

docs/doxygen-user/multi-user/installSolr.dox
```diff
@@ -0,0 +1,141 @@
...
+\image html solrinstall2.PNG
+
+
+- Start the service again. 
+
+\section install_solr_test Testing
+
+There are two tests that you should do before declaring that Solr has been configured. 
+
+- <b>Web Interface</b>: You should be able to see the Solr admin panel in a web browser via the URL <i>http://localhost:8983/solr/#/</i>.
```
esaunders: The URL we suggest using here (localhost) will not reveal firewall issues. We should suggest that they attempt to connect from another machine on the network.

docs/doxygen-user/multi-user/installSolr.dox
```diff
@@ -0,0 +1,141 @@
...
+ <br><br>
+ \image html solrinstall3.PNG
+ <br><br>
+
+If the service is appropriately started and you are unable to see the screenshot above, then it could be that port 8983 is blocked by your firewall. Contact your network administrator to open this port.
```
esaunders: In addition to port 8983 there is also the Zookeeper port at 9983.

PR #5032, opened 2019-07-15
TL - Fixed IllegalArg exception

kellykelly3:


Core/src/org/sleuthkit/autopsy/timeline/ui/listvew/datamodel/ListViewModel.java
```diff
@@ -119,4 +129,15 @@ public ListViewModel(FilteredEventsModel eventsModel) {
 
         return combinedEvents;
     }
+    
+    private boolean hasFileTypeEvents(List<EventType> eventTypes) {
+
+        for(EventType type: eventTypes) {
```
raman-bt: white space after for and if

Core/src/org/sleuthkit/autopsy/timeline/ui/listvew/datamodel/ListViewModel.java
```diff
@@ -106,11 +106,21 @@ public ListViewModel(FilteredEventsModel eventsModel) {
...
+                if(hasFileTypeEvents(eventTypes)) {
+                    for (int i = 0; i < eventIDs.size(); i++) {
+                        eventMap.put(eventTypes.get(i), eventIDs.get(i));
+                    }
+                    combinedEvents.add(new CombinedEvent(resultSet.getLong("time") * 1000,   eventMap));
+                } else {
+                    for (int i = 0; i < eventIDs.size(); i++) {
+                        eventMap.put(eventTypes.get(i), eventIDs.get(i));
+                        combinedEvents.add(new CombinedEvent(resultSet.getLong("time") * 1000,   eventMap));
+                        eventMap = new HashMap<>();
```
raman-bt: I'm not sure if I understand whats going on here. I understand that the  if(hasFileTypeEvents(eventTypes)) 

Why is eventMap being allocated in the loop? 

If the intent is to create a combinedEvent with a single event in eventMap each time, then a comment indicating  that would be helpful, and I'd say to move the allocation to the top of the loop.

Core/src/org/sleuthkit/autopsy/timeline/ui/listvew/datamodel/ListViewModel.java
```diff
@@ -106,11 +106,21 @@ public ListViewModel(FilteredEventsModel eventsModel) {
                 List<Long> eventIDs = unGroupConcat(resultSet.getString("eventIDs"), Long::valueOf);
                 List<EventType> eventTypes = unGroupConcat(resultSet.getString("eventTypes"),
                         typesString -> eventManager.getEventType(Integer.valueOf(typesString)).orElseThrow(() -> new TskCoreException("Error mapping event type id " + typesString + ".S")));
+                
                 Map<EventType, Long> eventMap = new HashMap<>();
-                for (int i = 0; i < eventIDs.size(); i++) {
-                    eventMap.put(eventTypes.get(i), eventIDs.get(i));
+                
+                if(hasFileTypeEvents(eventTypes)) {
```
bcarrier: We should add a comment here to the effect of:
"We want to merge together file sub-type events that are at the same time, but create individual events for other sub types of events (such as web history and web download)".

Core/src/org/sleuthkit/autopsy/timeline/ui/listvew/datamodel/ListViewModel.java
```diff
@@ -107,9 +107,11 @@ public ListViewModel(FilteredEventsModel eventsModel) {
                 List<EventType> eventTypes = unGroupConcat(resultSet.getString("eventTypes"),
                         typesString -> eventManager.getEventType(Integer.valueOf(typesString)).orElseThrow(() -> new TskCoreException("Error mapping event type id " + typesString + ".S")));
                 
+                // We want to merge together file sub-type events that are at 
+                //the same time, but create individual events for other sub
```
raman-bt: other "sub-types" ?

PR #5033, opened 2019-07-15
SevenZipExtractor improved filename decoding

ethteck:
Using mozilla's universalchardet to fix filenames by detecting the proper encoding of the names.

This works by appending the names of neighboring files and running this larger string through the `UniversalDetector` so files with shorter names that aren't as representative of the encoding are still transcoded properly.


ethteck:
Two comments for the reviewer:

My regression test results can be found here: https://drive.google.com/drive/folders/1-6ZaAAc-z8JVfNeswAG6j-ePntQUZX6j?usp=sharing

Also, I think develop should be merged into the target branch for this pr, sleuthkit:ethan_zip_detection, to make the diff easier to understand. I don't have the permissions to do this though.

.gitignore
```diff
@@ -82,7 +82,8 @@ hs_err_pid*.log
 /RecentActivity/release/
 /CentralRepository/release/
 
-/.idea/
+.idea/
+*.iml
```
wschaeferB: I am unsure what this change to the gitignore is fixing / why it is being modified is there are comment about why this is being changed somewhere that i missed? (as I have currently only done an initial skimming of all the comments)

rcordovano:
The diffs will update when you merge develop into your local feature branch
and push the branch to your fork.

PR #5034, opened 2019-07-15
5116 expand file search

APriestman:


Core/src/org/sleuthkit/autopsy/filequery/FileSearch.java
```diff
@@ -91,14 +99,94 @@ static SearchResults runFileSearchDebug(
...
+     * @param startingEntry      The first entry to return
+     * @param numberOfEntries    The number of entries to return
+     * @param caseDb             The case database
+     * @param centralRepoDb      The central repository database. Can be null if not needed.
+     * 
+     * @return A LinkedHashMap grouped and sorted according to the parameters
+     * 
+     * @throws FileSearchException 
+     */    
+    static List<AbstractFile> getGroupEntries(
```
rcordovano: Could we make this name reference files somehow?

Core/src/org/sleuthkit/autopsy/filequery/FileSearch.java
```diff
@@ -746,6 +836,516 @@ public int hashCode() {
...
+            } catch (TskCoreException ex) {
+                throw new FileSearchException("Error looking up hash set attributes", ex); // NON-NLS
+            }
+        }
+        
+        /**
+         * Callback to process the results of the CaseDbAccessManager select query. Will add
+         * the keyword list names to the list of ResultFile objects.
+         */
+        private static class SetHashSetNamesCallback implements CaseDbAccessManager.CaseDbAccessQueryCallback {
```
rcordovano: The overloading of "set" in this class name makes this name a little hard to parse. Perhaps we don't need the first "Set" in this and the other callback names.

Core/src/org/sleuthkit/autopsy/filequery/FileSearch.java
```diff
@@ -746,6 +836,516 @@ public int hashCode() {
...
+            } catch (TskCoreException ex) {
+                throw new FileSearchException("Error looking up interesting file set attributes", ex); // NON-NLS
+            }
+        }
+        
+        /**
+         * Callback to process the results of the CaseDbAccessManager select query. Will add
+         * the interesting file set names to the list of ResultFile objects.
+         */
+        private static class SetInterstingFileSetNamesCallback implements CaseDbAccessManager.CaseDbAccessQueryCallback {
```
rcordovano: There is a typo here.

PR #5035, opened 2019-07-15
4930 OOB Exception on Case Opening

dannysmyda:


Core/src/org/sleuthkit/autopsy/casemodule/CaseOpenAction.java
```diff
@@ -63,7 +63,7 @@
     private static final String PROP_BASECASE = "LBL_BaseCase_PATH"; //NON-NLS
     private static final Logger LOGGER = Logger.getLogger(CaseOpenAction.class.getName());
     private static JDialog multiUserCaseWindow;
-    private final JFileChooser fileChooser = new JFileChooser();
+    private volatile JFileChooser fileChooser;
```
rcordovano: Why does the JFileChooser reference need to be object state? Is it costly to make one each time we need one?

Also, looking at the saving of the JFileChooser reference, I notice that the static multiUserCaseWindow reference is erroneous. It is a singleton. There is no need to save a reference to it.

dannysmyda: Didn't give it much thought, i can create a brand new one each time. And okay ill fix it.

PR #5049, opened 2019-07-17
5114 file discovery

wschaeferB:


Core/src/org/sleuthkit/autopsy/filequery/SearchWorker.java
```diff
@@ -0,0 +1,95 @@
...
+    protected Void doInBackground() throws Exception {
+
+        try {
+            // Make a list of attributes that we want to add values for. This ensures the
+            // ResultFile objects will have all needed fields set when it's time to group
+            // and sort them. For example, if we're grouping by central repo frequency, we need
+            // to make sure we've loaded those values before grouping.
+            List<FileSearch.AttributeType> attrsForGroupingAndSorting = new ArrayList<>();
+            attrsForGroupingAndSorting.add(groupingAttr);
+            attrsForGroupingAndSorting.addAll(fileSort.getRequiredAttributes());
```
APriestman: This is no longer needed - the search method takes care of it

PR #5061, opened 2019-07-22
Japanese text search

akr4:
This is a WIP pull request.
Now the following features work with these changes:
- Search by Japanese text
- Highlight the Japanese words

Note that non-Japanese texts are stored in `text_general` type field in this version. It was stored in `text` field which type is `content_ws` in the previous version. I think `text_general` is better for English text search, but it can be reverted easily.

## Questions
- Is copying filenames to `CONTENT_GENERAL` or `CONTENT_JA` necessary? In the previous version, filenames are copied to `TEXT` field by `<copyField>` in Solr. However, filenames exist in METADATA section of `CONTENT_*` fields, so it seems not to be required.

### Dependency
- Added [language-detector](https://github.com/optimaize/language-detector)

### Solr schema
- Added `text_ja` type
- Added `language`, `content_ja` and `content_general` fields
- Updated the shcema version from 2.1 to 2.2.

### Index and query
The new process of index and query:
1. `Ingester` detects languages
2. `Ingester` indexes chunks and mini-chunks
  - contents of the chunks are always stored in `content_str`
  - stored in `content_ja` when Japanese, otherwise `content_general`
3. `LuceneQuery` uses `content_general` and `content_ja` for query and highlighting

index_and_query

### Remarks on `LuceneQuery_2_1` and `HighlightedText_2_1`
These classes were copied from `LuceneQuery` and `HighlightedText`, and not changed much. They provide the same functionalities of the original classes for index schema version 2.1 or earlier. On the other hand, `LuceneQuery` and `HighlightedText` are used with index schema version 2.2. They have a lot of duplicate code, but I keep them as-is for the following reasons:
- `LuceneQuery_2_1` and `HighlightedText_2_1` would not be updated much in the future.
- `LuceneQuery_2_1` and `HighlightedText_2_1`  should provide the same behavior in the future for compatibility.
- `LuceneQuery` and `HighlightedText` could be changed much in the future.
- It would be easier to change when these classes are separated.

## Mini-chunk approach

Mini-chunks are placed at the windows of chunks. Figure 1 shows that.
fig1

In Figure 2, a hit is in the base part of chunk <span>#</span>2 (not window). This is the only chunk which has hits, so we should pick chunk <span>#</span>2.
fig2

Figure 3 is a particular version of Figure 2. The hit is on the border of the base and the window of chunk <span>#</span>2. As of Figure 2, we should pick chunk <span>#</span>2.
fig3

In Figure 4, a hit is in the base part of chunk <span>#</span>3, but in the window of chunk <span>#</span>2 at the same time. Both chunk <span>#</span>2 and <span>#</span>3 have hits. In this case, we can know the hit is in the window of chunk <span>#</span>2 because of mini-chunk <span>#</span>2 also have a hit. Hits in mini-chunks means hits are in windows. So we can pick chunk <span>#</span>3 only.
fig4

In Figure 5, chunk <span>#</span>3 has three hits, and mini-chunk <span>#</span>2 has two hits. We can calculate how many hits exist in chunk <span>#</span>2 by the following formula:
```
<hit count in the base chunk> = <hit count in the chunk> - <hit count in the mini-chunk>
```
In this case, it turns out that the base part of chunk <span>#</span>2 has 1 (= 3 - 2) hit.
So we can pick chunk <span>#</span>2, and chunk <span>#</span>3.
fig5


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedTextFactory.java
```diff
@@ -0,0 +1,26 @@
+package org.sleuthkit.autopsy.keywordsearch;
+
+import org.apache.commons.lang3.math.NumberUtils;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.TskCoreException;
+
+class HighlightedTextFactory {
```
bcarrier: Can you add JavaDocs for the purpose of the class and what its methods do?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedTextUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.google.common.collect.Range;
+import com.google.common.collect.TreeRangeSet;
+import org.apache.commons.lang.StringEscapeUtils;
+import org.apache.commons.lang.StringUtils;
+import org.apache.solr.common.SolrDocumentList;
+
+import java.util.Collection;
+
+public class HighlightedTextUtil {
```
bcarrier: This can probably be package-scope, right?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
```diff
@@ -1316,6 +1342,35 @@ public static String getChunkIdString(long parentID, int childID) {
         return Long.toString(parentID) + Server.CHUNK_ID_SEPARATOR + Integer.toString(childID);
     }
 
+    /**
+     * Return a language specific content field corresponding to the given Server#Schema#LANGUAGE field value
+     * @param language Server#Schema#LANGUAGE field value
+     * @return language specific content field
+     */
+    public static Server.Schema getContentFieldName(String language) {
```
bcarrier: Do these need to be public?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/LanguageDetector.java
```diff
@@ -0,0 +1,55 @@
...
+import com.optimaize.langdetect.text.CommonTextObjectFactories;
+import com.optimaize.langdetect.text.TextObject;
+import com.optimaize.langdetect.text.TextObjectFactory;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+public class LanguageDetector {
```
bcarrier: This can be package-scope, right?

bcarrier:
- We should continue to use 'content' and not change to 'content_general' to make backward compatibility easier or duplicate data.

akr4:
@bcarrier I created new PR #5139 and close this, because the code is changed a lot. I applied your indications to the new PR. Thanks.

PR #5065, opened 2019-07-23
1310 Added Update Message to TL

kellykelly3:


kellykelly3:
@raman-bt Was there a trace stack with that message?

rcordovano:
@kellykelly3, regarding the empty catch block that Codacy is complaining about,  use of getCurrentCaseThrows should not be necessary here. There is no way to open the TimelineTopComponent if there is no open case.

@wschaeferB and @bcarrier, the appveyor built seems to have failed because of a mismatch with the  TSK Java bindings, perhaps we have an issue with our branch matching?

wschaeferB:
It looks more like it is matching correctly getting the correct branch and then doing some stuff and then switching to develop for whatever reason.

wschaeferB:
It appears to be fetching timeline-event-mgr-2 then fetching develop then checking out timeline-event-mgr-2 then attempting to pull from the sleuthkit user's timeline-event-mgr-2 and deciding that the branch does not exist and pulling from develop instead.

Command run:git checkout -b origin-timeline-event-mgr-2
Command run:git pull https://github.com/sleuthkit/sleuthkit.git timeline-event-mgr-2
Branch: timeline-event-mgr-2 does not exist for github user: sleuthkit
Command run:git checkout -b origin-develop
Command run:git pull https://github.com/sleuthkit/sleuthkit.git develop

PR #5069, opened 2019-07-24
1303 - added code size table columns to the available space

kellykelly3:


Core/src/org/sleuthkit/autopsy/corecomponents/DataResultViewerTable.java
```diff
@@ -539,8 +549,43 @@ protected void setColumnWidths() {
...
+                    } else {
+                        extraWidth = (int) ((outlineViewWidth - totalColumnWidth)/columnWidths.size());
+                    }
+                }
+                
+                for(int column = 0; column < columnWidths.size(); column++) {
+                    int columnWidth = columnWidths.get(column);
+                    
+                    if(cntMaxSizeColumns > 0) {
+                        if(columnWidth >= 300) {
```
wschaeferB: Would we want to use the same int columnWidthLimit = (column == 0) ? 350 : 300; here so that when the first column size fell between was greater or equal to 300 but less than 350 we wouldn't wind up with a horizontal scroll bar by allocating it space we didn't account for in when counting columns that needed extra size and dividing the available extra size?

kellykelly3: I thought about that.  I could do that if you think it will make a difference.

wschaeferB: I don't know that it matters in the larger scheme of things, since sometimes we will have scroll bars and sometimes we won't either way.  Just caused me a moment while reading it to figure out if it would have any undesirable effects, which I don't think it will, but thought I would mention it because of that.

PR #5070, opened 2019-07-26
1291 deprecate index artifact

kellykelly3:


Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
```diff
@@ -48,13 +48,19 @@
...
      * @throws org.sleuthkit.datamodel.TskCoreException
      */
+    @Deprecated 
     public void indexArtifact(BlackboardArtifact artifact) throws TskCoreException;
 
     /**
-     * Add the given Content object to the text index.
+     * Add the given Content object to the text index.  This message should only
+     * be used in atypical cases, such as indexing a report.  Artifacts are indexed 
+     * when postArtifact called and files are indexed during ingest.
```
rcordovano: Missing an "is" between "postArtifact" and "called." I would also like to see postArtifact qualified with the class name. Perhaps "org.sleuthkit.autopsy.datamodel.Blackboard.postArtifact."

Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
```diff
@@ -48,13 +48,19 @@
      * all of its attributes.
      *
      * @param artifact The artifact to index.
+     * 
+     * @deprecated postArtifact org.sleuthkit.autopsy.datamodel.Blackboard will
+     * take care of this functionality at ingest
```
rcordovano: This seems a bit choppy: method name, class name. Perhaps simply say: "Call org.sleuthkit.autopsy.datamodel.Blackboard.postArtifact instead."

Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
```diff
@@ -48,13 +48,19 @@
      * all of its attributes.
      *
      * @param artifact The artifact to index.
+     * 
+     * @deprecated postArtifact org.sleuthkit.autopsy.datamodel.Blackboard will
```
bcarrier: this should reference the TSK datamodel, not Autopsy.

Core/src/org/sleuthkit/autopsy/keywordsearchservice/KeywordSearchService.java
rcordovano: I would still like to see postArtifact qualified here.

PR #5074, opened 2019-07-29
Text translation exported public API adjustments

rcordovano:


Core/src/org/sleuthkit/autopsy/texttranslation/TranslationOptionsPanel.java
```diff
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.sleuthkit.autopsy.texttranslation.ui;
+package org.sleuthkit.autopsy.texttranslation;
```
wschaeferB: These change will mean that the options panel and its constructor are now exported in our public API.

PR #5080, opened 2019-07-30
5118 cleanup sout calls

wschaeferB:


Core/src/org/sleuthkit/autopsy/test/InterestingArtifactCreatorIngestModule.java
```diff
@@ -128,7 +126,7 @@ public ProcessResult process(AbstractFile file) {
             logger.log(Level.SEVERE, String.format("Failed to process file (obj_id = %d)", file.getId()), ex);
             return ProcessResult.ERROR;
         } catch (Blackboard.BlackboardException ex) {
-            Exceptions.printStackTrace(ex);
+            logger.log(Level.WARNING, "Blackboard Exception processing file with obj_id = ", ex);
```
eugene7646: I think we need to add the actual obj_id, so that it's written to the error message here

eugene7646: There is only one minor outstanding item, everything else looks good.

wschaeferB: good catch will make that change in a few minutes

PR #5082, opened 2019-07-30
Text trans api mods for 4.12.0

rcordovano:


Core/src/org/sleuthkit/autopsy/texttranslation/TranslationOptionsPanelController.java
```diff
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.sleuthkit.autopsy.texttranslation.ui;
+package org.sleuthkit.autopsy.texttranslation;
```
wschaeferB: This change will cause TranslationOptionsPanelController to be exported as it is public.

Core/src/org/sleuthkit/autopsy/texttranslation/TranslationOptionsPanel.java
```diff
@@ -27,14 +27,12 @@
...
 import org.sleuthkit.autopsy.core.UserPreferences;
-import org.sleuthkit.autopsy.texttranslation.NoServiceProviderException;
-import org.sleuthkit.autopsy.texttranslation.TextTranslationService;
 import org.sleuthkit.autopsy.coreutils.Logger;
 
 /**
  * Options panel to display translation options
  */
-public class TranslationOptionsPanel extends javax.swing.JPanel {
+final class TranslationOptionsPanel extends javax.swing.JPanel {
```
wschaeferB: constructor now public in non-public class

Core/src/org/sleuthkit/autopsy/texttranslation/TextTranslator.java
```diff
@@ -19,6 +19,7 @@
 package org.sleuthkit.autopsy.texttranslation;
 
 import java.awt.Component;
```
wschaeferB: unused import

Core/src/org/sleuthkit/autopsy/texttranslation/TranslationOptionsPanelController.java
rcordovano: Yes, that is intended, as we have since discussed. Thanks for your vigilance!

Core/src/org/sleuthkit/autopsy/texttranslation/TextTranslator.java
rcordovano: Thanks, will be removed in next commit.

Core/src/org/sleuthkit/autopsy/texttranslation/TranslationOptionsPanel.java
rcordovano: Thanks, will be fixed in next commit.

PR #5083, opened 2019-07-31
2427 incorrect logger class widely used

markmckinnon:
Change the logger class from using java.util.logging.Logger to org.sleuthkit.autopsy.coreutils.Logger where appropriate.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/casemodule/datasourcesummary/DataSourceSummaryDetailsPanel.form
+++ b/Core/src/org/sleuthkit/autopsy/casemodule/datasourcesummary/DataSourceSummaryDetailsPanel.form
@@ -40,7 +40,7 @@
             <Component class="javax.swing.JLabel" name="operatingSystemLabel">
               <Properties>
                 <Property name="text" type="java.lang.String" editor="org.netbeans.modules.i18n.form.FormI18nStringEditor">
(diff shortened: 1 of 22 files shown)


Core/src/org/sleuthkit/autopsy/casemodule/datasourcesummary/DataSourceSummaryDetailsPanel.form
```diff
@@ -40,7 +40,7 @@
             <Component class="javax.swing.JLabel" name="operatingSystemLabel">
               <Properties>
                 <Property name="text" type="java.lang.String" editor="org.netbeans.modules.i18n.form.FormI18nStringEditor">
-                  <ResourceString bundle="org/sleuthkit/autopsy/casemodule/datasourcesummary/Bundle.properties" key="DataSourceSummaryDetailsPanel.operatingSystemLabel.text" replaceFormat="org.openide.util.NbBundle.getMessage({sourceFileName}.class, &quot;{key}&quot;)"/>
+                  <ResourceString bundle="org/sleuthkit/autopsy/casemodule/datasourceSummary/Bundle.properties" key="DataSourceSummaryDetailsPanel.operatingSystemLabel.text" replaceFormat="org.openide.util.NbBundle.getMessage({sourceFileName}.class, &quot;{key}&quot;)"/>
```
wschaeferB: I think we would want to go the other direction here and change any instances of the package having a capital S to having a lower case s.

wschaeferB:
A comment attached to the legitimate uses of the Java logger would be just saying that is is purposefully the java logger would be useful in my opinion.  
The following classes still use the Java logger and the first three I had no idea it was purposefully the java logger, except that I assume they are because they weren't changed which is why a comment added I think would be helpful.
PortableCaseInterestingItemsListPanel
PortableCaseTagsListPanel
TikaTextExtractor
OpenCvLoader

PR #5085, opened 2019-07-31
1292 Set the default state of the Filter checkboxes

kellykelly3:


Core/src/org/sleuthkit/autopsy/timeline/ui/filtering/datamodel/SqlFilterState.java
```diff
@@ -74,7 +64,6 @@ public String getDisplayName() {
 
     @Override
     public String toString() {
-        activeProp.get();
```
bcarrier: You can replace this with a call to something like getActiveProp() from AbstractFilterState.

kellykelly3: It's not needed.  That's why I removed it.

PR #5087, opened 2019-07-31
4595 central repo error

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
```diff
@@ -276,6 +276,7 @@ public Long getFileObjectId() {
          * @param supported   Is this Type currently supported
          * @param enabled     Is this Type currently enabled.
          */
+        @Messages({"CorrelationAttributeInstance.nullName.message=Database name is null."})
```
APriestman: Is this used?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/EamDbUtil.java
```diff
@@ -171,27 +171,60 @@ public static boolean schemaVersionIsSet(Connection conn) {
...
+        }
+        //get lock necessary for upgrade
+        if (db != null) {
+            try {
+                // This may return null if locking isn't supported, which is fine. It will
+                // throw an exception if locking is supported but we can't get the lock
+                // (meaning the database is in use by another user)
+                lock = db.getExclusiveMultiUserDbLock();
+            } catch (EamDbException ex) {
+                LOGGER.log(Level.SEVERE, "Error updating central repository, unable to acquire exclusive lock", ex);
```
APriestman: Should this continue with the upgrade after failing to get the lock?

wschaeferB: No, good catch we definitely shouldn't in my attempts to remove nested try catch blocks it looks like I didn't do it correct.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
wschaeferB: Nope I forgot to actually use it after I made it, will fix that.

rcordovano:
@wschaeferB, regarding the Codacy issue, I'd rather see a little code duplication that the instanceof use to which Codacy objects.

wschaeferB:
Alright will be multiple catch blocks when #5098 is merged after this story, it was initially an instanceof check here to be consistent with the existing instanceof check (which is also removed in #5098 
)

PR #5088, opened 2019-08-01
reorganized main page of user docs.  Removed duplicate info

bcarrier:


APriestman:
Should we add a link to the discourse page at the bottom of the main page?

bcarrier:
> Should we add a link to the discourse page at the bottom of the main page?

Added

PR #5098, opened 2019-08-02
5373 user friendly exception

wschaeferB:


Core/src/org/sleuthkit/autopsy/exceptions/AutopsyException.java
```diff
@@ -0,0 +1,84 @@
...
+    public AutopsyException(String message) {
+        super(message);
+        this.userMessage = message;
+    }
+
+    /**
+     * Constructs an exception to be thrown by a autopsy with a user exception.
+     *
+     * @param message     Exception message.
+     * @param userMessage the user friendly message to include in this exception
```
rcordovano: I like to see us use sentence case for parameter docs, i.e., capitalize "The" here, please.

Core/src/org/sleuthkit/autopsy/exceptions/AutopsyException.java
```diff
@@ -0,0 +1,84 @@
...
+ * An exception to be thrown which can contain a user friendly message.
+ */
+public abstract class AutopsyException extends Exception {
+
+    private static final long serialVersionUID = 1L;
+
+    private final String userMessage;
+
+    /**
+     * Constructs an exception to be thrown by a autopsy without a user message.
```
rcordovano: I think that instead of saying "to be thrown by an autopsy" we should focus instead on something like "Constructs an AutopsyException with identical exception and user-friendly messages" and so on.

Core/src/org/sleuthkit/autopsy/exceptions/AutopsyException.java
```diff
@@ -0,0 +1,84 @@
...
+    public AutopsyException(String message, String userMessage, Throwable cause) {
+        super(message, cause);
+        this.userMessage = userMessage;
+    }
+
+    /**
+     * Get the user friendly message if one exists.
+     *
+     * @return the user friendly message if one exists, otherwise returns the
+     *         exceptions normal message
```
rcordovano: "exceptions" should be "exception's."

wschaeferB:
@rcordovano  Would I be correct that the AutopsyExceptionHandler which you desired moved as part of this story should be moved back?

rcordovano:
@wschaeferB you are correct. Please take note @bcarrier, the uncaught exception handler AutopsyExceptionHandler is exported, public API. It lives in the org.sleuthkit.autopsy.coreutils with the following description: "Replaces default NetBeans exception handler. Displays messages in a dialog." Note that the new class AutopsyException has been deliberately placed in org.sleuthkit.autopsy.exceptions which is currently not exported, public API to allow us to "test" out the concept of an exception that provides both a developer message and a user-friendly message. Moving the uncaught exception handler which is only used by the application framework through its ServiceProvider annotation (@ServiceProvider(service = Handler.class, supersedes = "org.netbeans.core.NbErrorManager")) was originally deemed acceptable by me.

PR #5101, opened 2019-08-06
General report module (5354)

eugene7646:


Core/src/org/sleuthkit/autopsy/report/ReportProgressLogger.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Logs the status message.
+     *
+     * @param statusMessage String to use as label text.
+     */
+    @Override
+    public void updateStatusLabel(String statusMessage) {
+        if (status != ReportStatus.CANCELED) {
+            logger.log(Level.INFO, "Report Progress Logger status update: {0}", statusMessage);
```
rcordovano: I don't think that "Report Progress Logger status update:" will add anything to the logs, which will tell us the call site of the message. Logging the message from the report module should be sufficient.

eugene7646: @rcordovano Done.

Core/src/org/sleuthkit/autopsy/report/ReportProgressLogger.java
```diff
@@ -0,0 +1,157 @@
...
+            switch (reportStatus) {
+                case COMPLETE: {
+                    status = ReportStatus.COMPLETE;
+                    logger.log(Level.INFO, "Report completed");
+                    break;
+                }
+                case ERROR: {
+                    status = ReportStatus.ERROR;
+                    logger.log(Level.INFO, "Report completed with errors");
+                    break;
```
rcordovano: The generic nature of the completed, cancelled, etc., errors is not ideal for log entries. I'm thinking we could make the messages less generic if we pass a report display name to the constructor, then we could have messages such as "CASE-UCO report completed", etc.

Core/src/org/sleuthkit/autopsy/report/ReportProgressLogger.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Logs the status message.
+     *
+     * @param statusMessage String to use as label text.
+     */
+    @Override
+    public void updateStatusLabel(String statusMessage) {
+        if (status != ReportStatus.CANCELED) {
+            logger.log(Level.INFO, "Status update: {0}", statusMessage);
```
rcordovano: I think we can just log the message without the "Status update:" prefix.

eugene7646: @rcordovano Done

Core/src/org/sleuthkit/autopsy/report/ReportProgressLogger.java
eugene7646: @rcordovano That only works while we run a single report at a time. So I would cross that bridge when we get there. Perhaps for now we can just say "Reporting has completed" or "Report generation completed"?

rcordovano: @eugene7646, okay. It will become moot if we get to do the resolution of the technical debt we are adding for this story, anyway.

PR #5111, opened 2019-08-09
New easy-setup script for Linux/OSX, new README.mds

ethteck:


esaunders:
In case it wasn't clear from my referencing of this PR from Sleuthkit PR 1611 above...the approach to building on Travis has changed (more is being done in the travis.yml file and less is being done in travis_build.sh).
I notice that your new setup script relies on travis_build.sh. If you decide to move forward with this PR you will need to take the bits you care about from the old travis_build.sh script and plug them into your new script.

ethteck:
This is super low priority, but I just wanted to mark it as ready for review since I feel it's complete.

PR #5112, opened 2019-08-09
Report config loader 5355

eugene7646:


Core/src/org/sleuthkit/autopsy/report/TableReportSettings.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Class for persisting the selection of the tag types and artifact types used
+ * by the TableReportGenerator class to drive report generation by
+ * TableReportModules.
+ */
+class TableReportSettings implements Serializable  {
+
+    private static final long serialVersionUID = 1L;
+    private Map<BlackboardArtifact.Type, Boolean> artifactTypeSelections = new HashMap<>();
+    private Map<String, Boolean> tagNameSelections = new HashMap<>();
```
rcordovano: As for file report settings, I think that we can simplify these settings to Sets by only storing the enabled selections, and that we need some "all" flags. Also, remember that we will need a flag for whether or not to include custom artifact types, since we cannot know what they might be, at least in the elastic computing, command line ingest execution context.

Core/src/org/sleuthkit/autopsy/report/FileReportSettings.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+
+/**
+ * Class for persisting the file properties used by FileReportGenerator to drive
+ * report generation by FileReportModules.
+ */
+class FileReportSettings implements Serializable  {
+    
+    private static final long serialVersionUID = 1L;
+    private Map<FileReportDataTypes, Boolean> filePropertiesInfo = new HashMap<>();
```
rcordovano: I'm thinking that we can simplify this Map to a Set. The FileReportGenerator only needs to know which types are enabled, and a Set would provide an O(1) lookup as the generator iterated through the available types. I am also wondering if we will want an "all" flag like we will need for artifact types and tag types for the TableReportGenerator. A nice benefit of the Set and the "all " flag is that we can then easily accommodate new data types in future releases.

Core/src/org/sleuthkit/autopsy/report/ReportingConfig.java
```diff
@@ -0,0 +1,86 @@
...
+
+    TableReportSettings getTableReportSettings() {
+        return this.tableReportSettings;
+    }
+
+    void setFileReportSettings(FileReportSettings settings) {
+        this.fileReportSettings = settings;
+    }
+
+    FileReportSettings getFileReportModuleSettings() {
```
rcordovano: getFileReportSettings

Core/src/org/sleuthkit/autopsy/report/ReportGenerator.java
```diff
@@ -137,18 +136,15 @@ void generateGeneralReport(GeneralReportModule generalReportModule) throws IOExc
...
      *
-     * @param artifactTypeSelections the enabled/disabled state of the artifact
-     * types to be included in the report
-     * @param tagSelections the enabled/disabled state of the tag names to be
-     * included in the report
+     * @param tableReportSettings settings for the table report
      */
-    void generateTableReport(TableReportModule tableReport, Map<BlackboardArtifact.Type, Boolean> artifactTypeSelections, Map<String, Boolean> tagNameSelections) throws IOException {
-        if (tableReport != null && null != artifactTypeSelections) {
+    void generateTableReport(TableReportModule tableReport, TableReportSettings tableReportSettings) throws IOException {
```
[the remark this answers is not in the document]
eugene7646: That is correct. ReportGenerator refactoring and modifying the wizard to use it is story #7.

Core/src/org/sleuthkit/autopsy/report/ReportingConfigLoader.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Utility class responsible for managing serialization and deserialization of
+ * all of the settings that make up a reporting configuration in an atomic,
+ * thread safe way.
+ */
+class ReportingConfigLoader {
+
+    private static final String REPORT_CONFIG_FOLDER = "ReportingConfigs"; //NON-NLS
+    private static final String REPORT_CONFIG_FOLDER_PATH = Paths.get(PlatformUtil.getUserConfigDirectory(), ReportingConfigLoader.REPORT_CONFIG_FOLDER).toAbsolutePath().toString();
+    private static final String REPORT_CONFIG_FILE_NAME = "ReportingConfiguration.properties";
```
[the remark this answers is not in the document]
eugene7646: I like the idea described in bullet #2. I'll update the code to do that.

Core/src/org/sleuthkit/autopsy/report/TableReportSettings.java
eugene7646: I will copy/paste your note into the design document so that it is not lost. I think actual work on this should be done as part of the ReportGeneration refactoring story. I'm trying to make only minimal necessary code changes at this point to keep everything working "as before".

Core/src/org/sleuthkit/autopsy/report/FileReportSettings.java
eugene7646: Same as with TableReportSettings - I will copy/paste your note into the design document so that it is not lost. I think actual work on this should be done as part of the ReportGeneration refactoring story. I'm trying to make only minimal necessary code changes at this point to keep everything working "as before".

eugene7646:
@rcordovano All done.

PR #5113, opened 2019-08-12
App DB Parser helper - raw cut.

raman-bt:
This PR is to review and revise the design of the App DB parser helper class.


Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,485 @@
...
+     * It creates a DeviceAccount instance to use as a self account.
+     * 
+     * @param moduleName
+     * @param dbFile
+     * @param accountsType
+     * @param datasource
+     * 
+     * @throws TskCoreException 
+     */
+    public AppDBParserHelper(String moduleName, AbstractFile dbFile, Account.Type accountsType, DataSource datasource) throws TskCoreException {
```
bcarrier: We can get the datasource from AbstractFile.

raman-bt:
Closing this PR.

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
raman-bt: Of course.

InternalPythonModules/android/imo.py
```diff
@@ -0,0 +1,122 @@
...
+            selfAccountId = None
+            accountDbs = AppSQLiteDB.findAppDatabases(dataSource, "accountdb.db", "com.imo.android.imous")
+            for accountDb in accountDbs:
+                accountResultSet = accountDb.runQuery("SELECT uid, name FROM account")
+                if accountResultSet:
+                    # We can determine the IMO user ID of the device owner. 
+                    # Therefore we can create and use a app account and use that 
+                    # as a 'self' account instead of a Device account
+                    if not selfAccountId:
+                        selfAccountId = accountResultSet.getString("name")
```
dannysmyda: what if there are multiple accounts?

InternalPythonModules/android/imo.py
```diff
@@ -0,0 +1,122 @@
...
+                accountDb.close();
+                        
+            friendsDbs = AppSQLiteDB.findAppDatabases(dataSource, "imofriends.db", "com.imo.android.imous")
+            for friendsDb in friendsDbs:
+                friendsDBHelper = AppDBParserHelper("IMO Parser", friendsDb.getDBFile(),
+                                                    Account.Type.MESSAGING_APP, Account.Type.MESSAGING_APP, selfAccountId )
+                contactsResultSet = friendsDb.runQuery("SELECT buid, name FROM friends")
+                if contactsResultSet is not None:
+                    while contactsResultSet.next():
+                        friendsDBHelper.addContact( contactsResultSet.getString("name"),  ## name for account
```
dannysmyda: How is the name unique?

raman-bt: It's not.  That was one of the open questions we discussed today.  Although then we decided to create the account with the id, and attach the name as an attribute to it, since we don't want  to expose all the account creation/modification to the parsing module, we are going to defer that.

The name is more useful than the buid.

InternalPythonModules/android/imo.py
raman-bt: In theory, it's possible.  Given the table is named 'account',  and it has a single row with the name that I provided when set up the up (so presumably the phone owner's name), it's unlikely to find multiple accounts in this table.

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,485 @@
...
+import org.sleuthkit.datamodel.Relationship;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskDataException;
+
+
+/**
+ * A helper class to support modules that parse SQLite databases from mobile 
+ * apps and create artifacts.
+ */
+public final class AppDBParserHelper  {
```
dannysmyda: Can this be renamed to BlackboardUtility or something?

raman-bt: I'm happy to rename it to anything we all agree on.  I'm a little loathe to call it a "Utility" because it has too much state information - a "Utility" class in my mind is all static methods without any state.

How about BlackboardHelper?


@bcarrier  had also alluded to the possibility of moving this into SleuthKit instead.  Although currently I started using some util methods from CoreUtils (NetworkUtils.extractDomain() to extract domain from a URL).  If we do want to move this to Sleuthkit then I should refrain from using any Autopsy methods....

PR #5114, opened 2019-08-12
Implement backwards compatible fix for client closeable case services

rcordovano:
1. Restore implementation of Closeable by deprecated Blackboard class.
2. Make sure all Services class and all case service implementations that implement Closeable implement close() as a no-op.
3. Remove call to Services.close no-op from Case class.


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/SolrSearchService.java
```diff
@@ -238,10 +238,6 @@ public void deleteTextIndex(CaseMetadata metadata) throws KeywordSearchServiceEx
                 "SolrSearchService.exceptionMessage.noCurrentSolrCore"));
     }
 
-    @Override
```
dannysmyda: Close was not restored.

PR #5118, opened 2019-08-13
Fix TskDbDiff CSV Splitting

dannysmyda:
This PR fixes the CSV split for "INSERT INTO .... VALUES (a, b, c, d,)" statements. Our current approach is fooled if a or b is a string containing a comma.

This will fix regression failures on the tsk_event_descriptions table. However, this may introduce corrected diffs as this split routine was used by all INSERT statements.


dannysmyda:
This PR has been verified to make it through all our gold standards.

PR #5125, opened 2019-08-14
Added decodetect library for better encoding detection on text files

ethteck:
This fixes two issues: 

1. Tika sometimes gets the MIME-type incorrect for some text files.
2. Tika sometimes is unable to guess the encoding or does not even support the proper encoding for some text files.


ethteck:
I made those changes.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/KeywordSearchIngestModule.java
```diff
@@ -632,10 +634,15 @@ private void indexFile(AbstractFile aFile, boolean indexContent) {
                 if (context.fileIngestIsCancelled()) {
                     return;
                 }
-                if (fileType.equals("application/octet-stream")) {
+                if (fileType.equals(MimeTypes.OCTET_STREAM)) {
                     extractStringsAndIndex(aFile);
                     return;
                 }
+                if (fileType.equals(MimeTypes.PLAIN_TEXT)) {
```
bcarrier: I think we should now remove this because the same thing will now happen in extractTextAndIndex()  since TextExtractor is being returned from the Factory.  Keeping this out reduces a bit of complexity.

ethteck: Agreed, I went ahead and removed it.

PR #5128, opened 2019-08-16
5333: Create report module that updates text classifier

bhkjersten:
This report module trains a document classifier.
To train:
1. Create a case.
2. Import one or more data sources
3. Run the ingest module for file type identification on all files.
4. Label some of your documents as "notable": Right click the file -> Add file Tag -> Notable Item (Notable)
5. Click the "Generate Report" and select "Text classifier model"

This adds to the existing model.

Training  is much slower in Autopsy than in the standalone experiments. This is largely do to the time it takes to find the documents.

To run a model:
1. Import one or more data sources.
2. Run the ingest modules for "File Type Identification" and "Text Classifier" on the directories you want to classify.
3. Any documents that were classified as possibly notable are marked with a black triangle inside a yellow circle in the "S" column of the UI. You can also see them by clicking on the document, going to the "Results" tab, and checking whether Type="Set Name", Value="Possible notable text" is present in the table.


Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainer.java
```diff
@@ -0,0 +1,243 @@
...
+import org.sleuthkit.autopsy.report.ReportProgressPanel.ReportStatus;
+import org.sleuthkit.autopsy.textextractors.TextExtractor;
+import org.sleuthkit.autopsy.textextractors.TextExtractorFactory;
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.ContentTag;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskData;
+
+@ServiceProvider(service = GeneralReportModule.class)
+public class TextClassifierTrainer implements GeneralReportModule {
```
wschaeferB: I think we should extend GeneralReportModuleAdapter instead of implementing GeneralReportModule so that we can use the default implementations where applicable.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainer.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Override
+    public String getRelativeFilePath() {
+        return "classifierTrainerReport.txt";
+    }
+
+    @Override
+    public JPanel getConfigurationPanel() {
+        configPanel = new TextClassifierTrainerConfigPanel();
```
wschaeferB: I believe the default GeneralReportModuleAdapter implementation of returning null is what we would want here instead of an over ride returning a panel with nothing in it.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainerConfigPanel.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.experimental.textclassifier;
+
+import javax.swing.JPanel;
+
+class TextClassifierTrainerConfigPanel extends JPanel {
```
wschaeferB: I  believe this class is unnecessary.

bhkjersten: Is this what I should put in TextClassiferTrainer:
`private JPanel configPanel = new JPanel();`

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Classifies a file as notable or not notable, based on a model trained on
+ * labeled data.
+ */
+public class TextClassifierFileIngestModule extends FileIngestModuleAdapter {
+    private final static Logger logger = Logger.getLogger(TextClassifierFileIngestModule.class.getName());
+    private Blackboard blackboard;
```
wschaeferB: Deprecated use org.sleuthkit.datamodel.Blackboard instead

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+import opennlp.tools.doccat.DocumentCategorizerME;
+import org.openide.util.NbBundle.Messages;
+import org.sleuthkit.autopsy.casemodule.Case;
+import org.sleuthkit.autopsy.casemodule.NoCurrentCaseException;
+import org.sleuthkit.autopsy.casemodule.services.Blackboard;
+import org.sleuthkit.autopsy.coreutils.Logger;
+import org.sleuthkit.autopsy.ingest.FileIngestModuleAdapter;
+import org.sleuthkit.autopsy.ingest.IngestJobContext;
+import org.sleuthkit.autopsy.ingest.IngestModule;
+import org.sleuthkit.autopsy.ingest.IngestModuleReferenceCounter;
```
wschaeferB: Unused import

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+        utils = new TextClassifierUtils();
+
+        try {
+            categorizer = TextClassifierUtils.loadModel();
+        } catch (IOException ex) {
+            throw new IngestModule.IngestModuleException("Unable to load model for text classifier module.", ex);
+        }
+
+        try {
+            blackboard = Case.getCurrentCaseThrows().getServices().getBlackboard();
```
wschaeferB: deprecated use org.sleuthkit.autopsy.casemodule.getArtifactsBlackboard instead

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+        if (isNotable) {
+            try {
+                BlackboardArtifact artifact = file.newArtifact(BlackboardArtifact.ARTIFACT_TYPE.TSK_INTERESTING_FILE_HIT);
+                artifact.addAttribute(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SET_NAME,
+                        Bundle.TextClassifierModuleFactory_moduleName_text(),
+                        "Possible notable text"));
+                try {
+                    //Index the artifact for keyword search
+                    blackboard.indexArtifact(artifact);
+                } catch (Blackboard.BlackboardException ex) {
```
wschaeferB: indexArtifact and BlackboardException are depreicated

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        boolean isNotable;
+        try {
+            isNotable = classify(file);
+        } catch (TextExtractorFactory.NoTextExtractorFound ex) {
+            logger.log(Level.SEVERE, "NoTextExtractorFound in categorizing : " + ex.getMessage());
+            return ProcessResult.ERROR;
+        } catch (InitReaderException ex) {
+            logger.log(Level.SEVERE, "InitReaderException in categorizing : " + ex.getMessage());
```
wschaeferB: These catches should probably be doing logger.log(Level, String, Exception) instead of logger.log(Level, String) to preserve stack traces

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -0,0 +1,121 @@
...
+                        Bundle.TextClassifierModuleFactory_moduleName_text(),
+                        "Possible notable text"));
+                try {
+                    //Index the artifact for keyword search
+                    blackboard.indexArtifact(artifact);
+                } catch (Blackboard.BlackboardException ex) {
+                    logger.log(Level.SEVERE, "Unable to index blackboard artifact " + artifact.getArtifactID(), ex); //NON-NL
+                }
+            } catch (TskCoreException ex) {
+                logger.log(Level.SEVERE, "TskCoreException in categorizing : " + ex.getMessage());
```
wschaeferB: Same with these logger uses.

Experimental/nbproject/project.properties
```diff
@@ -2,6 +2,10 @@ file.reference.c3p0-0.9.5.jar=release/modules/ext/c3p0-0.9.5.jar
 file.reference.jackson-core-2.7.0.jar=release/modules/ext/jackson-core-2.7.0.jar
 file.reference.LGoodDatePicker-10.3.1.jar=release/modules/ext/LGoodDatePicker-10.3.1.jar
 file.reference.mchange-commons-java-0.2.9.jar=release/modules/ext/mchange-commons-java-0.2.9.jar
+file.reference.opennlp-tools-1.9.1-sources.jar=C:\\cygwin64\\home\\Brian Kjersten\\repos\\git\\autopsy\\Experimental\\release\\modules\\ext\\opennlp-tools-1.9.1-sources.jar
+file.reference.opennlp-tools-1.9.1.jar=C:\\cygwin64\\home\\Brian Kjersten\\repos\\git\\autopsy\\Experimental\\release\\modules\\ext\\opennlp-tools-1.9.1.jar
+file.reference.org.osgi.compendium-5.0.0.jar=C:\\cygwin64\\home\\Brian Kjersten\\repos\\git\\autopsy\\Experimental\\release\\modules\\ext\\org.osgi.compendium-5.0.0.jar
```
wschaeferB: Use relative paths

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainerConfigPanel.java
wschaeferB: I think either returning null or using the default implementation (which returns null) of GeneralReportModuleAdapter if you decide to extend it instead of implementing the same interface as it would be the way we typically indicate there is no panel for these.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/Bundle.properties-MERGED
```diff
@@ -0,0 +1,6 @@
+ClassifierTrainerReportModule.srcModuleName.txt=Text classifier model
+# {0} - classifierDir
+TextClassifierFileIngestModule.noClassifiersFound.message=No classifiers were found in {0}, text classifier will not be executed.
```
wschaeferB: There are several changes to the bundle properties and the latest version of this MERGED file should be included in the PR.

Core/src/org/sleuthkit/autopsy/coreutils/PlatformUtil.java
```diff
@@ -157,6 +158,15 @@ public static String getOcrLanguagePacksPath() {
     public static String getObjectDetectionClassifierPath() {
         return getUserDirectory().getAbsolutePath() + File.separator + CLASSIFIERS_SUBDIRECTORY;
     }
+    
+    /**
+     * Get root path where the user's text classifiers are stored.
+     * 
+     * @return Absolute path to the text classifiers root directory. 
+     */
+    public static String getTextClassifierPath() {
```
bcarrier: I'm not sure this shoudl be part of the public API....

bhkjersten: Okay. I moved this class to TextClassifierUtils. I also moved ensureTextClassifierFolderExists() from Installer to there. If ensureTextClassifierFolderExists() really needs to be in the Installer class, then getTextClassifierPath needs to be public.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/IncrementalNaiveBayesTrainer.java
```diff
@@ -0,0 +1,276 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This trains a naive Bayes classifier incrementally. It is modeled after the
+ * class NaiveBayseTrainer in OpenNLP. The difference is that it stores the
+ * model data in a Map that can easily be added to. It can upload a model from
+ * disk, add to it, and write to disk.
+ */
+public class IncrementalNaiveBayesTrainer extends AbstractEventTrainer {
```
bcarrier: This should be package scope and not public.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierFileIngestModule.java
```diff
@@ -61,129 +38,85 @@
...
             return IngestModule.ProcessResult.OK;
         }
-        
-        boolean isInteresting;
+
+        boolean isNotable;
         try {
-            isInteresting = classify(file);
+            isNotable = classify(file);
         } catch (TextExtractorFactory.NoTextExtractorFound ex) {
```
bcarrier: We can probably do a multi-catch with this since we do the same thing each time.  If we log the exception, then we'll know what type it was.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainer.java
```diff
@@ -0,0 +1,256 @@
...
+        + "2. add the file tag \"Notable item (notable)\" to documents you "
+        + "know are notable.\n"
+        + "Once you've trained a model, you can classify documents by "
+        + "running the \"Text Classifier\" Ingest Module."})
+    @Override
+    public String getDescription() {
+        return NbBundle.getMessage(this.getClass(), "TextClassifierTrainer.getDesc.text");
+    }
+
+    public DoccatModel train(String oldModelPath, ObjectStream<DocumentSample> sampleStream) throws IOException {
```
bcarrier: Should be private.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierTrainer.java
```diff
@@ -0,0 +1,256 @@
...
+                label = TextClassifierUtils.NONNOTABLE_LABEL;
+                containsNonNotableDocument = true;
+            }
+            DocumentSample docSample = new DocumentSample(label, TextClassifierUtils.extractTokens(doc));
+            docSamples.add(docSample);
+
+            progressPanel.increment();
+        }
+
+        if (!containsNotableDocument) {
```
bcarrier: Why is this a requirement?  There could be cases where all documents were OK or Notable.

bhkjersten: This requirement is an oversimplification of the following two constraints:

1. OpenNLP will not train a model unless at least two categories occur in the training data. This does not have to mean that every batch contains both categories. 
2. IncrementalNaiveBayesTrainer.mergeInParameters() was written assuming that all of the categories ("outcomes" in OpenNLP's lingo) occurred in the first batch.

What this means is that the first batch needs to contain both classes. I'll change the code to use this requirement instead. I'll also look into whether I can get it to work even if the first batch is all of the same category.

PR #5130, opened 2019-08-19
5374 case uco portable case

dannysmyda:
This PR has only been partially tested. The final interaction with the PortableCaseReportModule remains untested due to Jira-5425


dannysmyda:
Review on this PR can proceed @eugene7646

PR #5131, opened 2019-08-19
Fix bug in tskdbdiff

dannysmyda:


test/script/tskdbdiff.py
```diff
@@ -437,7 +437,27 @@ def normalize_db_entry(line, files_table, vs_parts_table, vs_info_table, fs_info
...
+    fields_list = list(csv.reader([no_space_parens], quotechar="'"))[0]
+    #Add back in the quotechar for values that were originally wrapped (csv reader consumes this character)
+    fields_list_with_quotes = []
+    ptr = 0
+    for field in fields_list:
+        if(len(field) == 0):
+            field = "'" + field + "'"
+        else:
+            start = no_space_parens.find(field, ptr)
+            if(start != -1):
```
wschaeferB: I believe this condition is unnecessary as start - 1 will never be greater than or equal to 0 if start was equal to -1

PR #5135, opened 2019-08-22
1324 Make deleted files for recycled files

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+        long getDeletedTimeStamp() {
+            return deletedTimeStamp;
+        }
+
+        /**
+         * Returns the name of the deleted file.
+         * 
+         * @return String name of the deleted file
+         */
+        String getFileName() {
```
bcarrier: Is this full path (including "C:\") or just the final file name?  We should clarify and perhaps rename.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+            AbstractFile parent = getOrMakeFolder(skCase, dataSource, parentPath); 
+            return skCase.addVirtualDirectory(parent.getId(), folderName);
+           
+        } else {
+            return files.get(0);
+        }
+    }
+    
+    private void addFileToDirectory (SleuthkitCase skCase, AbstractFile rFile, Content parentDir, String fileName, long deletedTime) throws TskCoreException {
+        FsContent fsContent = (FsContent)rFile;
```
bcarrier: We shoudl either test before casting or simply make the argument be of type FsContent.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+        
+        String parentPath = getParentPath(path);
+        String folderName = getFileName(path);
+        
+        List<AbstractFile> files = null;
+        if(parentPath != null) {
+            if(!parentPath.equals("/")) {
+                parentPath = parentPath + "/";
+            }
+            files = skCase.findAllFilesWhere(String.format("fs_obj_id=%s AND parent_path='%s' AND name='%s'", dataSource.getFileSystemId(), parentPath, folderName != null ? folderName : ""));
```
bcarrier: We need to escape parent_path and name in case they have ' in them.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+    private BlackboardArtifact createArtifact(AbstractFile rFile, BlackboardArtifact.Type type, String fileName, String userName, long dateTime) throws TskCoreException{
+        BlackboardArtifact bba = rFile.newArtifact(type.getTypeID());
+        bba.addAttribute(new BlackboardAttribute(TSK_PATH, getName(), fileName));
+        bba.addAttribute(new BlackboardAttribute(TSK_DATETIME_DELETED, getName(), dateTime));
+        bba.addAttribute(new BlackboardAttribute(TSK_USER_NAME, getName(), userName == null || userName.isEmpty() ? "" : userName));
+        
+        return bba;
+    }
+    
+    private AbstractFile getOrMakeFolder(SleuthkitCase skCase, FsContent dataSource, String path) throws TskCoreException {
```
bcarrier: Could have FileSystem be the argument type.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+     * Process each individual iFile.
+     * 
+     * @param dataSource
+     * @param context
+     * @param recycleBinArtifactType Module created artifact type
+     * @param iFile The AbstractFile to process
+     * @param userNameMap Map of user ids to names
+     * @param tempRARecycleBinPath Temp directory path
+     */
+    private void processIFiles(Content dataSource, IngestJobContext context,  BlackboardArtifact.Type recycleBinArtifactType, AbstractFile iFile,  Map<String, String> userNameMap, Map<String, List<AbstractFile>> rFileMap, String tempRARecycleBinPath) {
```
bcarrier: Method name should be singular (processIFile)

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+        String tempFilePath = tempRARecycleBinPath + File.separator + iFile.getName();
```
bcarrier: We should make this more unique by adding an ID to the end of it. Could have naming collisions between volumes on the same data source.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: This is just the last bit of the path.  
If
C:\Temp\Directory\Folder2
Folder2 will be returned

If 
C:\Temp\Directory\Folder2\file.txt
file.txt will be returned.

I have added a proper function header with the description, but I was staring at the function trying to decide whether it needed a better name.  The current name was mirroring the function from Path.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+    /**
+     * 
+     * @param skCase
+     * @param folder
+     * @param children
+     * @param parentPath
+     * @param deletedTimeStamp
+     * @throws TskCoreException 
+     */
+    private void popuplateDirectory(SleuthkitCase skCase, AbstractFile folder, List<Content> recycledChildren, String parentPath, long deletedTimeStamp) throws TskCoreException {
```
bcarrier: I'd add deleted in here somewhere, something like populateDeletedDirectory()

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: Agreed, I have already changed this.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+            return;
+        }
+        
+        for(Content child: recycledChildren) {
+            if(child instanceof AbstractFile) {
+                AbstractFile fileChild = (AbstractFile) child;
+                if(fileChild.isFile()) {
+                    addFileToDirectory(skCase, fileChild, folder, fileChild.getName(), deletedTimeStamp);
+                } else if(fileChild.isDir()) {
+                    String newPath = parentPath + File.separator + fileChild.getName();
```
bcarrier: You should probably just hard code in "/" since that is what the DB uses and not rely on Java's, which I think is platform dependent.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: In case they have what in them?  Is that an apostrophe?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: You want the function to return an FsContent?  I can do that.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -0,0 +1,552 @@
...
+        if(files == null || files.isEmpty()) {
+            AbstractFile parent = getOrMakeFolder(skCase, dataSource, parentPath); 
+            return skCase.addVirtualDirectory(parent.getId(), folderName);
+           
+        } else {
+            return files.get(0);
+        }
+    }
+    
+    private void addFileToDirectory (SleuthkitCase skCase, AbstractFile rFile, Content parentDir, String fileName, long deletedTime) throws TskCoreException {
```
bcarrier: Shoudl have "Deleted" somewhere in here.  Such as addDeletedFileToDirectory()

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: Yes.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: Now that you say that it should actually be the "\" because that path is going to be a windows absolute path from the $I file.  getOrMake will take care of correcting the slashes.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
bcarrier: Yea.

https://github.com/sleuthkit/sleuthkit/blob/develop/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java#L8835

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
bcarrier: No, I meant that 'dataSource' could be of type FileSystem instead of FsContent. All we do with that argument is call getFileSystem() on it.

kellykelly3: SleuthkitCase.addVirtualDirectory returns a VirtualDirectory which extends SpecialDirectory which extends AbstractFile.  

The code its self doesn't need the return value from addOrMakeDirectory to be an FsContent.

kellykelly3: The dataSource that I pass in is the original %R file which is an AbstractFile and cannot be cast to a FileSystem.  Is there away to get a FileSystem from an AbstractFile?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
kellykelly3: Good catch, thank you.  I wouldn't have thought of that one.

PR #5139, opened 2019-08-26
Japanese search feature

akr4:
This PR adds a Japanese text search feature, which uses `text_ja` type in Solr and provides better results.

# Changes

## Library dependency
- Added [language-detector](https://github.com/optimaize/language-detector)

## Solr schema
- Added `text_ja` type
- Added `language`, `content_ja` fields
- Updated schema version from 2.1 to 2.2.

## Index and query
The new process of index and query:
1. `Ingester` detects languages
2. `Ingester` indexes chunks and mini-chunks
  - If the language is not Japanese, indexing is done in the same way as the current released version
  - If the language is Japanese
    - Do the same things as for non-Japanese documents
    - Store the content and the filename in the `content_ja` field
    - Store the language (`"ja"`) in the `language` field
    - Index mini-chunks

3. `LuceneQuery` and `HighlightedText` uses `text` and `content_ja` for query and highlighting.

## Mini-chunk approach

Mini-chunks are placed at the windows of chunks. Figure 1 shows that.
fig1

In Figure 2, a hit is in the base part of chunk <span>#</span>2 (not window). This is the only chunk which has hits, so we should pick chunk <span>#</span>2.
fig2

Figure 3 is a particular version of Figure 2. The hit is on the border of the base and the window of chunk <span>#</span>2. As of Figure 2, we should pick chunk <span>#</span>2.
fig3

In Figure 4, a hit is in the base part of chunk <span>#</span>3, but in the window of chunk <span>#</span>2 at the same time. Both chunk <span>#</span>2 and <span>#</span>3 have hits. In this case, we can know the hit is in the window of chunk <span>#</span>2 because of mini-chunk <span>#</span>2 also have a hit. Hits in mini-chunks means hits are in windows. So we can pick chunk <span>#</span>3 only.
fig4

In Figure 5, chunk <span>#</span>3 has three hits, and mini-chunk <span>#</span>2 has two hits. We can calculate how many hits exist in chunk <span>#</span>2 by the following formula:
```
<hit count in the base chunk> = <hit count in the chunk> - <hit count in the mini-chunk>
```
In this case, it turns out that the base part of chunk <span>#</span>2 has 1 (= 3 - 2) hit.
So we can pick chunk <span>#</span>2, and chunk <span>#</span>3.
fig5

## Termfreq
To determine whether or not chunks should be picked up in a search result, This code uses `termfreq` function in Solr query.
```
sum(
  termfreq("content_ja","term1"),
  termfreq("content_ja","term2")
)
```
`term1` and `term2` are terms included in a query. For example, 雨が降っ ている (It's raining) is divided into two terms 雨 and 降る. If your query has some terms, `termfreq` should be used for each term. So knowing these terms before the query is required. That is achieved by making an additional request to Solr. `QueryParse` is the class for that.


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
```diff
@@ -359,16 +362,33 @@ public String getText() {
...
-                final String highlightQuery = keywords.stream()
-                        .map(HighlightedText::constructEscapedSolrQuery)
-                        .collect(Collectors.joining(" "));
-
-                q.setQuery(highlightQuery);
-                q.addField(highlightField);
-                q.addFilterQuery(filterQuery);
-                q.addHighlightField(highlightField);
-                q.setHighlightFragsize(0); // don't fragment the highlight, works with original highlighter, or needs "single" list builder with FVH
+                if (2.2 <= indexSchemaVersion) {
```
esaunders: I see references to schema version 2.2 but the schema.xml file has not been updated.

esaunders:
Highlighting of English text does not appear to work correctly. I added the attached eng_utf16l.txt file as a Logical File and ran an exact match search for the string "self-treatment". Autopsy 4.12 correctly highlights 2 hits in the same document whereas the code on this branch results in 7 hits (see attachments).
autopsy-4-12-eng-search-highlighting
[eng_utf16l.txt](https://github.com/sleuthkit/autopsy/files/3591861/eng_utf16l.txt)

esaunders:
New source files are missing copyright/license header block.

esaunders:
When I index the attached file and search for 東京都 Autopsy reports 2 matches but I only see highlighting for a single match.
jp-wikipedia-search-highlighting
[Japanese.Wikipedia.utf8.txt](https://github.com/sleuthkit/autopsy/files/3592112/Japanese.Wikipedia.utf8.txt)

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
```diff
@@ -175,6 +188,12 @@ public String toString() {
             public String toString() {
                 return "chunk_size"; //NON-NLS
             }
+        },
+        TERMFREQ {
```
esaunders: This doesn't appear to be a field in the schema.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
akr4: Thank you for your review.
I updated `KeywordSearch/release/solr/solr/configsets/AutopsyConfig/conf/schema.xml`. Isn't that sufficient?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
akr4: `termfreq` is a function to get the number of occurrence of the term. It's not an actual field in indices, but it can be gotten from documents in the same way as fields.
```document.getFieldValue(Server.Schema.TERMFREQ.toString())```
Should `Server.Schema` have actual fields only?

**Comment by akr4** [comment]
@esaunders Thank you for your review. I will investigate the detail of "self-treatment" and 東京都 issues, and get back to you.

File: KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java (hunk 1 above)
Comment (esaunders) [reply KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java]: Yes, but I don't see the change in schema.xml above.

**Comment by esaunders** [comment]
I started my testing using a small data set that I have shared with you through another channel and I noticed the following...

The Chinese documents are being identified as Japanese (perhaps not so surprising) and the Japanese documents are not being identified as Japanese (perhaps the word "NOVA" at the beginning of the document is throwing language detection off or maybe there isn't enough content for the language detector to work with).

I understand that language detection can be difficult (especially with a small data set) but it might be worthwhile taking a look into this to see if there is anything we can do about it.

File: KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/LanguageDetector.java [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+import com.optimaize.langdetect.text.TextObjectFactory;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+class LanguageDetector {
+
+  public enum Language {
```
esaunders: I find the existence of 2 different enums named "Language" to be confusing. Is it possible to consolidate these enums and possible use Enum.name() and toString() methods in place of static strings (e.g. "en" and "ja" below)?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/LanguageDetector.java
```diff
@@ -0,0 +1,55 @@
...
+        .withProfiles(languageProfiles)
+        .build();
+
+    TextObjectFactory textObjectFactory = CommonTextObjectFactories.forDetectingOnLargeText();
+
+    TextObject textObject = textObjectFactory.forText(text);
+    return languageDetector.detect(textObject).transform(Optional::of).or(Optional.empty()).map(LdLocale::getLanguage).flatMap(LanguageDetector::toLanguage);
+  }
+
+  private static Optional<Language> toLanguage(String s) {
```
esaunders: Perhaps this can be replaced by something else in the consolidated enum I suggest above. There are a number of different ways you could do this, e.g. you might leverage Enum.valueOf().

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/QueryParser.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    Map<String, List<String>> fieldTermsMap;
+  }
+
+  /**
+   * Parse the given query string on Solr and return the result
+   */
+  static Result parse(String query, List<Server.Schema> fields) throws KeywordSearchModuleException, NoOpenCoreException {
+    SolrQuery q = new SolrQuery();
+    q.setShowDebugInfo(true);
```
esaunders: Does debug info really need to be turned on here?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
```diff
@@ -359,16 +362,33 @@ public String getText() {
...
+                        .collect(Collectors.joining(" OR "));
+                    q.setQuery(highlightQuery);
+                    for (Server.Schema field : LanguageSpecificContentQueryHelper.getQueryFields()) {
+                        q.addField(field.toString());
+                        q.addHighlightField(field.toString());
+                    }
+                    q.addField(Server.Schema.LANGUAGE.toString());
+                    // in case of single term literal query there is only 1 term
+                    LanguageSpecificContentQueryHelper.configureTermfreqQuery(q, keywords.iterator().next());
+                    q.addFilterQuery(filterQuery);
```
esaunders: These 2 lines appear to be the same in both the "if" and "else" blocks. Perhaps they should be moved out of the if..else.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
```diff
@@ -406,12 +426,40 @@ public String getText() {
                 if (responseHighlightID == null) {
                     highlightedContent = attemptManualHighlighting(response.getResults(), highlightField, keywords);
                 } else {
-                    List<String> contentHighlights = responseHighlightID.get(LuceneQuery.HIGHLIGHT_FIELD);
-                    if (contentHighlights == null) {
-                        highlightedContent = attemptManualHighlighting(response.getResults(), highlightField, keywords);
+                    SolrDocument document = response.getResults().get(0);
```
esaunders: It might make it a little easier to read this method if this block of code related to mini chunk highlighting was factored out into its own method.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
```diff
@@ -406,12 +426,40 @@ public String getText() {
...
+                    Object language = document.getFieldValue(Server.Schema.LANGUAGE.toString());
+                    if (2.2 <= indexSchemaVersion && language != null) {
```
esaunders: If language != null do we really need the schema version check?

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
esaunders: If it's coming back as a field in the document I'm fine with leaving it where it is. I would suggest that you add a comment to avoid confusion in the future if someone is looking at the code and wondering why there is no field with that name in the schema :-)

akr4:
About `self-treatment` issue, I had changed not using phrase queries by mistake. As a result of that, `self-treatment` matches `self` and `treatment` respectively. I fixed this problem in my environment. I will push once finishing the 東京都 issue.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/HighlightedText.java
akr4: The path in my previous comment is wrong. The correct path is `KeywordSearch/solr/solr/configsets/AutopsyConfig/conf/schema.xml
`. 
Diff:
https://github.com/sleuthkit/autopsy/pull/5139/files#diff-3f316c678543ca04318ac83da0a094d8

esaunders: When I click on the link you provided (on this side of the world :-) I see the changes to add the text_ja type and the content_ja and language fields but I do not see any update to the schema version. When I look at the entire file containing your changes I see the following:

<schema name="Autopsy Keyword Search" version="2.1">

What do you see?

akr4: I noticed that Autopsy is using the version attribute of `<schema>` tag. I will update it later. Thank you for pointing it out.

akr4:
I pushed new commits. This inclue:
- Enabled phrase query ("self-treatment" query now works fine)
- Use `original` highlighter for Japanese text
- Updated the version of `<schema>` tag.
- Added copyright/license block
- Added a comment on `TERMFREQ`
- Consolidated the Language classes

Now "self-treatment" issue is fixed by using phrase queries. However, using phrase queries on Japanese texts caused another problem. It seems that `FastVectorHighlighter` is not work in some cases. So I changed the code to use `original` highlighter when `language` value is `ja`.

Even with these changes, 東京都 issue still exists.
東京都 is divided into 東京 and 都 before generating snippets and matches individually. It looks like a single match, but two matches 東京 and 都 happen consecutively. 
New `unified` highlighter (from Solr 6.4) handles this case as expected. Once we migrate to Solr 8, I think it's good to consider to use it.
So for now, as a workaround, it would be possible to change the style of the highlighted words to show there're multiple matches.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/Server.java
akr4: I added a comment on `TERMFREQ`. Thank you for your suggestion.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/LanguageDetector.java
akr4: I consolidated the two `Language` classes. Now there is only `JAPANESE`.  A `Language` is detected means the document should be indexed to a language-specific field such as `content_ja`.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/QueryParser.java
akr4: Yes. I use the debug flag to get parsed query information here.
https://lucene.apache.org/solr/7_5_0//solr-core/org/apache/solr/handler/FieldAnalysisRequestHandler.html
But there is the non-debug API `FieldAnalysisRequestHandler` for the same purpose. I will come up with a commit which uses it.

esaunders:
Thanks for the update. At this point I'm waiting until this pull request is as close as possible to a "done" state before I take a look at it again. Let me know when you get to a point where you believe all of feedback has been addressed (including Codacy issues).

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/QueryParser.java
akr4: Changed to use `FieldAnalysisRequestHandler`. The result is the same but it's a official API (not debug).

akr4:
I pushed some commits:
- I updated `LanguageDetector` to use all language profiles which the implementation provides. I saw it now can detect Chinese text correctly in my environment.
- Use `FieldAnalysisRequestHandler` to parse queries instead of the debug info.

I will work on the Codacy result later.

akr4:
@esaunders Hi. Now I finished all the things to do.
appveyor and travis-ci still failed, but it seems not to be related to these changes.

akr4:
@esaunders Thank you for your review. What is the next step for this PR to be merged?

rcordovano:
@Akr4, @esaunders, and @bcarrier, I will merge this into the develop branch now. It will not be run through automated regression testing until next week, since we are currently using the testing infrastructure to validate the Autopsy 4.13.0 release.

akr4:
@rcordovano Thank you for the information.

rcordovano:
You are welcome, @akr4.

PR #5140, opened 2019-08-26
1324 recycle bin changes

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -498,16 +498,18 @@ private AbstractFile getOrMakeFolder(SleuthkitCase skCase, FsContent dataSource,
...
+                fileName,
+                recycleBinFile.getMetaAddr(),
+                (int) recycleBinFile.getMetaSeq(),
                 recycleBinFile.getAttrType(),
                 recycleBinFile.getAttributeId(),
-                fileName,
                 TskData.TSK_FS_NAME_FLAG_ENUM.UNALLOC,
+                (short) (TskData.TSK_FS_META_FLAG_ENUM.UNALLOC.getValue() | TskData.TSK_FS_META_FLAG_ENUM.USED.getValue()),
                 recycleBinFile.getSize(),
                 0, 0, 0, deletedTime,
```
markmckinnon: Can you add in the Timestamps from the $R file when you add it back in.  This keeps the file consistent with the $R file.

kellykelly3: This is not as straight forward as you might think.  Those values can only be added for only an $R that represents a deleted file.  Not all files.

kellykelly3: Fixed.

PR #5143, opened 2019-08-27
5369 show duplicates ui

wschaeferB:


wschaeferB:
The codacy issues on this PR which were not showing up for a time, are fixed in PR: https://github.com/sleuthkit/autopsy/pull/5155  (there is 1 more PR between this and that)

PR #5152, opened 2019-09-04
5394: Implement utility class supporting mobile app parsers and first…

raman-bt:
… parser

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
+++ b/Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
@@ -1,0 +1,1057 @@
+/*
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2019 Basis Technology Corp.
+ * Contact: carrier <at> sleuthkit <dot> org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.coreutils;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.logging.Level;
+import org.apache.commons.lang3.StringUtils;
+import org.sleuthkit.autopsy.casemodule.Case;
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.AccountFileInstance;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.BlackboardArtifact.ARTIFACT_TYPE;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.BlackboardAttribute.ATTRIBUTE_TYPE;
+import org.sleuthkit.datamodel.Account;
+import org.sleuthkit.datamodel.Blackboard;
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.Relationship;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskDataException;
+
+
+/**
+ * A helper class to support modules that parse SQLite databases from mobile 
+ * apps and create artifacts.
+ */
+public final class AppDBParserHelper  {
+    
+    private static final Logger logger = Logger.getLogger(AppDBParserHelper.class.getName());
+    
+    private final AbstractFile dbAbstractFile;
+    private final String moduleName;
+    
+    // 'self' account for the application. 
+    private final AccountFileInstance selfAccountInstance;
+    
+    // type of accounts to be created for the Application using this helper
+    private final Account.Type accountsType;
+    
+    /**
+     * Constructs a AppDB parser helper for the given DB file.
+     * 
+     * This is a constructor for Apps that that do not have any app specific account information 
+     * for device owner and will use  a 'Device' account in lieu.
+     * 
+     * It creates a DeviceAccount instance to use as a self account.
+     * 
+     * @param moduleName name module using the helper
+     * @param dbFile database file being parsed by the module
+     * @param accountsType account types created by this module
+     * 
+     * @throws TskCoreException 
+     */
+    public AppDBParserHelper(String moduleName, AbstractFile dbFile, Account.Type accountsType) throws TskCoreException {
+        
+        this.moduleName = moduleName;
+        this.dbAbstractFile = dbFile;
+        this.accountsType = accountsType;
+        this.selfAccountInstance = Case.getCurrentCase().getSleuthkitCase().getCommunicationsManager().createAccountFileInstance(Account.Type.DEVICE, ((DataSource)dbFile.getDataSource()).getDeviceId(), moduleName, dbFile);
+    }
+    
+    /**
+     * Constructs a AppDB parser helper for the given DB file.
+     * 
+     * This constructor is for Apps that do have app specific account information 
+     * for the device owner to create a 'self' account.
+     * 
+     * It creates a an account instance with specified type & id and uses it as
+     * a self account.
+     * 
+     * @param moduleName name module using the helper
+     * @param dbFile database file being parsed by the module
+     * @param accountsType account types created by this module
+     * @param selfAccountType self account type to be created for this module
+     * @param selfAccountId account unique id for the self account
+     * 
+     * @throws TskCoreException 
+     */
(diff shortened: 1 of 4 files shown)


Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
```diff
@@ -0,0 +1,249 @@
...
+        String localDiskPath = "";
+        try {
+            SleuthkitCase skCase = openCase.getSleuthkitCase();
+            String parentPath = parentPathSubstr.replace("\\", "/");
+            parentPath = SleuthkitCase.escapeSingleQuotes(parentPath);
+            String whereClause;
+            if (matchExactName) {
+                whereClause = String.format("LOWER(name) LIKE LOWER(\'%1$s\') AND LOWER(parent_path) LIKE LOWER(\'%%%2$s%%\') AND data_source_obj_id = " + dataSource.getId(), dbName, parentPath);
+            } else {
+                whereClause = String.format("LOWER(name) LIKE LOWER(\'%%%1$s%%\') AND LOWER(name) NOT LIKE LOWER(\'%%journal%%\') AND LOWER(parent_path) LIKE LOWER(\'%%%2$s%%\') AND data_source_obj_id = " + dataSource.getId(), dbName, parentPath);
```
dannysmyda: What example did you come across where you needed to wild card the database name?

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,1145 @@
...
+     * @param accountsType account types created by this module
+     * 
+     * @throws TskCoreException 
+     */
+    public AppDBParserHelper(String moduleName, AbstractFile dbFile, Account.Type accountsType) throws TskCoreException {
+        
+        this.moduleName = moduleName;
+        this.dbAbstractFile = dbFile;
+        this.accountsType = accountsType;
+        this.selfAccountInstance = Case.getCurrentCase().getSleuthkitCase().getCommunicationsManager().createAccountFileInstance(Account.Type.DEVICE, ((DataSource)dbFile.getDataSource()).getDeviceId(), moduleName, dbFile);
```
dannysmyda: What happens if the user runs into a TskCoreException here? Is the idea that they skip it or would it be better if they followed up with a call to "public AppDBParserHelper(String moduleName, AbstractFile dbFile)", where this value is just null? If its the latter how would they know?

InternalPythonModules/android/imo.py
```diff
@@ -0,0 +1,139 @@
...
+                                                            msgReadStatus,
+                                                            "",     # subject
+                                                            messagesResultSet.getString("last_message"),
+                                                            "")   # thread id
+                                                                                                
+                        # TBD: parse the imdata JSON structure to figure out if there is an attachment.
+                        #      If one exists, add the attachment as a derived file and a child of the message artifact.
+
+            
+            except SQLException as ex:
```
dannysmyda: Missing catch to TskCoreException.

InternalPythonModules/android/imo.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                accountResultSet = accountDb.runQuery("SELECT uid, name FROM account")
+                if accountResultSet:
+                    # We can determine the IMO user ID of the device owner. 
+                    # Therefore we can create and use a app account and use that 
+                    # as a 'self' account instead of a Device account
+                    if not selfAccountId:
+                        selfAccountId = accountResultSet.getString("name")
+            
+            except SQLException as ex:
```
dannysmyda: Missing catch to TskCoreException.

Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
```diff
@@ -142,6 +142,101 @@ public AbstractFile getDBFile() {
...
+     * @throws TskCoreException 
+     */
+    public void detachDatabase(String dbAlias) throws TskCoreException {
+       
+        try {
+            String detachDbSql = String.format("DETACH DATABASE '%s'", dbAlias);
+            statement.executeUpdate(detachDbSql); //NON-NLS
+        }
+        catch (SQLException ex) {
+            throw new TskCoreException("Error running DETACH DATABASE SQL. " + ex.getMessage(), ex);
```
dannysmyda: Why TskCoreException?

Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
```diff
@@ -142,6 +142,101 @@ public AbstractFile getDBFile() {
...
+                    findAndCopySQLiteMetaFile(absFile, absFile.getName() + "-wal");
+                    findAndCopySQLiteMetaFile(absFile, absFile.getName() + "-shm");
+                
+                    //run the ATTACH DATABASE sql command
+                    try {
+                        String attachDbSql = String.format("ATTACH DATABASE '%s' AS '%s'", localFilePath, dbAlias); //NON-NLS
+                        statement.executeUpdate(attachDbSql); 
+                    }
+                    catch (SQLException ex) {
+                        throw new TskCoreException("Error running ATTACH DATABASE SQL. " + ex.getMessage(), ex);
```
dannysmyda: Why TskCoreException?

Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
raman-bt: Skype database was named live:btforensics.db (i think) where btforensics is the profile name of the user.

Actually as it turns out, Autopsy did not correctly export that file so it isn't in the logical file data source.

Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
```diff
@@ -0,0 +1,328 @@
...
+     * 
+     * A list of AppSQLiteDB instances is returned, one for each 
+     * match found., 
+     * .
+     * @param dataSource data source to search in 
+     * @param dbName db file name to search
+     * @param matchExactName whether to look for exact file name or a pattern match
+     * @param parentPathSubstr path substring to match
+     * 
+     * @return AbstractFile for the DB if the database file is found.
```
bcarrier: Comment needs to be updated.

Core/src/org/sleuthkit/autopsy/coreutils/AppSQLiteDB.java
```diff
@@ -0,0 +1,328 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.ReadContentInputStream;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * An abstraction around an SQLite app DB found in a data source.
+ * This class makes a copy of it, opens a SQLite connection to it 
+ * and runs queries on it.
```
bcarrier: Should also  note in here that the WAL file will also be used, if found.

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,1145 @@
...
+     * @param direction call direction
+     * @param fromPhoneNumber originating phone number, may be empty
+     * @param toPhoneNumber recipient phone number, may be empty
+     * @param startDateTime start date/time
+     * @param endDateTime end date/time
+     * @param contactName contact name, may be empty
+     * 
+     * @return call log artifact
+     */
+    public BlackboardArtifact addCalllog( String otherAccountUniqueID,
```
dannysmyda: Not to complicate things, but would be not be shooting ourselves in the foot if we wanted to support group calls? Should this have a similar API to messages?

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,1145 @@
...
+     * @param contactName Name of contact
+     * @param phoneNumber primary phone number for contact
+     * @param homePhoneNumber home phone number 
+     * @param mobilePhoneNumber mobile phone number, 
+     * @param emailAddr Email address for contact
+     * 
+     * @return artifact created
+     * 
+     */
+    public BlackboardArtifact addContact(String contactAccountUniqueID, String contactName,
```
dannysmyda: I have run into examples where a contact has multiple phone numbers, what do we do in this case?

Core/src/org/sleuthkit/autopsy/coreutils/AppDBParserHelper.java
```diff
@@ -0,0 +1,1321 @@
...
+     * @param mediaType called media type
+     * @param otherAttributesList other attributes
+     * 
+     * @return calllog artifact
+     */
+    public BlackboardArtifact addCalllog(CommunicationDirection direction, 
+                                Account.Address fromAddress, 
+                                Collection<Account.Address> toAddressList, 
+                                long startDateTime, long endDateTime,
+                                CallMediaType mediaType,
```
dannysmyda: You don't actually do anything with CallMediaType

raman-bt: Yes, we don't currently have any attribute to show the call type. I added it to the interface now for completeness, so we wont have to alter the interface when we add the attribute later.

PR #5154, opened 2019-09-04
5427 Determine how much text is needed for classification

bhkjersten:
The TextClassificationIngestModule checks how many tokens are in the notable and nonnotable classes of the training data. If there's not enough of one class, throw an Exception.


rcordovano:
Can the Codacy issues for this be easily fixed?

bhkjersten:
Richard, those are small issues that can be easily fixed.

PR #5162, opened 2019-09-06
Verify text classifier works on all document types

bhkjersten:
This works on .doc, .pdf, and .html files.

I changed it so that we ignore short sentences, similarly to how the document summarizer works. I did not use the exact same technique as I did in the document summarizer because

1. De-hyphenation is less important to document categorization than it to extractive summarization. When you are selecting a sentence in a summary, it is vitally important to know whether an individual sentence contains a particular word. If you're just trying to classify a document, it's okay to have an erroneous word split in a sentence here or there. De-hyphenation uses an entire English vocabulary stored in the repo, so it's too expensive for this task.

2. When I applied a stop word filter, the F1 score on the motorcycle test plummeted.

The changes I made were the following:

1. Divide into sentences, using a machine-learned model that knows that the new line character is sometimes a sentence boundary.

2. Reject sentences that contain fewer than five words

I tried other sentence boundary detection techniques, but this is the only one that worked well on my corpus.


Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Writes a naive Bayes classifier model to disk.
+     *
+     * @param model the model
+     * @throws IOException If the model file cannot be written. Large files can
+     * cause this.
+     */
+    static void writeModel(NaiveBayesModel model) throws IOException {
+        ensureTextClassifierFolderExists();
+        FileWriter fw = new FileWriter(new File(MODEL_PATH));
```
wschaeferB: When this throws an exception the file writer may not get closed, consider a try with resources.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
```diff
@@ -0,0 +1,212 @@
...
+        String[] sentences = sentenceDetector.sentDetect(text);
+        for (String sentence : sentences) {
+            //Tokenize the file.
+            String[] sentenceTokens = TOKENIZER.tokenize(sentence);
+            ArrayList<String> tokensToKeep = new ArrayList<>();
+            if (sentenceTokens.length > 5) {
+                tokens.addAll(Arrays.asList(sentenceTokens));
+            }
+        }
+        return tokens.toArray(new String[0]);
```
wschaeferB: tokens.size() can be used in place of zero to provide the size necessary

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
```diff
@@ -0,0 +1,212 @@
...
+                throw new IOException("Error finding sentence detector module");
+            }
+            sentenceDetector = new NewlineHeuristicSentenceDetector(new SentenceDetectorME(new SentenceModel(new FileInputStream(sentenceModelFile))));
+        }
+        ArrayList<String> tokens = new ArrayList<>();
+        String[] sentences = sentenceDetector.sentDetect(text);
+        for (String sentence : sentences) {
+            //Tokenize the file.
+            String[] sentenceTokens = TOKENIZER.tokenize(sentence);
+            ArrayList<String> tokensToKeep = new ArrayList<>();
```
wschaeferB: tokensToKeep is unused

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
```diff
@@ -0,0 +1,212 @@
...
+            return cleanAndTokenize(text);
+        } catch (IOException ex) {
+            LOGGER.log(Level.WARNING, "Cannot extract tokens from file " + file.getName(), ex);
+            return new String[0];
+        }
+    }
+
+    private static String[] cleanAndTokenize(String text) throws IOException {
+        //Initialize the sentenceDetector if that hasn't been done yet.
+        if (sentenceDetector == null) {
```
wschaeferB: Codacy has pointed out that this is not thread safe.

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
wschaeferB: apparently no 0 is more efficient and my netbeans gives me bad tips

Experimental/src/org/sleuthkit/autopsy/experimental/textclassifier/TextClassifierUtils.java
wschaeferB: looks good

PR #5163, opened 2019-09-06
5452 Avoid the "java.io.UTFDataFormatException: encoded string too long" error

bhkjersten:
It looks like misunderstood the problem, but now I have the solution. The error happens during training, not model writing. 

While training, the TwoPassDataIndexer writes all the processed training data to a temporary file for later processing. This temporary file is what has the size limit. There's not way to change what a TwoPassDataIndexer does to get around this limit. The solution is to use a OnePassDataIndexer instead.


wschaeferB:
Needs to be re-targeted at text-classification

PR #5164, opened 2019-09-10
Adding idea project files to .gitignore

ethteck:


.gitignore
```diff
@@ -82,7 +82,8 @@ hs_err_pid*.log
 /RecentActivity/release/
 /CentralRepository/release/
 
-/.idea/
+.idea/
+*.iml
```
rcordovano: What are these files?

ethteck:
These are Intellij project files

.gitignore
rcordovano: What is the context for the presence of Intellij project files in Autopsy?

ethteck:
I've been trying to get Intellij development working and have been
moderately successful.

.gitignore
rcordovano: It seems like adding these files to .gitignore is premature.

ethteck:
Alright. Well one was already added recently in a different PR of mine; I'm
just fixing it to include not just the top level directory. The other is an
extension that nothing else uses as far as I know. I can just add them to
my global .gitignore though if you're against it.

PR #5168, opened 2019-09-11
5444 viber database parser

dannysmyda:


dannysmyda:
This still needs future API changes integrated into it. This is not up to date with the new call logs api.

PR #5176, opened 2019-09-12
5419 browsers maps apps

raman-bt:
This PR should be merged after:

https://github.com/sleuthkit/autopsy/pull/5152


InternalPythonModules/android/installedapps.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+Finds the SQLite DB for Android installed applications, parses the DB,
+and adds artifacts to the case.
+"""
+class InstalledApplicationsAnalyzer(general.AndroidComponentAnalyzer):
+    
+    moduleName = "Android Installed Applications Analyzer"
+    
+    def __init__(self):
+        self._logger = Logger.getLogger(self.__class__.__name__)
```
dannysmyda: Can you include the version of the app you parsed in variable or comment? You can check the About pages of the apps on the S5. It'd be nice to know for any future continued work.

InternalPythonModules/android/installedapps.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def analyze(self, dataSource, fileManager, context):
+        libraryDbs = AppSQLiteDB.findAppDatabases(dataSource, "library.db", True, "com.android.vending")
```
dannysmyda: Could these strings be private instance variables? See my parsers as examples.

InternalPythonModules/android/installedapps.py
```diff
@@ -0,0 +1,79 @@
...
+        for libraryDb in libraryDbs:
+            try:
+                libraryDbHelper = ArtifactsHelper(Case.getCurrentCase().getSleuthkitCase(),
```
dannysmyda: Like we talked about in Slack, I think this should be getCurrentCaseThrows(), otherwise an IllegalStateException is thrown which is trouble.

InternalPythonModules/android/operabrowser.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self):
+        self._logger = Logger.getLogger(self.__class__.__name__)
+
+    def analyzeCookies(self, dataSource, fileManager, context):
+            cookiesDbs = AppSQLiteDB.findAppDatabases(dataSource, "Cookies", True, "com.opera.browser")
+            for cookiesDb in cookiesDbs:
+                try:
+                    cookiesDbHelper = WebBrowserArtifactsHelper(Case.getCurrentCase().getSleuthkitCase(),
+                                        self.moduleName, cookiesDb.getDBFile())
+                    cookiesResultSet = cookiesDb.runQuery("SELECT host_key, name, value, creation_utc FROM cookies")
```
dannysmyda: Some comments on what these columns mean in terms of tsk language may be nice. Like value, what is that? I would have to look at the addWebCookie API to see the name of the 4th parameter to get a sense of what that actually means.

InternalPythonModules/android/installedapps.py
```diff
@@ -0,0 +1,79 @@
...
+                if ownershipResultSet is not None:
+                    while ownershipResultSet.next():
+                        purchase_time = ownershipResultSet.getLong("purchase_time") / 1000
+                        libraryDbHelper.addInstalledProgram(ownershipResultSet.getString("doc_id"),
+                                                    purchase_time)
+            
+            except SQLException as ex:
+                self._logger.log(Level.WARNING, "Error processing query result for installed applications. ", ex)
+            except (TskCoreException, BlackboardException) as ex:
+                self._logger.log(Level.WARNING, "Failed to adding installed application artifacts.", ex)
```
dannysmyda: I believe this constitutes a SEVERE. If it did not add to the case database this means the case is not complete and that is by definition SEVERE by our standards.

dannysmyda: A trace back in all of these spots would be nice since Jython is the only thing on the stacktrace otherwise.

InternalPythonModules/android/operabrowser.py
```diff
@@ -0,0 +1,200 @@
...
+                    self._logger.log(Level.WARNING, "Error processing query results for Opera cookies.", ex)
+                except (TskCoreException, BlackboardException) as ex:
+                    self._logger.log(Level.WARNING, "Failed to add Opera cookie artifacts.", ex)
+                finally:      
+                    cookiesDb.close()                    
+            	
+        
+
+    def analyzeHistory(self, dataSource, fileManager, context):
+            historyDbs = AppSQLiteDB.findAppDatabases(dataSource, "History", True, "com.opera.browser")
```
dannysmyda: same idea with the package name, it should be a class parameter so that if it needed to change we would only need to change in one spot.

InternalPythonModules/android/operabrowser.py
```diff
@@ -0,0 +1,200 @@
...
+                    self._logger.log(Level.WARNING, "Error processing query results for Opera history.", ex)
+                except (TskCoreException, BlackboardException) as ex:
+                    self._logger.log(Level.WARNING, "Failed to add Opera history artifacts.", ex)
+                finally:        
+                    historyDb.close()                    
+                
+        
+
+    def analyzeDownloads(self, dataSource, fileManager, context):
+            downloadsDbs = AppSQLiteDB.findAppDatabases(dataSource, "History", True, "com.opera.browser")
```
dannysmyda: Same comments from above and the installedprograms hold for all the rest of these.

InternalPythonModules/android/installedapps.py
dannysmyda: Same applies for all analyzers.

rcordovano:
@raman-bt, please resolve the merge conflicts, thanks!

PR #5212, opened 2019-09-18
1349 Remove super.createSheet from ContactNode

kellykelly3:


wschaeferB:
Looks good. Should it be targeting the release branch instead?

PR #5214, opened 2019-09-18
5456 Make adding interesting files faster.

jkho:


Core/src/org/sleuthkit/autopsy/logicalimager/dsp/AddLogicalImageTask.java
```diff
@@ -363,47 +427,36 @@ private void addInterestingFiles(Path resultsPath, boolean createVHD) throws IOE
...
                 List<AbstractFile> matchedFiles = Case.getCurrentCase().getSleuthkitCase().findAllFilesWhere(query);
+                List<Long> fileIds = new ArrayList<>();
                 for (AbstractFile file : matchedFiles) {
-                    addInterestingFileToArtifacts(file, ruleSetName, ruleName, artifacts);
+                    fileIds.add(file.getId());
+                }
+                String key = String.format("%s\t%s", ruleSetName, ruleName);
+                List<Long> value = new ArrayList<>();
+                if (interestingFileMap.containsKey(key)) {
+                    value = interestingFileMap.get(key);
```
APriestman: This might be simpler as:
if (interestingFileMap.containsKey(key)) {
 interestingFileMap.get(key).addAll(fileIds);
} else {
 interestingFileMap.put(key, fileIds);
}

Core/src/org/sleuthkit/autopsy/logicalimager/dsp/AddLogicalImageTask.java
```diff
@@ -454,12 +507,19 @@ private void addExtractedFiles(File src, Path resultsPath, List<Content> newData
                         Long.parseLong(atime),
                         Long.parseLong(mtime),
                         localFilesDataSource);
-
+                    String key = String.format("%s\t%s", ruleSetName, ruleName);
+                    List<Long> value = new ArrayList<>();
```
APriestman: This could also probably be done with using the "value" list

Core/src/org/sleuthkit/autopsy/logicalimager/dsp/AddLogicalImageTask.java
jkho: Fixed

Core/src/org/sleuthkit/autopsy/logicalimager/dsp/AddLogicalImageTask.java
jkho: I can't here. I am appending one fileId to the value(s) of a key (ruleSetName + ruleName).

PR #5227, opened 2019-09-19
5347 Changed ingest module docs to use postArtifact().

APriestman:
Updated link to blackboard.

### DIFF
--- a/docs/doxygen/modIngest.dox
+++ b/docs/doxygen/modIngest.dox
@@ -200,10 +200,10 @@ You can use one of the standard artifact types or create your own.
 
 After you make an artifact, you should call <a href="http://sleuthkit.org/sleuthkit/docs/jni-docs/4.6/classorg_1_1sleuthkit_1_1datamodel_1_1_blackboard.html">sleuthkit.Blackboard.postArtifact()</a>, which will:
 <ul>
-<li>Analyze the artifact and add any timestamps to the new timeline tables
-<li>Send an event over its new EventBus that the artifact(s) was added.
+<li>Analyze the artifact and add any timestamps to the Timeline tables
+<li>Send an event over the EventBus that the artifact(s) was added
 <ul>
-<li>Autopsy is a listener of this new eventbus and will rebroadcast it to other Autopsy modules
+<li>Autopsy is a listener of this EventBus and will rebroadcast it to other Autopsy modules
 <li>Keyword search will also listen for this and will index the artifact
 </ul>
 </ul>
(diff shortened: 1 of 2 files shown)


docs/doxygen/modIngest.dox
```diff
@@ -198,27 +198,23 @@ See \ref platform_blackboard  for details on posting results to it.  You use the
...
-	blackboard.indexArtifact(artifact); //Your artifact as the argument.
-}
-catch (BlackboardException ex) {
-	//YOUR EXCEPTION BEHAVIOR HERE.
-}
-\endcode
+After you make an artifact, you should call <a href="http://sleuthkit.org/sleuthkit/docs/jni-docs/4.6/classorg_1_1sleuthkit_1_1datamodel_1_1_blackboard.html">sleuthkit.Blackboard.postArtifact()</a>, which will:
+<ul>
+<li>Analyze the artifact and add any timestamps to the Timeline tables
+<li>Send an event over the EventBus that the artifact(s) was added
```
rcordovano: Do we want to refer to the event bus as "the EventBus" vs. "the Sleuth Kit event bus"?

docs/doxygen/modIngest.dox
```diff
@@ -198,27 +198,23 @@ See \ref platform_blackboard  for details on posting results to it.  You use the
...
+<ul>
+<li>Autopsy is a listener of this EventBus and will rebroadcast it to other Autopsy modules
```
rcordovano: I would say "rebroadcast the event" in place of "it."

docs/doxygen/modIngest.dox
```diff
@@ -198,27 +198,23 @@ See \ref platform_blackboard  for details on posting results to it.  You use the
...
+<li>Keyword search will also listen for this and will index the artifact
+</ul>
+</ul>
+
+This means you no longer have to make calls to:
+ - Index the artifact
+ - Fire the event to refresh the UI.
+ 
+If you are creating a large number of artifacts, you may see better performance if you save all the artifacts and do one bulk write at the end using <a href="http://sleuthkit.org/sleuthkit/docs/jni-docs/4.6/classorg_1_1sleuthkit_1_1datamodel_1_1_blackboard.html">sleuthkit.Blackboard.postArtifacts()</a>
+
```
rcordovano: I think it would be helpful to mention that we can batch making the artifacts as well as just "posting" them. I find "posting" to be unfortunately ambiguous - to me it sounds like that method would also add the artifacts.

APriestman: We can make artifacts in batches? I know we can batch up the attributes but I didn't think you could make an artifact without adding it to the DB at the same time. Regardless, the wording is a bit confusing (in the top paragraph it does use "post" to mean "add to DB") so I'll work on that.

rcordovano: No, it turns out that we cannot make artifacts in batches. It's attributes that we can make in batches. In fact, BlackboardArtifact does not have a public constructor to allow such a thing. Maybe we should change this and provide a way to add artifacts in batches, but that is a story for another day. Sorry for my confusion.

PR #5229, opened 2019-09-19
1352 - Timeline snap report exception fix

kellykelly3:


Core/src/org/sleuthkit/autopsy/timeline/actions/Bundle.properties-MERGED
```diff
@@ -36,6 +36,8 @@ SaveSnapShotAsReport.reportName.prompt=leave empty for default report name: {0}.
 # {0} - report file path
 SaveSnapShotAsReport.ReportSavedAt=Report saved at [{0}]
 SaveSnapShotAsReport.Success=Success
+SaveSnapShotAsReport_Path_Failure_Report=Failed to create report.  Invalid path: {0}
```
wschaeferB: It looks like this merged file was not the most recent changes and needs to be updated again.

PR #5233, opened 2019-09-20
1335 & 1354 vcard dups

kellykelly3:
Not when reviewing\testing this code: The changes for 1335 makes it hard to see the fix for 1354 unless you make a vcard with duplicate phone numbers.


Core/src/org/sleuthkit/autopsy/communications/relationships/ContactNode.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.sleuthkit.autopsy.communications.relationships;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
```
wschaeferB: Hashmap and Map are now unused imports

PR #5236, opened 2019-09-20
1353 - Added device back at case closure

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/FiltersPanel.java
```diff
@@ -257,9 +257,14 @@ public void addNotify() {
             //clear the device filter widget when the case changes.
             devicesMap.clear();
             devicesListPane.removeAll();
-
-            accountTypeMap.clear();
+            
+            accountTypeMap.clear();           
             accountTypeListPane.removeAll();
+            
+            CheckBoxIconPanel panel = createAccoutTypeCheckBoxPanel(Account.Type.DEVICE, true);
```
bcarrier: Seems like this should be an init/reset method that gets called both at initialization time and during reset.

PR #5238, opened 2019-09-20
Bug fix (5552)

eugene7646:


Core/src/org/sleuthkit/autopsy/report/infrastructure/ReportWizardFileOptionsVisualPanel.java
```diff
@@ -80,8 +80,10 @@ private void initOptionsList(FileReportSettings settings) {
         optionsList.setCellRenderer(new OptionsListRenderer());
         optionsList.setVisibleRowCount(-1);
 
-        selectAllButton.setEnabled(true);
-        deselectAllButton.setEnabled(false);
+        selectAllButton.setEnabled(false);
+        deselectAllButton.setEnabled(true);
+        
+        wizPanel.setFinish(true);
```
wschaeferB: it works in most cases but now you can get into an edge case where you can click finish with no items selected, see Jira for steps to create edge case.

eugene7646: @wschaeferB good catch! Fixed now.

PR #5242, opened 2019-09-23
1356 - cleaned up cvt layout

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/CVTTopComponent.java
```diff
@@ -19,7 +19,9 @@
 package org.sleuthkit.autopsy.communications;
 
 import com.google.common.eventbus.Subscribe;
+import java.awt.BorderLayout;
 import java.awt.Component;
+import java.awt.Dimension;
 import java.awt.Font;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
```
wschaeferB: gridbaglayout, splitpane, and  insets have become unused imports

PR #5250, opened 2019-09-23
Progress panel (5555)

eugene7646:


APriestman:
Can you fix that quick codacy issue?

eugene7646:
> Can you fix that quick codacy issue?

@APriestman I'd honestly rather leave it as is. Every single method in that interface does have "public" in front of it. It would look really odd to have a single method that is not declared "public", even if it doesn't actually make a difference.

PR #5258, opened 2019-09-24
1360 - Bulk postArtifacts calls added to ExtractRegistry

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -873,20 +865,22 @@ private boolean parseAutopsyPluginOutput(String regFilePath, AbstractFile regFil
             } // for
             return true;
         } catch (FileNotFoundException ex) {
-            logger.log(Level.WARNING, String.format("Error finding the registry file: %s", regFilePath), ex); //NON-NLS
+            logger.log(Level.SEVERE, String.format("Error finding the registry file: %s", regFilePath), ex); //NON-NLS
```
dannysmyda: Why have these been changed to severe? This method is far too hard to read, but these seem related to bad data which I would think are warnings.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -721,7 +718,7 @@ private boolean parseAutopsyPluginOutput(String regFilePath, AbstractFile regFil
                                                 itemMtime /= MS_IN_SEC;
                                             }
                                         } catch (ParseException ex) {
-                                            logger.log(Level.WARNING, "Failed to parse epoch time for installed program artifact.", ex); //NON-NLS
+                                            logger.log(Level.SEVERE, "Failed to parse epoch time for installed program artifact.", ex); //NON-NLS
```
dannysmyda: This seems like a malformed data exception, why is this logged as severe?

kellykelly3: They were originally SEVERE, I changed them to WARNING and I was asked by Richard to put them back the way they were.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
kellykelly3: They were originally SEVERE, I changed them to WARNING and I was asked by Richard to put them back the way they were.

rcordovano: @dannysmyda and @kellykelly3, I was referring to TskCoreExceptions that I believed indicated an INSERT into the case database to make a ready to go artifact had failed. IN other words, my request covered situations where the analysis result was complete and in hand, but we were unable to add it to the case.

PR #5259, opened 2019-09-24
Bug fix (5570)

eugene7646:


eugene7646:
> The status of the "All Tagged Results" checkbox and the hashset name be saved are no longer being restored to their previous values with this PR. I understand that individual tags will not be saved for the reasons given in the other stories.

@wschaeferB good catch! all done.

PR #5264, opened 2019-09-25
5587: Error from ShareIt module

raman-bt:
- filter out Null & blank strings while appending attachments list to msg body.


dannysmyda:
I could never reproduce the original bug on release, but this PR is only an improvement and has been verified to not change anything so I'm approving.

PR #5280, opened 2019-09-27
1348 cvt summary changes

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/Bundle.properties-MERGED
```diff
@@ -92,6 +92,7 @@ VisualizationPanel.forwardButton.text=
 VisualizationPanel.zoomPercentLabel.text=100%
 VisualizationPanel.zoomLabel.text=Zoom:
 VisualizationPanel.snapshotButton.toolTipText=Generate Snapshot report.
+SummaryViewer.contactsLabel.text=Entries in Book*:
```
wschaeferB: When I build this merged properties does not have the * between Book and the : and looks the same as the one in Bundle.properties I do not know if this means the MERGED file needs to be updated or it is one of our phantom diffs of merged files where they appear different for different users like the slashes.

Core/src/org/sleuthkit/autopsy/communications/relationships/Bundle.properties-MERGED
```diff
@@ -35,20 +35,23 @@ MessageViewer_viewMessage_all=All
...
 SummaryViewer.contactsDataLabel.text=contacts
 SummaryViewer.messagesLabel.text=Messages:
 SummaryViewer.callLogsLabel.text=Call Logs:
+SummaryViewer_Account_Description=This account represents a device in the case.
+SummaryViewer_Account_Description_MuliSelect=Summary information is not available when multiple accounts are selected.
 SummaryViewer_CaseRefNameColumn_Title=Case Name
 SummaryViewer_CentralRepository_Message=<Enable Central Resposity to see Other Occurrences>
 SummaryViewer_Creation_Date_Title=Creation Date
+SummaryViewer_Device_Account_Description=This account was referenced by a device in the case.
+SummaryViewer_FileRef_Message=<Select one Accout to see File References>
```
wschaeferB: This message was just changed so the merged file needs to be updated as well.

kellykelly3: I don't understand why this didn't update.

PR #5287, opened 2019-10-01
5562 tsk comment attributes with list of prev cases are added

markmckinnon:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
```diff
@@ -482,9 +485,11 @@ public void run() {
...
                                     List<CorrelationAttributeInstance> previousOccurences = dbManager.getArtifactInstancesByTypeValue(eamArtifact.getCorrelationType(), eamArtifact.getCorrelationValue());
+                                    List<String> caseDisplayNames;
                                     for (CorrelationAttributeInstance instance : previousOccurences) {
                                         if (!instance.getCorrelationCase().getCaseUUID().equals(eamArtifact.getCorrelationCase().getCaseUUID())) {
-                                            makeAndPostPreviousSeenArtifact(bbArtifact);
+                                            caseDisplayNames = dbManager.getListCasesHavingArtifactInstances(eamArtifact.getCorrelationType(), eamArtifact.getCorrelationValue());
+                                            makeAndPostPreviousSeenArtifact(bbArtifact, caseDisplayNames);
                                             break;
                                         }
                                     }
```
wschaeferB: I don't know if we are concerned about adding another CR query or not, it seems like something that shouldn't happen too often so it is probably not a big deal.

However we could modify this loop to call instance.getCorrelationCase().getDisplayName() and build the list as an else condition to the if.  The contents of the 'if' would need to be replaced with a flag being set to true and then if the flag was true they would be performed after looping. 

Once again I don't know if we care about one more query given how often this would happen and it would be a major sacrifice for code readability for a this small reduction in queries.

PR #5292, opened 2019-10-01
5604: CommunicationArtifactsHelper changes

raman-bt:
- get rid Account.Address
 - handle required attribute arguments.

### DIFF
--- a/InternalPythonModules/android/fbmessenger.py
+++ b/InternalPythonModules/android/fbmessenger.py
@@ -20,7 +20,6 @@ limitations under the License.
 from java.io import File
 from java.lang import Class
 from java.lang import ClassNotFoundException
-from java.lang import IllegalArgumentException
 from java.lang import Long
 from java.lang import String
 from java.sql import ResultSet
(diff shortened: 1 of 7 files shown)


InternalPythonModules/android/fbmessenger.py
```diff
@@ -156,8 +158,11 @@ def analyzeContacts(self, dataSource, fileManager, context):
                 self._logger.log(Level.WARNING, "Error processing query result for account", ex)
                 self._logger.log(Level.WARNING, traceback.format_exc())
             except TskCoreException as ex:
-                self._logger.log(Level.SEVERE, "Failed to add Facebook Messenger contact artifacts.", ex)
+                self._logger.log(Level.SEVERE, "Failed to add FB Messenger contact artifacts.", ex)
                 self._logger.log(Level.SEVERE, traceback.format_exc())
+            except IllegalArgumentException as ex:
```
dannysmyda: All of these IllegalArgumentExeceptions should not be caught and logged. This means the query is not right/there are programming errors we need to fix. The general exception firewall in Android analyzer will handle these.

raman-bt: These are from sending incorrect parameters to api - like missing a required parameter.  A lot of it may be dependant on the input data in the database.

It seems rather extreme to me to shutdown the entire module if we hit one bad row.

rcordovano:
The purpose of unchecked exceptions is to make programming errors into
catastrophic errors. We should only be throwing IllegalArgumentExceptions
from these APIs if the developer has used them incorrectly; the unchecked
exceptions should get the developer's attention during testing.

InternalPythonModules/android/fbmessenger.py
rcordovano: @raman-bt, the purpose of unchecked exceptions is to make programming errors into
catastrophic errors. We should only be throwing IllegalArgumentExceptions
from these APIs if the developer has used them incorrectly; the unchecked
exceptions should get the developer's attention during testing.

raman-bt: Got it. Removed catching the IllegalArgumentExceptions.

PR #5293, opened 2019-10-01
5565 Add support for webp image files

markmckinnon:
Add new library to support webp format and add supported file mime types of image/webp


esaunders:
Is there a reason why the JAR file needs to be checked into thirdparty? It looks like it can be downloaded at build time through ivy/maven.
https://mvnrepository.com/artifact/org.sejda.webp-imageio/webp-imageio-sejda/0.1.0

esaunders:
After getting the JAR using Ivy you still need the step (that you had previously) of adding it as a wrapped jar to the Autopsy-Core project.

esaunders:
I know this wasn't explicitly called out in the original story but I would think that (a) we want the webp files to appear in the results table under Views -> File Types -> By Extension -> Images and (b) the icon for these files in the results table should be the same as is used for jpeg and tif (but not gif apparently).

markmckinnon:
Will make above changes

markmckinnon:
Everything should be all set now.

esaunders:
Cool, I'll check it out now. Can you remove the Bundle.properties from the PR...I don't believe it is necessary? Thanks.

markmckinnon:
I put in the old copy of the bundle.properties so it should have removed it from the PR as a file changed.  I just checked the files and do not see that as a file that was changed now.  Unless I am looking at something wrong.

esaunders:
Ok, I'm not seeing it now. Don't know what it was showing me earlier.

PR #5308, opened 2019-10-02
1371, 1372, 1372 - inital pass cvt redesign

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/AccountsBrowser.java
```diff
@@ -65,20 +65,18 @@
...
+    final private RelationshipBrowser relationshipBrowser;
 
     /*
      * This lookup proxies the selection lookup of both he accounts table and
      * the messages table.
      */
     private final ProxyLookup proxyLookup;
 
-    public AccountsBrowser() {
+    public AccountsBrowser(RelationshipBrowser relationshipBrowser) {
```
wschaeferB: Just to repeat what codacy has said this Relationship browser argument and class variable can be removed as they are unused.

kellykelly3:
Fixed codacy issues, waiting on final arrow icons from Brian.  Once I have the icons it will take about 10 minutes to change them.

PR #5331, opened 2019-10-08
5579 - sanitize file names in embedded file extractor

wschaeferB:


wschaeferB:
making again with correct story number

PR #5332, opened 2019-10-08
5579 - sanitize file names in embedded file extractor

wschaeferB:


Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/DocumentEmbeddedContentExtractor.java
```diff
@@ -159,7 +161,8 @@ boolean isContentExtractionSupported(AbstractFile abstractFile) {
     void extractEmbeddedContent(AbstractFile abstractFile) {
         List<ExtractedFile> listOfExtractedImages = null;
         List<AbstractFile> listOfExtractedImageAbstractFiles = null;
-        this.parentFileName = EmbeddedFileExtractorIngestModule.getUniqueName(abstractFile);
+        //save the parent file name with out illegal windows characters
+        this.parentFileName = FileUtil.escapeFileName(EmbeddedFileExtractorIngestModule.getUniqueName(abstractFile));
```
eugene7646: As I have found out during my work on ArchiveDSP, escapeFileName() is not sufficient for file name sanitization (see JIRA-5416). This method gets rid of illegal Windows characters, but it happily "passes" the non-UTF8 characters. See ArchiveUtil.unpackArchiveFile() for proper file name sanitization, namely see that it uses utf8SanitizeFileName().

wschaeferB: Good catch, thank you.

PR #5338, opened 2019-10-10
fix name of build target for tsk debugging

wschaeferB:


docs/doxygen/debugTsk.dox
```diff
@@ -3,7 +3,7 @@
 If you find that you need to debug some of the C/C++ code from The Sleuth Kit (TSK), then here are the steps to follow:
 
 -# Ensure that you have the Debug version of the TSK JNI dll built (both 32-bit and 64-bit to be safe).  This assumes you built TSK from source and are not simply using the developer platform. You may have to build the libtsk_jni twice because sommetimes it complains about not being able to find a .map file. 
--# Run the 'dist-debug' target for the TSK DataModel project.  This copies the debug versions of the dll into the JAR file.  If you run the 'dist' target, then you will get Release versions of the dll and you won't have the needed symbols for debugging. 
+-# Run the 'Debug-PostgreSQL' target for the TSK DataModel project.  This copies the debug versions of the dll into the JAR file.  If you run the 'dist' target, then you will get Release versions of the dll and you won't have the needed symbols for debugging.
```
rcordovano: Should it actually be 'Debug-PostgreSQL' or 'Debug-SQLite'? I'm guessing that 'dist-debug' dates back to before we added PostgreSQL support and there was only one debug target.

wschaeferB: I do not know. What I can say is Debug-PostgreSQL worked and it was the first one I tried since it seemed to be the analogous Debug equivalent of dist-PostgreSQL. I do not know what the -SQLite targets are for.

rcordovano: Actually, upon further reflection, I think that you are right. We want the Debug-PostgreSQL targte for Autopsy debugging.

PR #5340, opened 2019-10-15
Updated and tested the new template code

dannysmyda:


bcarrier:
We should move this to a separate file since the original blog refers to the original file. Something like FindContactsDb_Updated.py or FindContactsDb_v2.py

dannysmyda:
@markmckinnon As far as I know, all that needs to be verified is that the template code can be compiled and run on the [contacts.db](https://github.com/sleuthkit/autopsy/blob/develop/pythonExamples/Aug2015DataSourceTutorial/contacts.db) file when ingested as a LogicalFileSet. For context, this is an updated template that showcases use of the new artifacts helper.

markmckinnon (approved):

PR #5345, opened 2019-10-16
5657 file discovery top component

wschaeferB:
Should be reviewed after PR #5165


Core/src/org/sleuthkit/autopsy/filequery/DiscoveryEvents.java
```diff
@@ -213,6 +213,20 @@ FileType getType() {
...
+    /**
+     * Event to signal that a search has been cancelled
+     */
+    static final class SearchCancelledEvent {
+
+        /**
+         * Construct a new SearchCancelledEvent.
+         */
+        SearchCancelledEvent() {
+            //no arg conustructor
```
APriestman: typo

PR #5346, opened 2019-10-16
Workarounds to ensure that dialogs are not hidden on macOS.

esaunders:


Core/src/org/sleuthkit/autopsy/casemodule/CaseOpenAction.java
```diff
@@ -84,18 +84,29 @@ void openCaseSelectionWindow() {
...
              * Pop up a file chooser to allow the user to select a case metadata
              * file (.aut file).
              */
-            int retval = fileChooser.showOpenDialog(WindowManager.getDefault().getMainWindow());
+            /**
+             * Passing the fileChooser as its own parent gets around an issue
+             * where the fileChooser was hidden behind the CueBannerPanel ("Welcome" dialog)
+             * on macOS.
+             */
+            int retval = fileChooser.showOpenDialog(fileChooser);
```
wschaeferB: This change causes us to re-introduce the issue where the dialog will be opened on the main display instead of over the current location of Autopsy. This does not strike me as a deal breaker but as it is a regression of sorts I wanted to make sure it was noted.

esaunders: I believe I've addressed this so it's ready for review again.

wschaeferB:
I tested only on Windows. The code works on windows without any regressions and the code changes make sense that they will fix the issue encountered on mac os.

PR #5348, opened 2019-10-17
5681-npe-running-email-parser-against-tb-mailbox

markmckinnon:
Check threaded messageId if null and skip it if it is.  Format code also.

### DIFF
--- a/thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
+++ b/thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
@@ -458,12 +458,11 @@ public final class ThunderbirdMboxFileIngestModule implements FileIngestModule {
             if (partialEmailsForThreading.size() > msgCnt) {
                 EmailMessage threaded = partialEmailsForThreading.get(msgCnt++);
 
-                if (threaded.getMessageID() != null) {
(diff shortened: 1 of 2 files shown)


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -435,38 +436,39 @@ static String getRelModuleOutputPath() throws NoCurrentCaseException {
...
-            if(partialEmailsForThreading.size() > msgCnt) {
+            if (partialEmailsForThreading.size() > msgCnt) {
                 EmailMessage threaded = partialEmailsForThreading.get(msgCnt++);
-                
-                if(threaded.getMessageID().equals(current.getMessageID()) &&
-                        threaded.getSubject().equals(current.getSubject())) {
-                    current.setMessageThreadID(threaded.getMessageThreadID());
-                }
+
+                if ((threaded.getMessageID() != null) &&
```
kellykelly3: It's fine to add the null check, but the way EmailMessage is suppose to work is that all the functions that return strings should return empty string instead of null.  So that means somewhere in the threading code the the message id is being set to null instead of empty string.

markmckinnon: I took out the code I previously put in and made a change to setMessageID to check for null and return "" if it was.

PR #5354, opened 2019-10-18
5594 set playback speed

dannysmyda:


raman-bt:
Move the "Volume" label a little to the right.

raman-bt:
> @raman-bt My next story is strictly UI changes. I have fixed it there. Is that okay?

Ok.

PR #5355, opened 2019-10-21
5560  FB messenger calllogs

raman-bt:


dannysmyda:
@raman-bt The sender also appears in the recipients list. As a result, the call logs tab in the CVT shows the account making an "outgoing" call to itself, which can't be.

raman-bt:
> @raman-bt The sender also appears in the recipients list. As a result, the call logs tab in the CVT shows the account making an "outgoing" call to itself, which can't be.

Fixed.  Earlier I had a check to not add the sender to recipient list, but I'd taken it out.

dannysmyda:
@raman-bt I see now that the id doesn't appear in the recipients of the results tree, however, there are still self links on the Account in CVT.

PR #5358, opened 2019-10-21
5595 playback slider UI changes

dannysmyda:


Core/src/org/sleuthkit/autopsy/contentviewers/MediaPlayerPanel.java
```diff
@@ -531,22 +549,236 @@ protected void done() {
...
+                        progressSlider.setValue((int) (relativePosition * PROGRESS_SLIDER_SIZE));
+                    }
+
+                    SwingUtilities.invokeLater(() -> {
+                        updateTimeLabel(position, duration);
+                    });
+                    sliderLock.release();
                 }
+            } catch (InterruptedException ex) {
+                //Video panel thread interrupted while waiting on lock.
```
raman-bt: does the sliderLock get released if there's an exception?

Core/src/org/sleuthkit/autopsy/contentviewers/MediaPlayerPanel.java
```diff
@@ -531,22 +549,236 @@ protected void done() {
...
+                //The VideoPanelUpdater (dedicated thread) keeps the
+                //slider in sync with the video position, so without 
+                //proper locking our change could be overwritten.
+                sliderLock.acquire();
+                slider.setValueIsAdjusting(true);
+                slider.setValue(value);
+                slider.setValueIsAdjusting(false);
+                sliderLock.release();
+            } catch (InterruptedException ex) {
+                //Thread (EDT) interrupted while waiting on lock.
```
raman-bt: will the lock be released if there was an exception?

Core/src/org/sleuthkit/autopsy/contentviewers/MediaPlayerPanel.java
dannysmyda: If this block is caught the lock was never acquired.

Core/src/org/sleuthkit/autopsy/contentviewers/MediaPlayerPanel.java
dannysmyda: Since I don't do anything in these blocks, I should instead use acquireUninterruptibly()

PR #5361, opened 2019-10-22
5665 creation of menus for collection

wschaeferB:


wschaeferB:
I do not know what I can do to that method to reduce its complexity, so I was hoping I could let this issue continue to exist.

kellykelly3:
That's what I figured.

PR #5362, opened 2019-10-22
5682 simplify app db queries

dannysmyda:


dannysmyda:
@rcordovano The story has been verified, but this PR is still hanging out.

PR #5380, opened 2019-10-29
3649-investigate-encoding-of-case-description

markmckinnon:
Add UTF-8 Encoding so it writes to the aut file correctly for non-latin characters.


Core/src/org/sleuthkit/autopsy/casemodule/CaseMetadata.java
```diff
@@ -427,9 +428,10 @@ void writeToFile() throws CaseMetadataException {
             transformer.transform(source, streamResult);
 
             /*
-             * Write the DOM to the metadata file.
+             * Write the DOM to the metadata file.  Add UTF-8 Characterset so it writes to the file
+             * correctly for non-latin charadters
```
esaunders: A wee typo...charadters...other than that it looks good.

PR #5383, opened 2019-10-30
5078-HTML-viewer-not-correctly-interpreting-Unicode

markmckinnon:
Determine encoding of html file and display html file using encoding.

### DIFF
--- a/Core/ivy.xml
+++ b/Core/ivy.xml
@@ -44,6 +44,7 @@
         <dependency conf="core->default" org="org.apache.opennlp" name="opennlp-tools" rev="1.9.1"/>
         
         <dependency org="org.sejda.webp-imageio" name="webp-imageio-sejda" rev="0.1.0"/>
+        <dependency org="net.sourceforge.jchardet" name="jchardet" rev="1.0"/>
         
         <dependency conf="core->default" org="commons-validator" name="commons-validator" rev="1.6"/>
         <dependency conf="core->default" org="net.htmlparser.jericho" name="jericho-html" rev="3.3"/>
--- a/Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
+++ b/Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
@@ -20,9 +20,11 @@ package org.sleuthkit.autopsy.contentviewers;
 
 import java.awt.Component;
 import java.awt.Cursor;
+import java.io.UnsupportedEncodingException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Level;
+import org.mozilla.universalchardet.UniversalDetector;
 import org.openide.util.NbBundle;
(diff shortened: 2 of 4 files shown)


Core/ivy.xml
```diff
@@ -44,6 +44,7 @@
         <dependency conf="core->default" org="org.apache.opennlp" name="opennlp-tools" rev="1.9.1"/>
         
         <dependency org="org.sejda.webp-imageio" name="webp-imageio-sejda" rev="0.1.0"/>
+        <dependency org="net.sourceforge.jchardet" name="jchardet" rev="1.0"/>
```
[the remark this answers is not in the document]
markmckinnon: No problem I will take a look at it.

Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
```diff
@@ -65,14 +67,37 @@ private String getHtmlText(AbstractFile abstractFile) {
...
             abstractFile.read(buffer, 0, fileSize);
-            return new String(buffer);
-        } catch (TskCoreException ex) {
+            String encoding = determineEncoding(buffer);
+            if (encoding != null) {
+                return new String(buffer, encoding);
+            } else {
+                return new String(buffer);
+            }
+        } catch (TskCoreException | UnsupportedEncodingException ex) {
```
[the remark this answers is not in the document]
markmckinnon: Did change to IOException using Tika so it should be ok since it fits the error message.

Core/src/org/sleuthkit/autopsy/contentviewers/HtmlViewer.java
```diff
@@ -65,14 +68,35 @@ private String getHtmlText(AbstractFile abstractFile) {
...
 
+    /**
+     * This method will try and determine the encoding of the html file based on its contents
+     * 
+     * @param buffer byte array of the html file to check
+     * 
+     * @return encoding type, null if encoding could not be determined
+     */
+    private String determineEncoding(byte[] buffer) throws IOException {
+
```
esaunders: I don't think any of the methods you use here throw an IOException so that's not needed.
Once you get rid of that you should notice that you will again have the possibility of an UnsupportedEncodingException in getHtmlText() above.

Also, detector.getDetect() can return null so you need to handle that before calling CharsetMatch.getName().

Personally, I would do away with this method and do the following in getHtmlText() but I'll leave it up to you.

            CharsetMatch match = new CharsetDetector().setText(buffer).detect();
            if (match != null) {
                return new String(buffer, match.getName());
            } else {
                return new String(buffer);
            }

markmckinnon: Changed code per your suggestion, much cleaner.

esaunders: I like it a lot :-)

PR #5395, opened 2019-11-05
Simplified the geolocation model

kellykelly3:


kellykelly3:
@rcordovano Codacy thinks Waypoint is a God class now.  Which I can see.  How do you feel about that?

PR #5405, opened 2019-11-07
5732 implement xry report validation

dannysmyda:


dannysmyda:
I would ignore Codacy for the time being. These classes will be extended in later stories to contain other non-static methods.

rcordovano:
Actually, there is an issue I messed, but Codacy did not. The utility classes need private constructors. I don't care about the naming rules violated. We need to suppress them at some point.

dannysmyda:
@rcordovano These classes were never intended to be utility classes. It just so happens that this portion of the design is utility-esque. I'd rather group these methods with their relevant classes than have them floating off somewhere non-obvious.

rcordovano:
Sounds good. Please resolve the merge conflict and I will merge.

dannysmyda:
> Sounds good. Please resolve the merge conflict and I will merge.

Done.

PR #5406, opened 2019-11-07
5654 Geolocation waypoint context menu

kellykelly3:


kellykelly3:
I have fixed all the codacy issues in my code.  I have a pass from Richard on KdTree and I did not write DeleteFileBlackboardArtifactTagAction class.  

The empty if makes sense in the context of what I am doing.  I put a comment in to explain.

Core/src/org/sleuthkit/autopsy/geolocation/Bundle.properties-MERGED
```diff
@@ -3,8 +3,11 @@ CTL_GeolocationTopComponentAction=GeolocationTopComponent
 CTL_GeolocationTopComponent=Geolocation
 GLTopComponent_initilzation_error=An error occurred during waypoint initilization.  Geolocation data maybe incomplete.
 GLTopComponent_name=Geolocation
+MayWaypoint_ExternalViewer_label=Open in ExternalViewer
 OpenGeolocationAction_displayName=Geolocation
 OpenGeolocationAction_name=Geolocation
 RefreshPanel.refreshLabel.text=The geolocation data has been updated, the visualization may be out of date.
 RefreshPanel.refreshButton.text=Refresh View
 RefreshPanel.closeButton.text=
+MapPanel.cordLabel.text=jLabel1
```
wschaeferB: we do not need this as a Bundle message it the label can be switched to being plain text with an empty string, if we do leave it as a bundle property though it "jLabel1" should probably be replaced with the empty string.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
```diff
@@ -18,69 +18,229 @@
...
+     * 
+     * @throws TskCoreException 
+     */
+    JMenuItem[] getMenuItems() throws TskCoreException{
+        List<JMenuItem> menuItems = new ArrayList<>();
+        BlackboardArtifact artifact = dataModelWaypoint.getArtifact();
+        Content content = artifact.getSleuthkitCase().getContentById(artifact.getObjectID());
+
+        menuItems.addAll(getTimelineMenuItems(dataModelWaypoint.getArtifact()));
+        menuItems.addAll(getDataModelActionFactoryMenuItems(artifact, content));
```
wschaeferB: getTimelineMenuItems is getting called twice as it is also called in getDataModelActionFactoryMenuItems

PR #5410, opened 2019-11-12
5629-shellbag-data-from-regripper-is-incorrect

markmckinnon:
Fix getting long file name string. If the extension verison was not a x03, x07 or x08 the start position was set to null to get the string so it would start at the beginning of the block so would get junk data.  This change has also been pushed to the Regripper2.8 repo as well

### DIFF
--- a/thirdparty/rr-full/plugins/itempos.pl
+++ b/thirdparty/rr-full/plugins/itempos.pl
@@ -1,10 +1,8 @@
 #-----------------------------------------------------------
-# shellbags.pl
-# RR plugin to parse (Vista, Win7/Win2008R2) shell bags
-#
+# itempos.pl
+# 
 # History:
-#   20130514 - added checking of ShellNoRoam\Bags subkeys for WinXP
-#   20120814 - created
+#   20191111 - Added default value to $jmp if $item{extver} cannot be determined.
 #
 # References
 #    http://c0nn3ct0r.blogspot.com/2011/11/windows-shellbag-forensics.html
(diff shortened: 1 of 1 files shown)


thirdparty/rr-full/plugins/itempos.pl
```diff
@@ -1,10 +1,8 @@
 #-----------------------------------------------------------
-# shellbags.pl
-# RR plugin to parse (Vista, Win7/Win2008R2) shell bags
-#
+# itempos.pl
+# 
 # History:
-#   20130514 - added checking of ShellNoRoam\Bags subkeys for WinXP
-#   20120814 - created
+#   20191111 - Added default value to $jmp if $item{extver} cannot be determined.
```
raman-bt: Why remove previous history?

thirdparty/rr-full/plugins/itempos.pl
```diff
@@ -21,8 +19,7 @@
 # provided, Andrew Case for posting the Registry Decoder code, and Kevin 
 # Moore for writing the shell bag parser for Registry Decoder, as well as 
 # assistance with some parsing.
-#
-# License: GPL v3 
+#
```
raman-bt: Is that license not valid any more?

thirdparty/rr-full/plugins/itempos.pl
markmckinnon: This was removed in 2018 by keydet89 in the the regripper2.8 repo.

thirdparty/rr-full/plugins/itempos.pl
markmckinnon: The license appears to have been removed from most files in favor of a license.txt file.  This was removed in 2018 by keydet89 in the the regripper2.8 repo.

PR #5411, opened 2019-11-12
5707, 5708 attchmnt msg contentviewer

raman-bt:


Core/src/org/sleuthkit/autopsy/datamodel/AttachmentNode.java
```diff
@@ -0,0 +1,207 @@
...
+import org.sleuthkit.datamodel.TskException;
+import org.sleuthkit.datamodel.blackboardutils.Attachment;
+import org.sleuthkit.datamodel.blackboardutils.FileAttachment;
+import org.sleuthkit.datamodel.blackboardutils.URLAttachment;
+
+/**
+ * Node for a message attachment.
+ *
+ */
+public class AttachmentNode extends DisplayableItemNode {
```
rcordovano: Does this class have to be public? If so, perhaps we could make it final.

Core/src/org/sleuthkit/autopsy/contentviewers/MessageContentViewer.java
```diff
@@ -633,16 +660,20 @@ static private String cleanseHTML(String htmlInString) {
         return doc.html();
     }
 
-    private static class AttachmentsChildren extends Children.Keys<AbstractFile> {
+    
+    /**
+     * Creates child nodes for message attachments. 
+     */
+    final class AttachmentsChildren extends Children.Keys<Attachment> {
```
rcordovano: It's not clear to me why this class was changed to be non-static and to have package access instead of private access. Is this a necessary change?

Core/src/org/sleuthkit/autopsy/datamodel/AttachmentNode.java
raman-bt: AttachmentNode gets used in MessageContentViewers which is in a different package.  I could have left the AttachmentNode in the ContentViewers package but I figured it fits better in the datamodel package with the rest of the nodes.

I'd also hoped to be able to reuse it in MediaViewer, until I realized that the "AttachmentNode" as used in the MediaViewer was in spirit - a "AttachmentThumbnailNode".

I certainly can be final though.

rcordovano: Sounds good, let's make it final.

rcordovano:
This PR has a merge conflict now, please resolve.

PR #5412, opened 2019-11-13
5733 implement xry report parsing

dannysmyda:


Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/AbstractSingleKeyValueParser.java
```diff
@@ -0,0 +1,165 @@
...
+abstract class AbstractSingleKeyValueParser implements XRYFileParser {
+    
+    private static final Logger logger = Logger.getLogger(AbstractSingleKeyValueParser.class.getName());
+
+    private static final char KEY_VALUE_DELIMITER = ':';
+
+    @Override
+    public void parse(XRYFileReader reader, Content parent) throws IOException, TskCoreException {
+        Path reportPath = reader.getReportPath();
+        logger.log(Level.INFO, String.format("INFO: Processing report at [ %s ]", reportPath.toString()));
```
rcordovano: Is putting "INFO" in the message intended to make it easier to grep for these messages during field trials? If so, I think that is a good idea, but using something more distinctive would be better.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYContactsFileParser.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.Content;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Parses XRY Contacts-Contacts files and creates artifacts.
```
rcordovano: Is "Contacts-Contacts" a typo?

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYCallsFileParser.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Parses XRY Calls files and creates artifacts.
+ */
+final class XRYCallsFileParser extends AbstractSingleKeyValueParser {
+
+    private static final Logger logger = Logger.getLogger(XRYCallsFileParser.class.getName());
+
+    //Human readable name of this parser.
+    private static final String PARSER_NAME = "XRY Calls";
```
rcordovano: Did Brian specify that each type of file parser should associates its name with the attributes it parses? If not, I think the precedents would indicate we use the name of the XRY DSP here - it is the "module."

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/AbstractSingleKeyValueParser.java
```diff
@@ -0,0 +1,165 @@
...
+                            + "namespace keyword [ %s ]. Applying to all key value pairs following it.", xryLine));
+                    namespace = xryLine.trim();
+                    continue;
+                }
+
+                //Find the XRY key on this line. Assume key is the value between
+                //the start of the line and the first delimiter.
+                int keyDelimiter = xryLine.indexOf(KEY_VALUE_DELIMITER);
+                if (keyDelimiter == -1) {
+                    logger.log(Level.SEVERE, String.format("Expected a key value "
```
rcordovano: I am not convinced that these errors rise to the level of SEVERE. A WARNING is the usual level for corrupt data. Are you thinking about this differently?

Also, I would add something distinct to all of the log messages (see prior comment) output by these parsers to make it easier to grep for the messages during field trials.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYFileReader.java
```diff
@@ -138,6 +175,23 @@ public String nextEntity() throws IOException {
...
+     * @throws NoSuchElementException if there are no more XRY entities to
+     * read.
+     */
+    public String peek() throws IOException {
+        if(hasNextEntity()) {
+            return xryEntity.toString();
+        } else {
+            throw new NoSuchElementException();
+        }
+    }
```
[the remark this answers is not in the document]
dannysmyda: This means the client has misused the API. They should have called hasNextEntity(). I'll add this to the comment header.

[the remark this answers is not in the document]
dannysmyda: Yes, we need to peek ahead while we search for message segments. I'm about to push up code that utilizes this method.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/AbstractSingleKeyValueParser.java
dannysmyda: I missed this. During testing all messages were system.out statements. I wanted to add a sense of realism. The intention of these messages is to give context in the log when a severe is encountered.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYContactsFileParser.java
dannysmyda: No that is the name of the report file, or at least the one Brian gave me.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYCallsFileParser.java
dannysmyda: No, I wasn't thinking much about this. I'll change these

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/AbstractSingleKeyValueParser.java
dannysmyda: I know. I was thinking of marking them as severe for the time being to make sure these were not missed. I can replace them with warning and prepend "XRY DSP" if you prefer.

rcordovano: That sounds good. We definitely want to make it easy to get feedback from the field trials..

PR #5418, opened 2019-11-14
5658 Geolocation waypoint filtering

kellykelly3:


Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
```diff
@@ -0,0 +1,490 @@
...
+        String query = String.format(GEO_ARTIFACT_QUERY,
+                BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DATETIME.getTypeID(),
+                BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DATETIME_CREATED.getTypeID());
+
+        // That are in the list of artifacts for the given data Sources
+        query += String.format("AND artifact_id IN(%s)", getWaypointListQuery(dataSources)); //NON-NLS
+        query += mostRecentQuery;
+
+        if (noTimeStamp) {
+            query = String.format("%s UNION %s", buildQueryForWaypointsWOTimeStamps(dataSources), query); //NON-NLS
```
wschaeferB: Currently items which do not have a timestamp are not being included when show all is selected unless the no time stamp checkbox is checked, however since the no time stamp checkbox is disabled when show all is selected this results in a user experience that is not desirable.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Waypoint.java
```diff
@@ -262,186 +262,6 @@ private static String getLabelFromArtifact(Map<BlackboardAttribute.ATTRIBUTE_TYP
 
         return attributeMap;
```
wschaeferB: import java.util.logging.Level and import org.sleuthkit.datamodel.SleuthkitCase are unused.

Core/src/org/sleuthkit/autopsy/geolocation/VerticalLabelUI.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
```
wschaeferB: Use standard header

Core/src/org/sleuthkit/autopsy/geolocation/HidingPane.java
```diff
@@ -0,0 +1,120 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
```
wschaeferB: Standard header.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -0,0 +1,355 @@
...
+        gridBagConstraints.gridy = 3;
+        gridBagConstraints.fill = java.awt.GridBagConstraints.HORIZONTAL;
+        gridBagConstraints.anchor = java.awt.GridBagConstraints.NORTHWEST;
+        gridBagConstraints.weightx = 1.0;
+        gridBagConstraints.weighty = 1.0;
+        gridBagConstraints.insets = new java.awt.Insets(0, 15, 0, 15);
+        add(checkboxPanel, gridBagConstraints);
+
+        try {
+            initCheckboxList();
```
wschaeferB: We need to initalize the list of checkboxes based on the current case at some other time.

If you close the case and open another it will still be populated with the data sources from the original case.  

Also as Richard points out in his email we can't be sure there is an open case when a top component is constructed and this panel is constructed in the top component constructor.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
kellykelly3: Good catch.  Thank you.

PR #5419, opened 2019-11-14
5734 implement xry dsp

dannysmyda:


Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    @Override
+    public JPanel getPanel() {
+        return configPanel;
+    }
+
+    @Override
+    public boolean isPanelValid() {
+        return true;
```
rcordovano: Shouldn't we do the validation that occurs in the run method here? It seems that at a minimum, we need to be sure that some folder path has been entered in the text field.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Processes the XRY folder the examiner selected. The heavy lifting 
+     * is handed off to a dedicated thread. This function will
+     * test the minimum requirements needed to successfully process the input.
+     */
```
rcordovano: I am being very nit-picky here about the use of the term "dedicated" here, but FYI, I want to be sure that you realize that the thread that will do the heavy is in a Swing thread pool. To me, "dedicated" implies you have set up a thread or thread pool specifically for this DSP.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
```diff
@@ -0,0 +1,243 @@
...
+    @Override
+    public void reset() {
+        //Clear the current selected file path.
+        configPanel.clearSelectedFilePath();
+    }
+
+    /**
+     * Relieves the EDT from add images to the case database and processing the
+     * XRY report files.
+     */
```
rcordovano: The grammar here is not quite spot on.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
```diff
@@ -0,0 +1,243 @@
...
+        @NbBundle.Messages({
+            "XRYDataSourceProcessor.unexpectedError=Internal error occurred while processing XRY report"
+        })
+        public void done() {
+            try {
+                LocalFilesDataSource newDataSource = get();
+                callback.done(DataSourceProcessorCallback.DataSourceProcessorResult.NO_ERRORS, 
+                        Lists.newArrayList(), Lists.newArrayList(newDataSource));
+            } catch (InterruptedException ex) {
+                //DSP was cancelled. Not an error.
```
rcordovano: It would probably be a good idea to log a warning here, since there may be incomplete data in the case now.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
dannysmyda: Okay, is there a way to pass along messages for why it is not a valid panel?

rcordovano: I'd take a look at what the other DSPs do.

If you address the other little comments, I will take this PR and we can send this to field trial without the more nuanced error handling.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/Bundle.properties-MERGED
```diff
@@ -0,0 +1,12 @@
+XRYDataSourceProcessor.dataSourceType=Import Tool Report
```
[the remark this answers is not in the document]
dannysmyda: From my DSP design doc, I have indication that it should be named "Import Tool Report". Revisiting the initial design story, it seems like this data source processor was intended to be generic. A user could select the type of tool they wanted to import from a drop down. Given that this panel only supports XRY right now, I'll change it to XRY Logical Report.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYDataSourceProcessor.java
dannysmyda: I see. I was able to move it and still show the messages.

PR #5435, opened 2019-11-20
5769 collect attachments for viber

markmckinnon:


InternalPythonModules/android/viber.py
```diff
@@ -364,3 +372,11 @@ def get_message_text(self):
 
     def get_thread_id(self):
         return str(self.result_set.getInt("thread_id")) 
+        
+    def get_file_attachment(self):
```
dannysmyda: Do you know if this field is always singular? If we don't already have some multi-attachment data, we could make some.

markmckinnon: The data that is there is singular,  If you can make multi attachments then we can look to see how they show up in the database.

dannysmyda:
Hmm, should this omit message content that starts with file://, since attachments are stored in the msg body (by viber, not us)? Perhaps Raman could address this later, when he addes multi-attachment support.

PR #5450, opened 2019-11-26
Japanese UI Translations 2019

ethteck:


ethteck:
I will remove the scripts I've added for setup / localization before this is merged. We are waiting on a final update in the next few days, and then I will edit the branch to remove these and request the merge.

PR #5453, opened 2019-11-27
Japanese UI Translations 2019

ethteck:
We are waiting on a final update in the next few days at which point I will request this be moved out of draft state.


ethteck:
@bcarrier @rcordovano I think what's best in the interest of time is for us to go ahead and merge this if possible. Then, if there are further changes tomorrow, I can create an additional PR. 

This PR consists of only changes to the JA bundle files.

PR #5455, opened 2019-11-27
5601 warn user if they already unpackaged a portable case

markmckinnon:


Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
```diff
@@ -261,6 +292,33 @@ private File locate7ZipExecutable() {
 
             return exeFile;
         }
+
+        /**
+         * Check to see if the case has been unpacked before
+         */
+        private String hasCaseBeenUnpackedBefore(String packedCaseName) {
```
dannysmyda: Can this be renamed to something like getPreviouslyUnpackedCaseFileName? I expected it to return a boolean given its name.

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
```diff
@@ -142,65 +149,89 @@ public void propertyChange(PropertyChangeEvent evt) {
         @NbBundle.Messages({
             "UnpackageWorker.doInBackground.errorFinding7zip=Could not locate 7-Zip executable",
             "UnpackageWorker.doInBackground.errorCompressingCase=Error unpackaging case",
-            "UnpackageWorker.doInBackground.canceled=Unpackaging canceled by user",})
+            "UnpackageWorker.doInBackground.canceled=Unpackaging canceled by user",
+            "UnpackageWorker.doInBackground.previousSeenCase=Case {0} has been previously opened do you want to open it again?"})
```
dannysmyda: Could we change this to specify that a portable case by the same name has already been extracted? We don't know if we've opened the requested case unless we examined its contents.

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
```diff
@@ -142,65 +149,89 @@ public void propertyChange(PropertyChangeEvent evt) {
...
-                Process process = procBuilder.start();
-
-                while (process.isAlive()) {
-                    if (this.isCancelled()) {
-                        setDisplayError(Bundle.UnpackageWorker_doInBackground_canceled());
-                        return null;
+            // Check to see if this case has been already opened before
+            String caseUnpackedBefore = hasCaseBeenUnpackedBefore(packagedCase);
+            if (caseUnpackedBefore != null) {
+                if (MessageNotifyUtil.Message.confirm(Bundle.UnpackageWorker_doInBackground_previousSeenCase(packagedCase))) {
```
dannysmyda: Shouldn't it fall through to unzipping the portable case if I say no? Granted I haven't tested the code yet, I'm still only reading it. The if else block implies it will only open the previously seen case.

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
bcarrier: Agree on the renaming and we've been trying to return Optional more instead of 'null' to prevent NPEs.

markmckinnon: Renamed.

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
markmckinnon: Yes. Fixed

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
markmckinnon: Return empty string and check for it

Core/src/org/sleuthkit/autopsy/casemodule/UnpackagePortableCaseProgressDialog.java
markmckinnon: Changed to "Case with name {0} has been previously opened do you want to open it again?"

Core/src/org/sleuthkit/autopsy/casemodule/Bundle.properties-MERGED
```diff
@@ -354,6 +354,7 @@ UnpackagePortableCaseProgressDialog.title.text=Unpackage Portable Case Progress
 UnpackageWorker.doInBackground.canceled=Unpackaging canceled by user
 UnpackageWorker.doInBackground.errorCompressingCase=Error unpackaging case
 UnpackageWorker.doInBackground.errorFinding7zip=Could not locate 7-Zip executable
+UnpackageWorker.doInBackground.previousSeenCase=Case {0} has been previously opened do you want to open it again?
```
dannysmyda: This is out of sync with the Bundle messages string.

dannysmyda:
This PR has been tested. Once the Bundle.properties-MERGED file is updated, it is good to go.

Core/src/org/sleuthkit/autopsy/casemodule/Bundle.properties-MERGED
markmckinnon: Pushed up file

PR #5459, opened 2019-12-02
5804 No timeline data if some ingest modules not run

rcordovano:
* Made all timeline filters disabled by default. This way the lack of file types if ingest has not been run does not produce an empty display by default.
* Changed filter display names as desired by Brian Carrier.
* Changed reset button label from "Default" to "Reset" as desired by Brian Carrier. 

See also: https://github.com/sleuthkit/sleuthkit/pull/1704.


esaunders:
I've approved but there is one issue that Codacy is complaining about.

PR #5460, opened 2019-12-02
5817, 5841, 5842, 5843 geolocation window bug fixes

kellykelly3:
This PR fixes four minor issues with the Geolocation window.


Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -135,7 +135,7 @@ GeoFilter getFilterState() throws GeoLocationUIException {
     private void initCheckboxList() throws TskCoreException {
         final SleuthkitCase sleuthkitCase = Case.getCurrentCase().getSleuthkitCase();
 
-        checkboxPanel.clearList();
+//        checkboxPanel.clearList();
```
wschaeferB: remove instead of comment out

wschaeferB:
There is a bug introduced in this PR where you get a 'case is closed' TskCoreException if you open a case and open the geolocation tool, then open a different case (the 2nd case must have waypoints, the first it does not matter) when opening the geolocation tool in the 2nd case you get the exception, this issue does not exist in the current develop branch.

wschaeferB:
As for the refreshing the data source list, is only updating the list when the user clicks refresh after a new waypoint is added the behavior we want? Or do we want to update the the list when a data source is added or ingest is finished, currently we have different behaviors in different locations as discussed in JIRA 5720 and I just want to make sure this behavior is the behavior we desire.

kellykelly3:
> As for the refreshing the data source list, is only updating the list when the user clicks refresh after a new waypoint is added the behavior we want? Or do we want to update the the list when a data source is added or ingest is finished, currently we have different behaviors in different locations as discussed in JIRA 5720 and I just want to make sure this behavior is the behavior we desire.

I went for consistency with CVT window.

PR #5461, opened 2019-12-02
5767: Add attachments to Skype messages.

raman-bt:


InternalPythonModules/android/skype.py
```diff
@@ -425,12 +434,12 @@ def get_message_text(self):
         content = self.result_set.getString("content")
 
         if content is not None:
-            file_path = self.result_set.getString("device_gallery_path")
-
-            #if a file name and file path are associated with a message, append it
-            if file_path is not None:
-                return general.appendAttachmentList(content, [file_path]) 
-
+##            file_path = self.result_set.getString("device_gallery_path")
```
dannysmyda: Could we opt instead to just delete this? It is no longer a working solution.

PR #5486, opened 2019-12-10
5858 adjust xry parsers

dannysmyda:
In addition to making adjustments, some light refactoring is included in this PR:
1) Optional is returned instead of null
2) The process for recognizing XRY keys and making blackboard attributes from them was improved. Rather than doing string matches, ENUMs were created to ensure that there are no accidental typos. This is the bulk majority of the new line additions.


Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYCallsFileParser.java
```diff
@@ -161,15 +256,60 @@ private String removeDateTimeLocale(String dateTime) {
...
+         * the date time string are reversed:
+         * 
+         *      UTC+4 AM 1:23:54 1/3/1990
+         * 
+         * The java time package will correctly parse this date time format.
+         */
+        String reversedDateTime = reverseOrderOfDateTimeComponents(dateTimeWithoutLocale);
+        /**
+         * Furthermore, the DateTimeFormatter's timezone offset letter ('O') does
+         * not recognized UTC but recognizes GMT. According to
```
rcordovano: recognized => recognize

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYCallsFileParser.java
```diff
@@ -39,101 +37,198 @@
...
+         * Fetches the enum type for the given XRY key.
+         * 
+         * It is assumed that XRY key string is recognized. Otherwise,
+         * an IllegalArgumentException is thrown. Test all membership
+         * with contains() before hand.
+         * 
+         * @param xryKey
+         * @return 
+         */
+        public static XRY_KEY fromName(String xryKey) {
```
rcordovano: You might want to use the existing Enum method valueOf. I don't feel strongly about this.

dannysmyda: I changed this to 'fromDisplayName', indicating the that string being passed in will differ from the enum name.

dannysmyda:
The new cyclomatic complexity issues are a result of me rearranging the ordering of the switch statement. I think Codacy believes they are new methods.

PR #5488, opened 2019-12-11
5709: Email attachments

raman-bt:
- Use CommunicationsArtifactsHelper to add the attachments to email messages.
 - Updated the DirectoryTree & DataResultsTable to not show message artifacts for 'new' cases, but continue to show them for older cases.

### DIFF
--- a/thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
+++ b/thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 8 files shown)


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/ThunderbirdMboxFileIngestModule.java
```diff
@@ -129,6 +134,15 @@ public ProcessResult process(AbstractFile abstractFile) {
             logger.log(Level.WARNING, null, ex);
         }
 
+        try {
+            communicationArtifactsHelper = new CommunicationArtifactsHelper(currentCase.getSleuthkitCase(),
+                    EmailParserModuleFactory.getModuleName(), abstractFile, Account.Type.EMAIL);
+        } catch (TskCoreException ex) {
+            logger.log(Level.SEVERE, String.format("Failed to create CommunicationArtifactsHelper for file %s", abstractFile.getName()), ex);
```
rcordovano: In general, I think we should log object ID, not just names, to identify files so that there is no ambiguity.

PR #5495, opened 2019-12-13
5883 & 5882 Geolocation custom artifact handling

kellykelly3:


APriestman:
Can you add the null check for LastKnownWaypoint.getLabelFromArtifact() to this PR?

PR #5516, opened 2019-12-18
5875: For an attachment file, the Message Viewer should show the mess…

raman-bt:
…age it was attached to.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/contentviewers/MessageContentViewer.java
+++ b/Core/src/org/sleuthkit/autopsy/contentviewers/MessageContentViewer.java
@@ -39,6 +39,8 @@ import org.openide.nodes.Children;
 import org.openide.nodes.Node;
 import org.openide.util.NbBundle;
 import org.openide.util.lookup.ServiceProvider;
+import org.sleuthkit.autopsy.casemodule.Case;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/contentviewers/MessageContentViewer.java
```diff
@@ -498,10 +502,54 @@ private static boolean isMessageArtifact(BlackboardArtifact nodeArtifact) {
...
+     * @param node Node to check.
+     * @return Blackboard artifact for the node, null if there isn't any.
+     */
+    private BlackboardArtifact getNodeArtifact(Node node) {
+        BlackboardArtifact nodeArtifact = node.getLookup().lookup(BlackboardArtifact.class);
+
+        if (nodeArtifact == null) {
+            try {
+                SleuthkitCase tskCase = Case.getCurrentCaseThrows().getSleuthkitCase();
+                AbstractFile file = node.getLookup().lookup(AbstractFile.class);
```
rcordovano: Is it guaranteed that we will have an  AbstractFile in Node?

raman-bt: No, there's no guarantee..  I did have a null check that got removed in the shuffle.  I'll update it.

PR #5568, opened 2020-01-08
5881 Streamline Gold Updates Part Two

dannysmyda:
This keeps the build script from exiting if the diffs do not pass.


dannysmyda:
Changes are currently untested.

dannysmyda:
Changes have now been tested. Appears to be working, however I simulated failure to test an essential code path. We will just have to wait and see what happens when Jenkins failures naturally.

PR #5578, opened 2020-01-10
5922-ExtractRecycleBin.parseIFile-makes-assumptions-about-size-of-data-read

markmckinnon:
Add $RECYCLE.BIN as the directory to look for $I files.

### DIFF
--- a/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
+++ b/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
@@ -71,6 +71,8 @@ final class ExtractRecycleBin extends Extract {
     private static final Logger logger = Logger.getLogger(ExtractRecycleBin.class.getName());
 
     private static final String RECYCLE_BIN_ARTIFACT_NAME = "TSK_RECYCLE_BIN"; //NON-NLS
+    
+    private static final String RECYCLE_BIN_DIR_NAME = "$RECYCLE.BIN"; //NON-NLS
 
     private static final int V1_FILE_NAME_OFFSET = 24;
     private static final int V2_FILE_NAME_OFFSET = 28;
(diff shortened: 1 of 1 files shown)


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRecycleBin.java
```diff
@@ -266,37 +269,34 @@ private void popuplateDeletedDirectory(SleuthkitCase skCase, AbstractFile parent
...
-     * <tr><td>28</td><td>var</td><td>File Name</td></tr>
-     * </table>
-     *
+     * Offset  Size  Description
+     *   0       8     Header
+     *   8       8     File Size
+     *  16       8     Deleted TimeStamp
+     *  24       4     File Name Length
+     *  28     var     File Name
+     *
```
rcordovano: Why remove the formatting? Did this not build /format properly in doxygen? Does it build /format properly now?

markmckinnon: I will check it out.

rcordovano: Install Doxygen.
http://www.doxygen.nl/download.html
Download and run the Windows installer. 

Run doxygen for Sleuthkit in ~sleuthkit/bindings/java/doxygen, then run doxygen for Autopsy in ~autopsy/docs/doxygen, ~autopsy/docs/doxygen-user, and ~autopsy/docs/doxygen-dev

markmckinnon: From what I see that table never shows up in the docs.  I also checked online in the 4.13 docs and I do not see it as well.

rcordovano: Actually, that makes sense - this is not a public class. I'll accept the PR as is.

PR #5579, opened 2020-01-10
1379 & 1389 drone dat ingest module

kellykelly3:


kellykelly3:
I screwed up the sleuthkit PR that goes with this PR by adding changes for the 1390 PR.

See https://github.com/sleuthkit/autopsy/pull/5595 which includes all the code from this PR.

PR #5600, opened 2020-01-16
1391-Prevented BlackboardArtifactNode from showing attributes of type JSON

kellykelly3:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -841,7 +841,8 @@ private void fillPropertyMap(Map<String, Object> map, BlackboardArtifact artifac
                         || attributeTypeID == ATTRIBUTE_TYPE.TSK_TAGGED_ARTIFACT.getTypeID()
                         || attributeTypeID == ATTRIBUTE_TYPE.TSK_ASSOCIATED_ARTIFACT.getTypeID()
                         || attributeTypeID == ATTRIBUTE_TYPE.TSK_SET_NAME.getTypeID()
-                        || attributeTypeID == ATTRIBUTE_TYPE.TSK_KEYWORD_SEARCH_TYPE.getTypeID()) {
+                        || attributeTypeID == ATTRIBUTE_TYPE.TSK_KEYWORD_SEARCH_TYPE.getTypeID() 
+                        || attribute.getValueType() == BlackboardAttribute.TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.JSON) {
```
raman-bt: I'm ok with this change and it tested fine.  

Codacy however does complain and thinks its a new issue, although it's not.  May be we put a 'continue' in that empty block  - which is what is effectively happening? That should satisfy Codacy, I think.

PR #5602, opened 2020-01-16
5971 - Cache Unique Path

dannysmyda:


rcordovano:
> DataContentPanel.setNode() still calls Content.getUniquePath() on the EDT when selecting blackboard artifact nodes. Not sure if this should be a separate story, but marking as request changes for the time being.

Yes, we'll create a separate story for this.

PR #5610, opened 2020-01-21
5465 case uco reference local files

dannysmyda:
This includes changes from JIRA-5718.


eugene7646:
@dannysmyda looks like a lot of file names have changed, so we lost history for those files, making it look like each file is brand new and there are 900 lines of changes. That's a lot of code to review and I'd rather not spend time reviewing code that hasn't been touched. Can you narrow it down for me a little bit, as to maybe which are the main methods I should look at?

rcordovano:
There is a Codacy issue that can be easily resolved.

PR #5617, opened 2020-01-28
1404 - Changed drone module name and description

kellykelly3:


Core/src/org/sleuthkit/autopsy/modules/drones/DroneIngestModuleFactory.java
```diff
@@ -33,8 +33,8 @@
 public class DroneIngestModuleFactory extends IngestModuleFactoryAdapter {
 
     @Messages({
-        "DroneIngestModule_Name=Drone",
-        "DroneIngestModule_Description=Description"
+        "DroneIngestModule_Name=Drone Analyzer",
+        "DroneIngestModule_Description=Analyzes file generated by drones."
```
rcordovano: I think "files" should be plural in the module description.

kellykelly3: Correct.

PR #5618, opened 2020-01-28
6004 Check for empty database paths.

APriestman:


Experimental/src/org/sleuthkit/autopsy/experimental/configuration/SharedConfiguration.java
```diff
@@ -1243,9 +1243,9 @@ private void downloadHashDbSettings(File remoteFolder) throws SharedConfiguratio
             HashDbManager hashDbManager = HashDbManager.getInstance();
             hashDbManager.loadLastSavedConfiguration();
             for (HashDbManager.HashDb hashDb : hashDbManager.getAllHashSets()) {
-                if (hashDb.hasIndexOnly()) {
+                if (hashDb.hasIndexOnly() && (!hashDb.getIndexPath().isEmpty())) {
```
rcordovano: What does it mean for a hash database to have an index only, but no index path?

APriestman: Central repo DBs are like that - we set them as index only but they have no path to that index on disk. The second test that databasePath is non-empty isn't currently necessary but seemed safer to include in case we ever change this (and we'll never want to copy an empty path...)

PR #5622, opened 2020-01-30
1400 geolocation map mouse click improvements

kellykelly3:


kellykelly3:
I am not going to address the remaining codacy issues.  I did not write the search function that have complexity and would take some careful thought to simplify.  I would rather not take that risk.  

Codacy is also complaining about the search function taking a TreeSet instead of a Set, again I did not write that function.  search uses a function that is specific to TreeSet and not in the Set interface.  So I think we should let that one stand.

Core/src/org/sleuthkit/autopsy/geolocation/KdTree.java
```diff
@@ -247,258 +163,208 @@ public boolean contains(T value) {
...
-            list.addAll(getTree(root.greater));
-        }
-
-        return list;
-    }
-
-    /** 
-     * Searches the K nearest neighbor.
+     * @param numNeighbors Number of neighbors to retrieve. Can return more than
+     *                     K, if last nodes are equal distances.
```
raman-bt: "can return more than numNeighbors"

Core/src/org/sleuthkit/autopsy/geolocation/KdTree.java
```diff
@@ -507,173 +373,258 @@ public EuclideanComparator(XYZPoint point) {
...
-            this.z = 0;
+        public XYZPoint(Double latitude, Double longitude) {
+            x = latitude;
+            y = longitude;
+            z = 0;
         }
 
         /**
-         * Default constructor
+         * Returns the x\latitude value for the point.
```
raman-bt: Is that a backslash for a reason? Should it simply be x/latitude? makes it hard to read.

Core/src/org/sleuthkit/autopsy/geolocation/KdTree.java
```diff
@@ -507,173 +373,258 @@ public EuclideanComparator(XYZPoint point) {
...
          *
-         * @param x
-         * @param y
-         * @param z
+         * @return x\Latitude value for the point
```
raman-bt: same comment regarding backslash.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
```diff
@@ -434,16 +439,16 @@ private Point getLocationForDetailsPopup() {
...
+            return new ArrayList<>();
         }
 
         // Convert the mouse click location to latitude & longitude
-        GeoPosition geopos = mapViewer.getTileFactory().pixelToGeo(mouseClickPoint, mapViewer.getZoom());
+        GeoPosition geopos = mapViewer.convertPointToGeoPosition(clickPoint);
 
         // Get the 5 nearest neightbors to the point
-        Collection<MapWaypoint> waypoints = waypointTree.nearestNeighbourSearch(10, MapWaypoint.getDummyWaypoint(geopos));
+        Collection<MapWaypoint> waypoints = waypointTree.nearestNeighbourSearch(1, MapWaypoint.getDummyWaypoint(geopos));
```
raman-bt: Are we getting 5 nbrs or 1? comment is inconsistent with actual param.

PR #5627, opened 2020-01-31
5966-open-source-GPX-module-and-update-how-it-makes-artifacts

markmckinnon:
GPX internal python module update for open source and geoartifacthelper

### DIFF
--- a/InternalPythonModules/GPX_Module/GPX_Parser_Module.py
+++ b/InternalPythonModules/GPX_Module/GPX_Parser_Module.py
@@ -37,6 +37,8 @@ from org.sleuthkit.datamodel import BlackboardArtifact
 from org.sleuthkit.datamodel import BlackboardAttribute
 from org.sleuthkit.datamodel import TskCoreException
 from org.sleuthkit.datamodel.blackboardutils import GeoArtifactsHelper
+from org.sleuthkit.datamodel.blackboardutils.attributes import GeoWaypoint
+from org.sleuthkit.datamodel.blackboardutils.attributes import GeoTrackPoints
 from org.sleuthkit.autopsy.datamodel import ContentUtils
 from org.sleuthkit.autopsy.ingest import IngestModule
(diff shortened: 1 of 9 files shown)


eugene7646:
@markmckinnon @rcordovano is this all brand new code? The PR shows 4,500 lines of new code and not a single line of deleted code. If it's all new code, it will be quite an effort to review it all...

InternalPythonModules/GPX_Module/GPX_Parser_Module.py
```diff
@@ -0,0 +1,265 @@
...
+                    except Blackboard.BlackboardException as e:
+                        if GPXParserDataSourceIngestModuleFactory.debuglevel: self.log(Level.SEVERE, "GPX: Error using geo artifact helper with blackboard  for waypoints" )
+
+                if GPXParserDataSourceIngestModuleFactory.debuglevel: self.log(Level.INFO, "GPX: ROUTES")
+                for route in gpx.routes:    
+                    firstTimeThru = 0
+                    startingPoint = list()
+                    endingPoint = list()
+                    for point in route.points:
+                        # If first time in loop only populate starting point
```
eugene7646: What is the purpose of this logic? It seems like the logic will break if we have a route that has only 1 point in it, because the end point will not be set and the artifact will not be created?

markmckinnon: A route needs to have a starting and ending point.  If there is only a starting point and no ending point then it is not a route.  Does that make sense?

eugene7646: Oh it definitely makes sense. I'm just thinking that if somehow only one GPS point got saved as a "route" (for whatever reason, bug, etc), it might still be of value to examiners?

markmckinnon:
@eugene7646 I tested the gpxpy code and if there is only a starting point and no ending point in the gpx file it will still produce that point.  I have made changes so that if it happens it will use the starting point as the ending point then create the point so it is not lost.  Nice catch.

PR #5637, opened 2020-02-06
1408 & 1409: Added support for TSK_GEO_WAYPOINTS attribute in TSK_GPS_ROUTE

kellykelly3:


Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Route.java
```diff
@@ -25,14 +25,17 @@
...
 import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.blackboardutils.attributes.GeoTrackPoints.GeoWaypoints;
+import org.sleuthkit.datamodel.blackboardutils.attributes.GeoWaypoint;
 
 /**
  * A Route represents a TSK_GPS_ROUTE artifact which has a start and end point
- * however the class was written with the assumption that routes may have
- * more that two points.
+ * however the class was written with the assumption that routes may have more
+ * that two points.
```
raman-bt: "more than"

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Route.java
```diff
@@ -82,19 +84,62 @@
...
+     * Gets the route waypoint attributes from the map and creates the list of
+     * route waypoints.
+     *
+     * @param artifact     Route artifact
+     * @param attributeMap Map of artifact attributes
+     *
+     * @throws GeoLocationDataException
+     */
+    @Messages({
+        "Route_point_label=Route point for route"
```
raman-bt: should this be "Route points...."?

rcordovano: Isn't a route made up of way points?

PR #5638, opened 2020-02-07
1385 - Added doxygen page for drone analzyer module.

kellykelly3:


docs/doxygen-user/drone.dox
```diff
@@ -0,0 +1,26 @@
+/*! \page drone_page Drone Analyzer
+
+\section drone_overview Overview
+
+The Drone Analyzer module allows you to analyze files from a drone.
+
+Currently, the Drone Analyzer module, works on images obtained from the internal SD card found in the following DJI drone models:
```
APriestman: The comma after "module" is unnecessary

docs/doxygen-user/drone.dox
```diff
@@ -0,0 +1,26 @@
...
+- Phantom 4
+- Phantom 4 Pro
+- Inspire 1
+- Inspire 2
+- Mavic Pro
+- Mavic Air
+
+\section drone_config Running the Module
+
+To enable the Drone Analyzer ingest module select the checkbox in the Ingest Modules setting screen.
```
APriestman: You could reference "ingest_configure" here. You can use quotes to use your own text instead of the heading you're linking to. For example: select the checkbox in the \ref ingest_configure "ingest modules configuration screen"

docs/doxygen-user/drone.dox
```diff
@@ -0,0 +1,26 @@
...
+- Mavic Pro
+- Mavic Air
+
+\section drone_config Running the Module
+
+To enable the Drone Analyzer ingest module select the checkbox in the Ingest Modules setting screen. 
+
+\section drone_results Viewing Results
+
+Results are show in the Results tree under "Extracted Content".  GPS results can also be seen in the \ref geolocation_page window and the KML Report.
```
APriestman: I would probably add a screenshot of the artifacts with one of them selected so you can see what the result viewer should look like.

PR #5644, opened 2020-02-11
5342-Create-artifacts-for-recently-used-Windows-Media-Player-files-from-regripper-output

markmckinnon:
Add mpmru parsing, change where artifacts are added in adobemru so it does not get checked every time.

### DIFF
--- a/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
+++ b/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
@@ -1244,7 +1244,7 @@ class ExtractRegistry extends Extract {
     }
     
      /**
-     * Create recently used artifacts from mpmru records
+     * Create recently used artifacts to parse the mpmru records
      * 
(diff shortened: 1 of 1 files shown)


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1235,10 +1237,51 @@ private void parseAdobeMRUList(String regFileName, AbstractFile regFile, Buffere
...
+    }
+    
+     /**
+     * Create recently used artifacts from mpmru records
+     * 
+     * @param regFileName name of the regripper output file
+     * 
+     * @param regFile registry file the artifact is associated with
+     * 
+     * @param reader buffered reader to parse adobemru records
```
raman-bt: to parse the "mpmru" records

markmckinnon: Changed comment

PR #5648, opened 2020-02-13
6040 file discovery documents

wschaeferB:


Core/src/org/sleuthkit/autopsy/filequery/FileSearch.java
```diff
@@ -239,6 +247,80 @@ static SearchResults runFileSearchDebug(String userName,
...
+     */
+    private static TextSummarizer getLocalSummarizer() throws IOException {
+        Collection<? extends TextSummarizer> summarizers
+                = Lookup.getDefault().lookupAll(TextSummarizer.class
+                );
+        if (!summarizers.isEmpty()) {
+            summarizerToUse = summarizers.iterator().next();
+            return summarizerToUse;
+        }
+        throw new IOException("No summarizers found");
```
APriestman: Perhaps it would be better to return null instead of throwing an exception? Nothing has really gone wrong. And then you wouldn't need that empty catch block.

PR #5650, opened 2020-02-14
environment variable setup script for development

gdicristofaro:


dannysmyda:
The code looks good. I'll test it with the new PC that I need to set up.

developers/envvarsetup.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def add_path(paths):
+    print("Adding to path: {0}".format(paths))
+    # work around for command prompt to access PATH via %PATH%
+    cmd = "cmd.exe /k SETX PATH \"%PATH%;{0}\" /M & exit".format(path_env_sep.join(paths))
+    os.system(cmd)
+
+
+set_var("JAVA_HOME", open_jdk_home)
```
dannysmyda: All uses of foo + sep + bar can be replaced with os.path.join which will handle formatting paths:
python_escape
python_windows

PR #5651, opened 2020-02-18
5343 create artifacts for files which were recently used by other applications of interest

markmckinnon:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1238,7 +1250,7 @@ private void parseAdobeMRUList(String regFileName, AbstractFile regFile, Buffere
                 line = line.trim();
             }
         }    
-        if (bbartifacts != null) {
+        if (bbartifacts.isEmpty()) {
```
wschaeferB: good catch on realizing it can't be null but could be empty fixing an existing bug.

However, I think we want a not empty here like you have below.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1279,7 +1291,225 @@ private void parseMediaPlayerMRUList(String regFileName, AbstractFile regFile, B
...
+     * 
+     * @param regFileName name of the regripper output file
+     * 
+     * @param regFile registry file the artifact is associated with
+     * 
+     * @param reader buffered reader to parse adobemru records
+     * 
+     * @throws FileNotFound and IOException
+     */
+    private void parseGenericMRUList(String regFileName, AbstractFile regFile, BufferedReader reader) throws FileNotFoundException, IOException {
```
wschaeferB: The regFileName argument of all of these methods seems to be unused now, and could be removed.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
markmckinnon: Fixed

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
markmckinnon: Removed

PR #5668, opened 2020-03-02
6035 cr psql default

gdicristofaro:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbChoice.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.centralrepository.datamodel;
+
+/**
+ * the database choices available for central repo
+ */
+public class CentralRepoDbChoice {
```
APriestman: Could this be an enum? Then we could use case statements later.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbChoice.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public static final CentralRepoDbChoice SQLITE = new CentralRepoDbChoice("Sqlite", "SQLite", CentralRepoPlatforms.SQLITE);
+
+    public static final CentralRepoDbChoice POSTGRESQL_MULTIUSER = 
+        new CentralRepoDbChoice("PostgreSQL_Multiuser", "PostgreSQL using multi-user settings", CentralRepoPlatforms.POSTGRESQL);
+
+    public static final CentralRepoDbChoice POSTGRESQL_CUSTOM = 
+        new CentralRepoDbChoice("PostgreSQL", "Custom PostgreSQL", CentralRepoPlatforms.POSTGRESQL);
+
+    public static final CentralRepoDbChoice[] CHOICES = new CentralRepoDbChoice[]{
```
APriestman: Perhaps UI_CHOICES? To make it more clear

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbManager.java
```diff
@@ -34,6 +35,45 @@
 
     private static final String CENTRAL_REPO_DB_NAME = "central_repository";
 
+
+
+    private static volatile CentralRepoDbChoice SAVED_CHOICE = null;
```
APriestman: I think since this isn't a constant it should be normal lower case

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbChoice.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package org.sleuthkit.autopsy.centralrepository.datamodel;
+
+/**
+ * the database choices available for central repo
+ */
+public class CentralRepoDbChoice {
+    public static final CentralRepoDbChoice DISABLED = new CentralRepoDbChoice("Disabled", CentralRepoPlatforms.DISABLED);
+
+    public static final CentralRepoDbChoice SQLITE = new CentralRepoDbChoice("Sqlite", "SQLite", CentralRepoPlatforms.SQLITE);
```
APriestman: If the second string is displayed in the UI, it should be in a bundle message

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoPostgresSettingsUtil.java
```diff
@@ -0,0 +1,177 @@
...
+        
+        CaseDbConnectionInfo muConn;
+        try {
+            muConn = UserPreferences.getDatabaseConnectionInfo();
+        } catch (UserPreferencesException ex) {
+            LOGGER.log(Level.SEVERE, "Failed to import settings from multi-user settings.", ex);
+            return settings;
+        }
+        
+        handleTry(() -> settings.setHost(valOrDefault(muConn.getHost(), DEFAULT_HOST)));
```
APriestman: Should we do more than just log an error if these parameters are invalid? There may be empty fields in the settings object. If this is what we want to do, add a comment that this is what's happening.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoPostgresSettingsUtil.java
```diff
@@ -0,0 +1,177 @@
...
+                else {
+                    return retVal;
+                }
+            }
+        } catch (NumberFormatException ex) {
+            return defaultVal;
+        }
+    }
+    
+    private static void handleTry(TryHandler handler) {
```
APriestman: This could use a comment and a more descriptive name. As it stands, logException() would be more accurate.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoPostgresSettingsUtil.java
```diff
@@ -0,0 +1,177 @@
...
+        Map<String, String> map = new HashMap<String, String>();
+        map.put(HOST_KEY, settings.getHost());
+        map.put(PORT_KEY, Integer.toString(settings.getPort()));
+        map.put(DBNAME_KEY, settings.getDbName());
+        map.put(BULK_THRESHOLD_KEY, Integer.toString(settings.getBulkThreshold()));
+        map.put(USER_KEY, settings.getUserName());
+        try {
+            map.put(PASSWORD_KEY, TextConverter.convertTextToHexText(settings.getPassword())); // NON-NLS
+        } catch (TextConverterException ex) {
+            LOGGER.log(Level.SEVERE, "Failed to convert password from text to hex text.", ex);
```
APriestman: In the previous section this is a warning and the default password is used. Why is it different here? Also if the error is "SEVERE" we probably shouldn't just log it and continue on.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresSettingsLoader.java
```diff
@@ -0,0 +1,66 @@
...
+package org.sleuthkit.autopsy.centralrepository.datamodel;
+
+/**
+ * interface to load or save postgres settings
+ */
+public interface PostgresSettingsLoader {
+    PostgresConnectionSettings loadSettings();
+    void saveSettings(PostgresConnectionSettings settings);
+    
+    PostgresSettingsLoader CUSTOM_LOADER = new Custom();
```
APriestman: These names could be a bit more descriptive.

Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
```diff
@@ -75,6 +75,7 @@
     public static final String SHOW_ONLY_CURRENT_USER_TAGS = "ShowOnlyCurrentUserTags";
     public static final String HIDE_SCO_COLUMNS = "HideCentralRepoCommentsAndOccurrences"; //The key for this setting pre-dates the settings current functionality //NON-NLS
     public static final String DISPLAY_TRANSLATED_NAMES = "DisplayTranslatedNames";
+    private static final boolean DISPLAY_TRANSLATED_NAMES_DEFAULT = true;
```
APriestman: Why is this in this PR?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresCentralRepoSettings.java
```diff
@@ -377,95 +305,72 @@ public void setPort(int port) throws CentralRepoException {
...
-     * @return the JDBC_BASE_URI
-     */
-    String getJDBCBaseURI() {
-        return JDBC_BASE_URI;
+    @Override
+    public DatabaseTestResult testStatus() {
+        if (verifyConnection()) {
+            if (verifyDatabaseExists()) {
+                if (verifyDatabaseSchema()) {
+                    return DatabaseTestResult.TESTEDOK;
```
APriestman: I don't think you named this, but an underscore would really help readability for "TESTEDOK"

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoPostgresSettingsUtil.java
gdicristofaro: My goal for this and the other thing you mentioned in this file was to preserve the the same behavior here: https://github.com/sleuthkit/autopsy/blob/develop/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PostgresCentralRepoSettings.java#L73.  Would it be worth changing right now or should this wait for another time?

Core/src/org/sleuthkit/autopsy/core/UserPreferences.java
gdicristofaro: Sorry about that.  I'll correct it.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbChoice.java
gdicristofaro: nevermind.  I got over my mental block.  I think this is a lot more ergonomic.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoPostgresSettingsUtil.java
APriestman: I don't think you've quite preserved that logic, though I may be reading it wrong. It seems like in your new code if the exception gets thrown the field doesn't get saved to the default.

gdicristofaro: Oh okay.  I'll fix that.

APriestman:
Please address some of the codacy complaints

Core/src/org/sleuthkit/autopsy/datamodel/AbstractAbstractFileNode.java
```diff
@@ -105,7 +105,7 @@
                             this.content.getName(), this.content.getId()), ex);
         }
 
-        if (UserPreferences.displayTranslatedFileNames()) {
+        if (TextTranslationService.getInstance().hasProvider() && UserPreferences.displayTranslatedFileNames()) {
```
APriestman: Still have a translation change here.

Core/src/org/sleuthkit/autopsy/centralrepository/optionspanel/GlobalSettingsPanel.java
```diff
@@ -33,33 +33,58 @@
...
 import org.sleuthkit.autopsy.centralrepository.datamodel.PostgresCentralRepoSettings;
 import org.sleuthkit.autopsy.centralrepository.datamodel.SqliteCentralRepoSettings;
+import java.awt.Component;
+import java.util.logging.Level;
 
 /**
  * Main settings panel for the Central Repository
  */
 @SuppressWarnings("PMD.SingularField") // UI widgets cause lots of false positives
 public final class GlobalSettingsPanel extends IngestModuleGlobalSettingsPanel implements OptionsPanel {
```
APriestman: Would it be possible to implement PropertyChangeListener here instead of writing your own?

gdicristofaro: Sure.  I will fix that.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbManager.java
```diff
@@ -216,38 +365,45 @@ public boolean wasConfigurationChanged() {
...
+        if (selectedDbChoice == CentralRepoDbChoice.DISABLED)
+            return null;
+        
+        throw new CentralRepoException("Unknown database type: " + selectedDbChoice);
     }
 
+    /**
+     * create central repo database if it does not already exist
+     * @return      true if successful; false if unsuccessful
+     * @throws CentralRepoException     if
```
APriestman: Random "if" here

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbManager.java
```diff
@@ -216,38 +365,45 @@ public boolean wasConfigurationChanged() {
...
+        if (selectedDbChoice == CentralRepoDbChoice.SQLITE)
+            return new RdbmsCentralRepoFactory(CentralRepoPlatforms.SQLITE, dbSettingsSqlite);
+        if (selectedDbChoice == CentralRepoDbChoice.DISABLED)
+            return null;
+        
+        throw new CentralRepoException("Unknown database type: " + selectedDbChoice);
     }
 
+    /**
+     * create central repo database if it does not already exist
```
APriestman: Can you switch to capitalizing your comments per Richard's style email?

PR #5669, opened 2020-03-02
6013: Insert row in CR accounts table when accounts are created in Ca…

raman-bt:
…se database.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
@@ -1,0 +1,136 @@
+/*
+ * Central Repository
+ *
+ * Copyright 2020 Basis Technology Corp.
+ * Contact: carrier <at> sleuthkit <dot> org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.autopsy.centralrepository.datamodel;
+
+import org.sleuthkit.datamodel.Account;
+
+
+/**
+ * This class abstracts an Account as stored in the CR database.
+ */
+public final class CRAccount {
+    
+	// primary key in the Accounts table in CR database
+	private final long account_id;
+
+	private final CRAccountType crAccountType;
+	private final String typeSpecificID;
+        
+    /**
+     * A CRAccounType encapsulates an account type and the correlation type
+     * that it maps to.
+     */
+    public static final class CRAccountType {
+
+        // id is the primary key in the account_types table
+        private final int crAccountTypeID;
+        private final Account.Type acctType;
+        private final int correlationTypeID;
+
+        CRAccountType(int acctTypeID, Account.Type acctType, int correlation_type_id) {
+            this.acctType = acctType;
+            this.correlationTypeID = correlation_type_id;
+            this.crAccountTypeID = acctTypeID;
+        }
+        
+
+        /**
+         * @return the acctType
+         */
+        public Account.Type getAcctType() {
+            return acctType;
+        }
+
+        public int getCorrelationTypeId() {
+            return this.correlationTypeID;
+        }
+
+        public int getCRAccountTypeId() {
+            return this.crAccountTypeID;
+        }
+    }
+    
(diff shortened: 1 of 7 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
```diff
@@ -0,0 +1,136 @@
...
+        
+
+        /**
+         * @return the acctType
+         */
+        public Account.Type getAcctType() {
+            return acctType;
+        }
+
+        public int getCorrelationTypeId() {
```
dannysmyda: If this is used to just follow up with a query to get the Type instance with this id, can't the CentralRepoAccountType just be initialized with that instance?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
```diff
@@ -0,0 +1,136 @@
...
+     * that it maps to.
+     */
+    public static final class CRAccountType {
+
+        // id is the primary key in the account_types table
+        private final int crAccountTypeID;
+        private final Account.Type acctType;
+        private final int correlationTypeID;
+
+        CRAccountType(int acctTypeID, Account.Type acctType, int correlation_type_id) {
```
dannysmyda: Why mix of camel and underscore casing?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
```diff
@@ -0,0 +1,136 @@
...
+import org.sleuthkit.datamodel.Account;
+
+
+/**
+ * This class abstracts an Account as stored in the CR database.
+ */
+public final class CRAccount {
+    
+	// primary key in the Accounts table in CR database
+	private final long account_id;
```
dannysmyda: Why is this cased using underscores?

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
```diff
@@ -0,0 +1,136 @@
...
+        public int getCorrelationTypeId() {
+            return this.correlationTypeID;
+        }
+
+        public int getCRAccountTypeId() {
+            return this.crAccountTypeID;
+        }
+    }
+    
+    public CRAccount(long account_id, CRAccountType accountType, String typeSpecificId) {
```
dannysmyda: Same here

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
```diff
@@ -210,6 +228,24 @@ public Long getFileObjectId() {
...
+     */
+    public Long getAccountId() {
+        return accountId;
+    }
+
+    /**
+     * Set the accountId of the account associated with this correlation
+     * attribute.
+     */
+    public void setAccountId(Long accountId) {
```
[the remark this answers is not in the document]
raman-bt: Is the comment about why the setter is public? Or why there is a setter at all?

Many of the Correlation attributes do not have any account.  And depending upon the usage of the CorrelationAttributeInstance (for e.g. in the Other instance viewer in main Autopsy window) the backing account is not needed at all.   I'm keeping the Id instead of the account itself so that it can be read where needed and is not unnecessarily preloaded all the time.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
raman-bt: Yes, currently the only call to getCorrelationTypeId() is followed by getting the type instance.  

But I think as we start using t and build out the UI, the CentralRepoAccountType may be instantiated without needing  the corresponding Correlationtype all the time.  That's why I'm keeping the is so that the CorrelationType can be read in as needed.

dannysmyda:
@raman-bt There is nothing to test with this right now, correct?

raman-bt:
> @raman-bt There is nothing to test with this right now, correct?

You can test with ingesting any of the images with email.  And also our Android image.

Be sure to delete your CentralRepo first so it gets recreated with new tables (we don't have upgrade yet)

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeInstance.java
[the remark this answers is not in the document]
dannysmyda: I'm cool with this. Just to clarify my comment was questioning the setter at all and then my next proposal was making the signature:

public void setAccount(CentralRepoAccount account)

I should've used signature and not abstraction. But anyway, it's a slightly stricter contract. We know this attributeinstance can only have an accountId that is stored in the DB.

In its current state, Long is too broad. The range of negatives it supports isn't logically valid for a CentralRepoAccount, but of course it'll compile. That's all.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CRAccount.java
dannysmyda: Ok.

PR #5674, opened 2020-03-04
6100: Suppress creation of Persona tables.

raman-bt:


raman-bt:
@dannysmyda: Found a bug in this.  Please hold off on test/review. I'll update it shortly.

raman-bt:
> @dannysmyda: Found a bug in this. Please hold off on test/review. I'll update it shortly.

Updated the PR. Good to go now.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/RdbmsCentralRepoFactory.java
```diff
@@ -779,6 +766,43 @@ static String getCreatePersonaAccountsTableStatement(CentralRepoPlatforms select
...
 
     
+     /**
+      * Inserts the default content in accounts related tables.
+      * 
+      * @param conn Database connection to use.
+      * 
+      * @return True if success, false otherwise.
+      */
+    private boolean insertDefaultAccountsTablesContent(Connection conn) {
```
dannysmyda: Can this method just return a SQLException and use a try-with-resources block? Recalling what has been said about exception swallowing with try-catch-finally blocks, I believe try-with-resources handles that case for you automatically. Here are the notes from the JLS https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3.1. Subtle, but the use of addSuppressed() is interesting. Read about it here: https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#addSuppressed(java.lang.Throwable). It looks like it was designed specifically to fill this gap.

dannysmyda: I believe suppressed messages make their way into the logged trace, but I could be wrong.

dannysmyda: If this method is too much trouble to change due to intricate design using the return flag, then its ok.

raman-bt: @dannysmyda: do you mean "just throw a SQLException"? 

I can certainly use try-with-resource in the method to manage the resources, but at this time Id prefer to not change the return value type.  I think the way this and other similar methods are used, is currently expecting a boolean return value. 

And this is going to get used in the upgrade path as well, and may further have some conflicts with changes Greg is making.

dannysmyda: Ok no worries. At first, I wasn't sure if the extended try-with-resources block you just wrote was equivalent to your previous code, but it is according to https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3.2. Suppression will happen, and then the SQLException we actually care about will be caught in the block you define :).

PR #5683, opened 2020-03-05
6020 viber parser produces an illegal argument exception from comm helper

markmckinnon:


InternalPythonModules/android/viber.py
```diff
@@ -117,13 +117,28 @@ def parse_contacts(self, contacts_db, helper):
...
             contacts_parser = ViberContactsParser(contacts_db)
             while contacts_parser.next():
-                helper.addContact( 
-                    contacts_parser.get_contact_name(), 
-                    contacts_parser.get_phone(),
-                    contacts_parser.get_home_phone(),
-                    contacts_parser.get_mobile_phone(),
-                    contacts_parser.get_email()
-                )
+                if contacts_parser.get_phone() is not None:
```
dannysmyda: If this value is a String, I think we should check for blankness too. Blankness can be described as the following conditions:

is_blank(None) //True
is_blank("") //True
is_blank("   ") //True

InternalPythonModules/android/viber.py
```diff
@@ -117,13 +117,28 @@ def parse_contacts(self, contacts_db, helper):
...
+                    helper.addContact( 
+                        contacts_parser.get_contact_name(), 
+                        contacts_parser.get_phone(),
+                        contacts_parser.get_home_phone(),
+                        contacts_parser.get_mobile_phone(),
+                        contacts_parser.get_email()
+                    )
+                else:
```
dannysmyda: I think this should check for blankness as well. 

`elif not is_blank(contacts_parser.get_contact_name())`

InternalPythonModules/android/viber.py
markmckinnon: using isspace() to check for "", " ", and special characters as well as None

InternalPythonModules/android/viber.py
markmckinnon: using isspace() to check for "", " ", and special characters as well as None

InternalPythonModules/android/viber.py
```diff
@@ -117,13 +117,29 @@ def parse_contacts(self, contacts_db, helper):
...
+                    current_case = Case.getCurrentCase().getSleuthkitCase()
+                    attributes = ArrayList()
+                    artifact = contacts_db.getDBFile().newArtifact(BlackboardArtifact.ARTIFACT_TYPE.TSK_CONTACT)
+                    attributes.add(BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_NAME.getTypeID(), self._PARSER_NAME, contacts_parser.get_contact_name()))
+                    artifact.addAttributes(attributes)
+                    
+                    try:
+                    # Post the artifact to blackboard
+                       current_case.getBlackboard().postArtifact(artifact, self._PARSER_NAME)
+                    except Blackboard.BlackboardException as e:
```
dannysmyda: I think we should use the catch block around the while loop or, in other words, just omit the block here.

dannysmyda: I think we should use the catch block around the while loop or, in other words, just omit the block here.

markmckinnon: Removed inner try block.

PR #5686, opened 2020-03-06
6107 prompt user to use CR only in release mode

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/Installer.java
```diff
@@ -59,21 +60,38 @@ private Installer() {
...
         ieListener.installListeners();
+        if (Version.getBuildType() == Version.Type.RELEASE) {
+            centralRepoCheckAndSetup();
+        }
+        // now run regular module startup code
+        try {
+            CentralRepoDbManager.upgradeDatabase();
+        } catch (CentralRepoException ex) {
+            LOGGER.log(Level.SEVERE, "There was an error while upgrading the central repository database", ex);
+            if (RuntimeProperties.runningWithGUI()) {
```
rcordovano: I  notice that this call to reportUpgradeError is conditional on RuntimeProperties.runningWithGUI, which it should be. However, the call to reportUpgradeError on line 143 is not conditional in this way. Maybe it would be better to put a  RuntimeProperties.runningWithGUI conditional in reportUpgradeError.

rcordovano: I'm going to accept this PR and put in the requested change (fix) in a separate PR for the sake of this weekend's automated testing.

PR #5690, opened 2020-03-09
6015: Upgrade existing central repositories to reflect accounts schema

raman-bt:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbUpgrader13To14.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public CaseDbSchemaVersionNumber upgradeSchema(CaseDbSchemaVersionNumber dbSchemaVersion, Connection connection) throws CentralRepoException, SQLException {
+
+        if (dbSchemaVersion.compareTo(new CaseDbSchemaVersionNumber(1, 4)) < 0) {
+
+            try (Statement statement = connection.createStatement();) {
+
+                CentralRepoPlatforms selectedPlatform = CentralRepoPlatforms.getSelectedPlatform();
+
+                // Create account_types and accounts tables which are referred by X_instances tables
```
rcordovano: I think it would be more accurate to say "...referenced by the X_acct_instances tables."

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/RdbmsCentralRepo.java
```diff
@@ -3808,6 +3808,10 @@ public void upgradeSchema() throws CentralRepoException, SQLException, Incompati
                         throw new CentralRepoException("Currently selected database platform \"" + selectedPlatform.name() + "\" can not be upgraded.", Bundle.AbstractSqlEamDb_cannotUpgrage_message(selectedPlatform.name()));
                 }
             }
+
+            // Upgrade to 1.4
+            CaseDbSchemaVersionNumber newVer = new CentralRepoDbUpgrader13To14().upgradeSchema(dbSchemaVersion, conn);
```
rcordovano: As noted in previous comments, newVer is not used.

PR #5691, opened 2020-03-09
5052 utility class for AbstractNode background tasks

rcordovano:


Core/src/org/sleuthkit/autopsy/datamodel/utils/BackgroundTaskRunner.java
```diff
@@ -0,0 +1,163 @@
...
+     * Implementations of this interface do a task in a background thread
+     * supplied by this utility.
+     */
+    @FunctionalInterface
+    public interface NodeTask {
+
+        /**
+         * Performs a task in a thread supplied by this utility.
+         *
+         * @param future The future of the actual background task executing this
```
dannysmyda: Future.cancel() will set the interrupt flag on the underlying thread. What advantages are there of passing the future instance through instead of checking Thread.currentThread().isInterrupted()?

Core/src/org/sleuthkit/autopsy/datamodel/utils/BackgroundTaskRunner.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        @Override
+        public void run() {
+            AbstractNode node = weakNodeRef.get();
+            if (node == null) {
+                return;
+            }
+
+            if (future.isCancelled()) {
```
dannysmyda: For instance, this could be a check to Thread.currentThread().isInterrupted(). Then the problem of passing the Future before execution is gone.

Core/src/org/sleuthkit/autopsy/datamodel/utils/BackgroundTaskRunner.java
dannysmyda: This impl invites deadlock. Future.get() will block one of the Threads indefinitely.

Core/src/org/sleuthkit/autopsy/datamodel/utils/BackgroundTaskRunner.java
rcordovano: I have used that technique before. I agree that it is a good simplification. It just means that the task client doing the cancellation needs to call Future.cancel(true).

Core/src/org/sleuthkit/autopsy/datamodel/utils/BackgroundTaskRunner.java
rcordovano: See reply below.

Core/src/org/sleuthkit/autopsy/datamodel/utils/AbstractNodePropertySheetTask.java
```diff
@@ -0,0 +1,140 @@
...
+ * method of the task Future.
+ *
+ * Only weak references to the AbstractNode and its PropertyChangeListener are
+ * held prior to task execution so that a queued task does not interfere with
+ * garbage collection if the node has been destroyed by the NetBeans framework.
+ *
+ * A thread pool with descriptively named threads (node-background-task-N) is
+ * provided for executing instances of the tasks.
+ */
+public abstract class AbstractNodePropertySheetTask implements Runnable {
```
dannysmyda: The signature should likely be `PropertySheetTask<T extends AbstractNode>`. See the comment on computePropertyValue for more details.

Core/src/org/sleuthkit/autopsy/datamodel/utils/AbstractNodePropertySheetTask.java
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * IMPORTANT: Implementations of this method should check for cancellation
+     * by calling Thread.currentThread().isInterrupted() at approoraite
+     * intervals.
+     *
+     * @param node The AbstractNode.
+     *
+     * @return The result of the computation as a PropertyChangeEvent.
+     */
+    protected abstract PropertyChangeEvent computePropertyValue(AbstractNode node) throws Exception;
```
dannysmyda: Lets assume the background task logic we are submitting is for a LocalFileNode. Its routine depends entirely on the result of LocalFileNode.hasContentChildren(). How will I be able to utilize this class for my use case?

Core/src/org/sleuthkit/autopsy/datamodel/utils/AbstractNodePropertySheetTask.java
rcordovano: Kk.

Core/src/org/sleuthkit/autopsy/datamodel/utils/AbstractNodePropertySheetTask.java
rcordovano: Using generic, as suggested.

PR #5696, opened 2020-03-10
6082: Create Email & Phone correlation attribute instances only for a…

raman-bt:
…ctual accounts created in CaseDB

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
@@ -84,22 +84,11 @@ public class CorrelationAttributeUtil {
             BlackboardArtifact sourceArtifact = getCorrAttrSourceArtifact(artifact);
             if (sourceArtifact != null) {
                 int artifactTypeID = sourceArtifact.getArtifactTypeID();
-                if (artifactTypeID == ARTIFACT_TYPE.TSK_KEYWORD_HIT.getTypeID()) {
-                    BlackboardAttribute setNameAttr = sourceArtifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_SET_NAME));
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -169,58 +158,6 @@ private static BlackboardArtifact getCorrAttrSourceArtifact(BlackboardArtifact a
...
-
-            /*
-             * Validate the phone number. Three to five digit phone numbers may
-             * be valid, but they are too short to use as correlation
-             * attributes.
-             */
-            if (value.length() > 5) {
-                corrAttr = makeCorrAttr(artifact, CentralRepository.getInstance().getCorrelationTypeById(CorrelationAttributeInstance.PHONE_TYPE_ID), value);
-            }
-        }
```
rcordovano: It seems like this normalization should be retained in a method to be called by makeCorrAttrFromAcctArtifact when it process a TSK_ACCOUNT artifact for a phone account.

raman-bt: The CommunicationsManager is already doing the same normalization on phone numbers before creating accounts. Do we really want to repeat that?

W.r.t the length check, my thought was that since we are now only creating these attributes for actual phone numbers which have been retrieved from call logs or messages, and not from Keyword searches, the odds of getting short false positives are much less.  On the other hand, we may be dropping on the floor real short codes such as 411/611/911.

I can certainly put it all back in if you prefer that.

rcordovano: I f we have it covered elsewhere, no worries.

dannysmyda: Just to chime in here, there is no length check in the CommManager. Also, keyword searches were calling makeCorrAttrFromArtifactAttr(), not this function. I'm not sure if any of this matters. It seems important to know if 911 was called, but maybe everyone in the CR has called 911 so the original comment stands. In any case, the code is still picking up the phone numbers that appear in CONTACT, MESSAGE, and CALLOG artifacts, so I'm going to approve.

PR #5699, opened 2020-03-11
6111: CVT displays incorrect To phone number for Outgoing calls.

raman-bt:


Core/src/org/sleuthkit/autopsy/communications/relationships/CallLogNode.java
```diff
@@ -107,6 +100,59 @@ long getCallDuration(BlackboardArtifact artifact) throws TskCoreException {
...
+    private String getPhoneNumber(BlackboardArtifact artifact) {
+        String direction = getAttributeDisplayString(artifact, TSK_DIRECTION);
+
+        String phoneNumberToReturn;
+        String fromPhoneNumber = getAttributeDisplayString(artifact, TSK_PHONE_NUMBER_FROM);
+        String toPhoneNumber = getAttributeDisplayString(artifact, TSK_PHONE_NUMBER_TO);
+        String phoneNumber = getAttributeDisplayString(artifact, TSK_PHONE_NUMBER);
+        switch (direction.toLowerCase()) {
+            case "incoming": // NON-NLS 
+                phoneNumberToReturn = getFirstNonBlank(fromPhoneNumber, phoneNumber, toPhoneNumber);
```
dannysmyda: Nice! This is known as coalescing.

Core/src/org/sleuthkit/autopsy/communications/relationships/CallLogNode.java
```diff
@@ -107,6 +100,59 @@ long getCallDuration(BlackboardArtifact artifact) throws TskCoreException {
...
+     * Returns a blank string if all the input strings are blank.
+     * 
+     * @param string1 First string to check
+     * @param string2 Second string to check
+     * @param string3 Third string to check
+     * 
+     * @retunr first non blank string if there is one, blank string otherwise.
+     * 
+     */
+    private String getFirstNonBlank(String string1, String string2, String string3 ) {
```
dannysmyda: Coalescing blankness is a very common operation. Since you are using commons lang, they've already implemented this for you. 

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#firstNonBlank-T...-

I suggest replacing the above getFirstNonBlank calls with StringUtils.firstNonBlank(...) and then, before returning, check if phoneNumberToReturn is null. If so, you can sub an empty value like you're doing here.

raman-bt: The firstNonBlank() method in StringUtils is available from version 3.8 of commons-lang3.

We are currently using 3.0 (but Thunderbird uses 3.8.1)

@rcordovano: is it ok to update the commons-lang3 to latest ?

rcordovano: Yes.

PR #5700, opened 2020-03-11
6092 context content viewer should support multiple contexts

markmckinnon:


wschaeferB:
I know it is generally a pain to get working correctly but there does not appear to be a vertical scroll bar when all the height of the contents exceed the height of the content viewer.

PR #5707, opened 2020-03-13
6104 persona accounts test

dannysmyda:


Core/test/qa-functional/src/org/sleuthkit/autopsy/commonpropertiessearch/InterCaseTestUtils.java
```diff
@@ -204,7 +202,6 @@
         //kitchen sink
         ArrayList<IngestModuleTemplate> kitchenSink = new ArrayList<>();
         kitchenSink.add(exifTemplate);
-        kitchenSink.add(iOsTemplate);
```
raman-bt: *Unrelated* to your changes, but since I pulled down most recent develop (to ignore device accounts) I had to change line 223 in this file:

I had to change the call to getCorrelationTypes() to getDefinedCorrelationTypes(). 

You may want to make that change in your PR before merging it.

PR #5712, opened 2020-03-16
6081 use helper for artifact creation

dannysmyda:


Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYUtils.java
```diff
@@ -0,0 +1,151 @@
...
+
+    // Pattern is in reverse due to a Java 8 bug, see calculateSecondsSinceEpoch()
+    // function for more details.
+    private static final DateTimeFormatter DATE_TIME_PARSER
+            = DateTimeFormatter.ofPattern("[(XXX) ][O ][(O) ]a h:m:s M/d/y");
+
+    private static final String DEVICE_LOCALE = "(device)";
+    private static final String NETWORK_LOCALE = "(network)";
+
+    public static boolean isPhoneValid(String phoneNumber) {
```
raman-bt: None of these Util methods are being used outside this package, at least not at the moment. 
Leave them with package access instead of public?

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYContactsFileParser.java
```diff
@@ -19,49 +19,30 @@
...
-        put("index", null);
-        put("account name", null);
-        
-    }};
-    
     @Override
     boolean canProcess(XRYKeyValuePair pair) {
-        String normalizedKey = pair.getKey().toLowerCase();
-        return XRY_KEYS.containsKey(normalizedKey);
+        return XryKey.contains(pair.getKey());
```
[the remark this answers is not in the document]
dannysmyda: Yes, the key is lower-cased in XryKey.contains() method.

Core/src/org/sleuthkit/autopsy/datasourceprocessors/xry/XRYUtils.java
dannysmyda: Sounds good.

dannysmyda:
Yes, because I changed the method signature, it seems that I was attributed to writing the function.

PR #5717, opened 2020-03-17
6151: Do not add account_id column to non-account correlation attribu…

raman-bt:
…te instance tables

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbUpgrader13To14.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbUpgrader13To14.java
@@ -51,7 +51,7 @@ public class CentralRepoDbUpgrader13To14 implements CentralRepoDbUpgrader {
                     if (type.getId() >= CorrelationAttributeInstance.ADDITIONAL_TYPES_BASE_ID) {
 
                         // these are new Correlation types - new tables need to be created
-                        statement.execute(String.format(RdbmsCentralRepoFactory.getCreateArtifactInstancesTableTemplate(selectedPlatform), instance_type_dbname, instance_type_dbname));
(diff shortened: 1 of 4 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoDbUpgrader13To14.java
```diff
@@ -61,18 +61,19 @@ public void upgradeSchema(CaseDbSchemaVersionNumber dbSchemaVersion, Connection
...
-                        statement.execute(sqlStr);
-
-                        // SQLite does NOT allow adding a constraint with Alter Table statement.
-                        // The alternative would be to create new tables, copy all data over, and delete old tables - potentially a time consuming process. 
-                        // We decided to not add this constraint for SQLite, since there likely aren't many users using SQLite based Central Repo.
-                        if (selectedPlatform == CentralRepoPlatforms.POSTGRESQL) {
-                            sqlStr = String.format(getAlterArtifactInstancesAddAccountIdConstraintTemplate(), instance_type_dbname);
+                    } else {    // existing attributes 
+                        // Alter the existing _instance tables for Phone and Email attributes to add account_id column 
+                        if (type.getId() == CorrelationAttributeInstance.EMAIL_TYPE_ID || type.getId() == CorrelationAttributeInstance.PHONE_TYPE_ID) {
```
wschaeferB: You could flatten this logic a bit by combining the else and the first conditional statement since everything in the else block also requires the first conditional to be true to happen. 

This would likely make codacy not complain about the deeply nested inner conditional.

raman-bt: Uhh, I'd done just that, yesterday evening.  But apparently I did not push that commit up, as I though I had.

PR #5727, opened 2020-03-20
6077 display image for summary

wschaeferB:


Core/src/org/sleuthkit/autopsy/filequery/FileSearch.java
```diff
@@ -50,6 +50,7 @@
 import org.netbeans.api.progress.ProgressHandle;
 import org.opencv.core.Mat;
 import org.opencv.highgui.VideoCapture;
+import org.openide.util.Exceptions;
```
dannysmyda: Is this an unused import? It's interesting codacy didn't flag this.

PR #5738, opened 2020-03-25
6172: Validation of correlation attribute value not happening correctly

raman-bt:


InternalPythonModules/android/general.py
```diff
@@ -37,3 +40,15 @@ def appendAttachmentList(msgBody, attachmentsList):
         body = body + "\n".join(list(filter(None, attachmentsList)))
 
     return body
+
+"""
+Checks if the given string might be a phone number.
+"""
+def isValidPhoneNumer(data):
```
dannysmyda: Why can't you just call out to the CommunicationUtils class?

PR #5741, opened 2020-03-25
6163 New panel for type filtering in the geolocation tool

ethteck:


kellykelly3:
Please address the codacy issues by adding a private constructor and changing showLabel to a local var.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -415,19 +490,20 @@ int getMostRecentNumDays() {
          * 
          * @throws TskCoreException 
          */
-        private boolean isGPSDataSource(SleuthkitCase sleuthkitCase, DataSource dataSource) throws TskCoreException {
+        private Map<ARTIFACT_TYPE, Long> getGPSDataSources(SleuthkitCase sleuthkitCase, DataSource dataSource) throws TskCoreException {
```
kellykelly3: Update method header.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -415,19 +490,20 @@ int getMostRecentNumDays() {
...
+            HashMap<ARTIFACT_TYPE, Long> ret = new HashMap<>();
             for (BlackboardArtifact.ARTIFACT_TYPE type : GPS_ARTIFACT_TYPES) {
-                if (sleuthkitCase.getBlackboardArtifactsTypeCount(type.getTypeID(), dataSource.getId()) > 0) {
-                    return true;
+                long count = sleuthkitCase.getBlackboardArtifactsTypeCount(type.getTypeID(), dataSource.getId());
+                if (count > 0) {
+                    ret.put(type, count);
```
kellykelly3: Why are you keeping count?  You are not adding the type unless count is greater than one.  I just don't think you need to keep count.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -440,8 +516,14 @@ public void done() {
...
 
             if (sources != null) {
-                for (Pair<String, DataSource> source : sources) {
-                    checkboxPanel.addElement(source.getKey(), source.getValue());
+                for (Pair<String, DataSource> source : sources.dataSources) {
+                    dsCheckboxPanel.addElement(source.getKey(), null, source.getValue());
+                }
+                for (Map.Entry<ARTIFACT_TYPE, Long> entry : sources.artifactTypes.entrySet()) {
+                    String dispName = entry.getKey().getDisplayName() + " (" + entry.getValue() + ")";
+                    String iconPath = "/" + IconsUtil.getIconFilePath(entry.getKey().getTypeID());
```
kellykelly3: Different OSs use different path separators, ie \ vs /.  use Paths.get instead

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
```diff
@@ -459,7 +460,7 @@ private WaypointBuilder() {
      *
      * @throws GeoLocationDataException
      */
-    static public void getAllWaypoints(SleuthkitCase skCase, List<DataSource> dataSources, boolean showAll, int cntDaysFromRecent, boolean noTimeStamp, WaypointFilterQueryCallBack queryCallBack) throws GeoLocationDataException {
+    static public void getAllWaypoints(SleuthkitCase skCase, List<DataSource> dataSources, List<ARTIFACT_TYPE> artifactTypes, boolean showAll, int cntDaysFromRecent, boolean noTimeStamp, WaypointFilterQueryCallBack queryCallBack) throws GeoLocationDataException {
```
kellykelly3: Make sure the method header is updated.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
```diff
@@ -480,9 +481,10 @@ public void process(ResultSet rs) {
                             int artifact_type_id = rs.getInt("artifact_type_id"); //NON-NLS
                             long artifact_id = rs.getLong("artifact_id"); //NON-NLS
 
-                            BlackboardArtifact.ARTIFACT_TYPE type = BlackboardArtifact.ARTIFACT_TYPE.fromID(artifact_type_id);
-
-                            waypoints.addAll(getWaypointForArtifact(skCase.getBlackboardArtifact(artifact_id), type));
+                            ARTIFACT_TYPE type = ARTIFACT_TYPE.fromID(artifact_type_id);
```
kellykelly3: I am a little torn about whether to do this or include the types in the query.  Was there a reason that you didn't look at extending the query?  It might be too messy.  That figuring out the query was a pain.

Core/src/org/sleuthkit/autopsy/geolocation/CheckBoxJList.java
```diff
@@ -91,18 +109,33 @@ public void mousePressed(MouseEvent e) {
...
-    class CellRenderer extends JCheckBox implements ListCellRenderer<CheckBoxJList.CheckboxListItem> {
+    class CellRenderer extends JPanel implements ListCellRenderer<CheckBoxJList.CheckboxListItem> {
 
         private static final long serialVersionUID = 1L;
 
+        private final JCheckBox checkbox = new JCheckBox();
+        private final JLabel label = new JLabel();
+
+        public CellRenderer() {
+            setLayout(new BorderLayout(2, 0));
```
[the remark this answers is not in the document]
rcordovano: @kellykelly3 and @ethteck, if the issue is a leaking this pointer, you need a method to create instances of the class via two stage construction.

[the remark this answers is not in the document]
kellykelly3: The issue complaining about  is calling functions that can be overridden in the constructor.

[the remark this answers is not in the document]
rcordovano: Similar issue, same solution.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
ethteck: This just seemed like the right place to do type filtering. I don't really have a specific reason other than this seeming to be a good way to do it to me.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
ethteck: We display the count for each artifact type in the UI, so that's why I'm keeping count here.

ethteck:
@kellykelly3 I believe most if not all of your comments have been addressed. Let me know what you think.

ethteck:
> Please address the codacy issues by adding a private constructor and changing showLabel to a local var.

This is generated code and other UI constructs have the same issue unfortunately.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
kellykelly3: So you didn't look into adjusting the query at all?

ethteck: I did, but I just settled on this. It sounds like you prefer I do the logic in the query?

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -440,8 +518,14 @@ public void done() {
...
             if (sources != null) {
-                for (Pair<String, DataSource> source : sources) {
-                    checkboxPanel.addElement(source.getKey(), source.getValue());
+                for (Pair<String, DataSource> source : sources.dataSources) {
+                    dsCheckboxPanel.addElement(source.getKey(), null, source.getValue());
+                }
+                for (Map.Entry<ARTIFACT_TYPE, Long> entry : sources.artifactTypes.entrySet()) {
+                    String dispName = entry.getKey().getDisplayName() + " (" + entry.getValue() + ")";
+                    String iconPath = File.separator + IconsUtil.getIconFilePath(entry.getKey().getTypeID());
+                    Icon icon = new ImageIcon(getClass().getResource(iconPath));
```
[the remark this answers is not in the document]
ethteck: Ah, I see. I'll look into this, thanks. Yeah the slash should be "/" and it should probably just happen in getIconFilePath. I excluded it from that because the original caller seemed to break when I added it there. But I'll dig into it more.

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
```diff
@@ -49,24 +55,26 @@
...
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_BOOKMARK,
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_LAST_KNOWN_LOCATION,
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_ROUTE,
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_SEARCH,
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACK,
-        BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACKPOINT
+    private static final ARTIFACT_TYPE[] GPS_ARTIFACT_TYPES = {
+        ARTIFACT_TYPE.TSK_GPS_BOOKMARK,
+        ARTIFACT_TYPE.TSK_GPS_LAST_KNOWN_LOCATION,
+        ARTIFACT_TYPE.TSK_GPS_ROUTE,
```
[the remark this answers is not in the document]
ethteck: I'll add the other type for now. I guess we'll want to rethink this approach going forward

Core/src/org/sleuthkit/autopsy/geolocation/GeoFilterPanel.java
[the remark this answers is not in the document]
ethteck: I believe I have fixed this @kellykelly3. Let me know if you continue to see the NPE.

kellykelly3:
Please take care of the codacy issues.  The cyclomatic complexity issue could probably be fixed by just returning the string at the end instead of in each if.

ethteck:
@kellykelly3 that unfortunately didn't fix it. Also, this function already existed in the code base and is only coming up as a new error because I moved it. I think a switch statement would be better, but it doesn't work because the ArtifactType.getId() function is not static.

PR #5758, opened 2020-03-30
6165 Always populate name, orig name cols

rcordovano:


dannysmyda:
Network delay for translation makes it practically impossible to navigate the data sources tree. Not sure what the issue is.

PR #5771, opened 2020-04-02
6186 horizontal resizing

gdicristofaro:


Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
+/*
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2019 Basis Technology Corp.
```
wschaeferB, line 4: Year is incorrect.

Also I am not a fan of having multiple classes with identical names is there anyway we can rename one of our two WrapLayout classes?

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+ * Inspired by WrapLayout
+ * https://tips4java.wordpress.com/2008/11/06/wrap-layout/ and FlowLayout
+ * https://raw.githubusercontent.com/mynawang/Java8-Source-Code/master/src/main/jdk8/java/awt/FlowLayout.java.
+ */
+public class WrapLayout implements LayoutManager, java.io.Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * The <code>WrapLayout</code> manager allows a seperation of components
```
wschaeferB, line 53: separation

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+     * and between the components and the borders of the <code>Container</code>.
+     *
+     * @serial
+     * @see #getHgap()
+     * @see #setHgap(int)
+     */
+    private int hgap = 0;
+
+    /**
+     * The <code>WrapLayout</code> manager allows a seperation of components
```
wschaeferB, line 64: separation

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+    /**
+     * Adds the specified component to the layout. Not used by this class.
+     * NOTE: This is not used for this layout
+     *
+     * @param name The name of the component.
+     * @param comp The component to be added.
+     */
+    @Override
+    public void addLayoutComponent(String name, Component comp) {
+    }
```
wschaeferB, line 206: This doesn't feel like something we want to add to the public API a method where it does nothing and just has a note saying it is not used for this layout in the comment.

Also codacy doesn't like the empty body being undocumented.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+    /**
+     * Removes the specified component from the layout. Not used by this class.
+     * NOTE: This is not used for this layout
+     *
+     * @param comp The component to remove.
+     */
+    @Override
+    public void removeLayoutComponent(Component comp) {
```
wschaeferB, line 215: This doesn't feel like something we want to add to the public API a method where it does nothing and just has a note saying it is not used for this layout in the comment.

Also codacy doesn't like the empty body being undocumented.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+    /**
+     * This class provides metrics on the parent container dimensions.
+     */
+    private static class ParentDimensions {
+
+        /**
+         * The full width that the component can consume.
+         */
+        final int outerWidth;
```
wschaeferB, line 364: Should be private with package level getters and setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+        /**
+         * The full width that subcomponent rows can consume. This is the
+         * outerWidth accounting for left and right insets.
+         */
+        final int innerWidth;
```
wschaeferB, line 370: Should be private with package level getters and setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+        /**
+         * The insets of the parent container.
+         */
+        final Insets insets;
```
wschaeferB, line 375: Should be private with package level getters and no setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+     * @param target The target container.
+     *
+     * @return The dimensions for laying out components.
+     */
+    private ParentDimensions getTargetDimensions(Container target) {
+        //  Each row must fit with the width allocated to the containter.
+        //  When the container width = 0, the preferred width of the container
+        //  has not yet been calculated so lets ask for the maximum.
+
+        int targetWidth = target.getSize().width;
```
wschaeferB, line 397: Assigned value never used.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+            container = container.getParent();
+        }
+
+        targetWidth = container.getSize().width;
+
+        if (targetWidth == 0) {
+            targetWidth = Integer.MAX_VALUE;
+        }
+
+        int hgap = getHgap();
```
wschaeferB, line 410: local variable hides field

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+    /**
+     * Items in the collection will be aligned opposite to the rest. For
+     * instance, if items should be displayed left to right based on locale,
+     * these components will be on the right.
+     *
+     * @return The components to display with opposite alignment.
+     */
+    public Collection<Component> getOppositeAlignedItems() {
+        return oppositeAlignedItems;
```
wschaeferB, line 127: return of collection field.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+     * Lays out the container. This method lets each
+     * <i>visible</i> component take its preferred size by reshaping the
+     * components in the target container and creating new rows.
+     *
+     * @param target The specified component being laid out.
+     */
+    @Override
+    public void layoutContainer(Container target) {
+        synchronized (target.getTreeLock()) {
+            synchronized (this.oppositeAlignedItems) {
```
wschaeferB, line 289: getting locks inside locks seems like a recipe for problems to me but I will let @rcordovano  weigh in on this one.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+     * container.
+     *
+     * @param target    Target to get layout size for.
+     * @param preferred Should preferred size be calculate.d
+     *
+     * @return The dimension to layout the target container.
+     */
+    private Dimension layoutSize(Container target, boolean preferred) {
+        synchronized (target.getTreeLock()) {
+            synchronized (this.oppositeAlignedItems) {
```
wschaeferB, line 430: synchronizing inside synchronization

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+         */
+        final int height;
+
+        /**
+         * The minimum width of the row which is the sum of the preferred widths
+         * of the subcomponents.
+         */
+        final int width;
+
+        public WrapLayoutRow(List<Component> components, List<Component> oppositeAligned, int height, int width) {
```
wschaeferB, line 490: public constructor in private class should be package level constructor.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+    /**
+     * A row of components in the WrapLayout.
+     */
+    private class WrapLayoutRow {
+
+        /**
+         * The normally aligned components in the order that they will be laid
+         * out.
+         */
+        final List<Component> components;
```
wschaeferB, line 470: Should be private with package level getters and no setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+        /**
+         * The opposite aligned components in the order that they will be laid
+         * out.
+         */
+        final List<Component> oppositeAligned;
```
wschaeferB, line 476: Should be private with package level getters and no setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+        /**
+         * The minimum height of the row which is the maximum of the preferred
+         * heights of the components.
+         */
+        final int height;
```
wschaeferB, line 482: Should be private with package level getters and no setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
```diff
@@ -0,0 +1,607 @@
...
+
+        /**
+         * The minimum width of the row which is the sum of the preferred widths
+         * of the subcomponents.
+         */
+        final int width;
```
wschaeferB, line 488: Should be private with package level getters and no setters.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: This is a method of the interface, java.awt.LayoutManager.  I documented that it was not in use in the method body.  Is there a particular way this should be documented?

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: This is a method of the interface, java.awt.LayoutManager.  I documented that it was not in use in the method body.  Is there a particular way this should be documented?

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: To which variable are you referring?  I believe targetWidth is used a couple times in getTargetDimensions.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: Do you have a preference how this is handled (i.e. returning a new list generated from the original or something else)?

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: My aim was to synchronize on both items similar to https://stackoverflow.com/a/4604038 to get access to the tree lock to avoid problems with altering objects in a container during layout: https://docs.oracle.com/javase/7/docs/api/java/awt/Component.html#getTreeLock() as well as prevent access to the collection created for objects that will be oppositely aligned during layout operations.

gdicristofaro:
> A couple codacy complaints, some clean up of WrapLayout and a couple questions about things that will need feedback from Richard or Brian see Jira for my main concern.

I left getAlignOnBaseline for consistency with FlowLayout.  If there is a specific way unused methods should be commented, I would be glad to do so.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
gdicristofaro: I corrected the year and there is a question in Jira for the second concern.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
wschaeferB: Generally we have been using Collections.unmodifiableCollection(...)

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
wschaeferB: Putting a comment in the body generally makes codacy happy about its "Document Empty Body" suggestion. I usually go with something like //Empty - This method intentionally left blank.  But there is no like required phrasing that I know of.

My bigger concern is that these methods are expected to do things when implementing this class and that a user calling them will not quickly realize they are not doing anything when making use of the public API, I almost feel like them throwing an Unsupported exception would be preferable if they are not used by our code.

Core/src/org/sleuthkit/autopsy/guiutils/WrapLayout.java
rcordovano: @gdicristofaro and @wschaeferB, nested locking is really risky and the consequence of a mistake, now or in maintenance, is a deadlock. Given that this code is UI code that will be thread-confined in the EDT, there should be no need for any synchronization at all in this class. Swing components get their thread safety by thread-confinement by design. Please remove all synchronization.

PR #5775, opened 2020-04-03
6236: Other occurrences not working for new types in CVT Summary panel

raman-bt:


APriestman:
The codacy suggestion looks reasonable so please address.

PR #5780, opened 2020-04-06
6228 - Fix CR Regressions

dannysmyda:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -59,28 +62,58 @@ private static String getEmailAddressAttrDisplayName() {
...
     /**
      * Makes zero to many correlation attribute instances from the attributes of
-     * an artifact.
+     * artifacts that have correlatable data. The intention of this method is to
+     * use the results to correlate with, not to save. If you 
+     * want to save, please use makeCorrAttrsToSave. An artifact that can
+     * have correlatable data != An artifact that should be the source of data
+     * in the CR, so results may be too lenient.
      *
-     * IMPORTANT: The correlation attribute instances are NOT added to the
```
dannysmyda, line 82: Whoops, I don't know how this got deleted. I'll add it back in.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -158,6 +192,51 @@ private static String getEmailAddressAttrDisplayName() {
...
+     *
+     * @return The correlation instance artifact or null, if the phone number is
+     *         not a valid correlation attribute.
+     *
+     * @throws TskCoreException     If there is an error querying the case
+     *                              database.
+     * @throws CentralRepoException If there is an error querying the central
+     *                              repository.
+     */
+    private static void makeCorrAttrFromArtifactPhoneAttr(List<CorrelationAttributeInstance> corrAttrInstances, BlackboardArtifact artifact) throws TskCoreException, CentralRepoException {
```
dannysmyda, line 209: Just FYI, this is a revert of previous changes.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -158,6 +192,51 @@ private static String getEmailAddressAttrDisplayName() {
...
+        } else if (null != artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO))) {
+            value = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO)).getValueString();
+        }
+
+        /*
+         * Normalize the phone number.
+         */
+        if (value != null) {
+            try {
+                value = CommunicationsUtils.normalizePhoneNum(value);
```
dannysmyda, line 229: With the exception of this

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -59,28 +62,58 @@ private static String getEmailAddressAttrDisplayName() {
         return Bundle.CorrelationAttributeUtil_emailaddresses_text();
     }
 
+    // Defines which artifact types act as the sources for CR data.
+    // Most notably, does not include KEYWORD HIT, CALLLOGS, MESSAGES, CONTACTS
+    // TSK_INTERESTING_ARTIFACT_HIT (See JIRA-6129 for more details on the
+    // interesting artifact hit).
+    private static final Set<Integer> SOURCE_TYPES_FOR_CR_INSERT = new HashSet<Integer>() {{
```
dannysmyda, line 69: Motivation for defining the types, and not their negation (by that I mean not keyword hit, not calllog, not contacts), is this will be safer to keep it up to date. Basically, if a new type comes along for correlation only, we are good. If a new type comes along for insert, during testing they won't see the expected behavior and update this set. However, with the negation, any new artifact types accepted for correlation will automatically be saved, which is too accepting for accidents.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
```diff
@@ -455,11 +455,8 @@ public void run() {
             List<CorrelationAttributeInstance> eamArtifacts = new ArrayList<>();
 
             for (BlackboardArtifact bbArtifact : bbArtifacts) {
-                // If the incoming artifact is of type TSK_INTERESTING_ARTIFACT_HIT, 
-                // do not resolve to the source artifact, as correlation attributes 
-                // for the source artifact would have already been created, 
-                // when the event for that source artifact was received.
-                List<CorrelationAttributeInstance> convertedArtifacts = CorrelationAttributeUtil.makeCorrAttrsFromArtifact(bbArtifact, false);
+                // makeCorrAttrToSave will filter out artifacts which should not be soruces of CR data.
```
rcordovano, line 458: soruces => sources

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -158,6 +192,51 @@ private static String getEmailAddressAttrDisplayName() {
...
+         * Extract the phone number from the artifact attribute.
+         */
+        String value = null;
+        if (null != artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER))) {
+            value = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER)).getValueString();
+        } else if (null != artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM))) {
+            value = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM)).getValueString();
+        } else if (null != artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO))) {
+            value = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO)).getValueString();
+        }
```
rcordovano, line 222: I realize that the code above is restored code, but it looks like it contains another bug. The contact book entry, call log entry and message artifacts all support multiple phone number attributes. I think we need to loop over all of the attributes in the artifact, rather than stop after "harvesting" one phone number.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
raman-bt: I know this is simply reverting to what 'was'...

But we could have email address attributes in the CONTACT_ARTIFACT, which should produce an CR attribute.  Perhaps we should make a story to come back and revisit this later.

rcordovano: I'd like us to make the fix now.

dannysmyda: Yes, it can have email, email_to, email_cc, email_bcc, phone, phone_to, phone_from, phone_home, phone_office, phone_mobile. I looked at the code and the same logic applies to some other artifacts. I think creating a story would be a good idea.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
dannysmyda: See Raman's comment for new story proposal.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
raman-bt: I think a comment here to indicate that this list should be updated with any new artifacts that may be source for CR attributes (for creation) would be useful.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
rcordovano: @dannysmyda, please create the story, assign it to yourself, etc. Thanks!

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
rcordovano: @dannysmyda, please create the story or stories, assign to self, etc. Thanks!

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
dannysmyda: Good idea, I've added a comment. Thanks.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
dannysmyda: I'll just do it here then.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
dannysmyda: I'll just do it here then, for the phone attr method.

PR #5782, opened 2020-04-07
6161 render waypoints different colors depending on their type

ethteck:


Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
```diff
@@ -64,6 +67,18 @@
...
+    
+    static {
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_BOOKMARK.getTypeID(), Color.BLUE);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_LAST_KNOWN_LOCATION.getTypeID(), Color.RED);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_ROUTE.getTypeID(), Color.YELLOW);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_SEARCH.getTypeID(), Color.GREEN);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACK.getTypeID(), Color.ORANGE);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACKPOINT.getTypeID(), Color.ORANGE);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_METADATA_EXIF.getTypeID(), Color.CYAN);
+    }
```
kellykelly3, line 81: Why put the colors here instead of in Waypoint?  I would have added a getColor to Waypoint and have the subclass set the color if they want it to be something other than the default.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
```diff
@@ -685,17 +689,85 @@ private void zoomOutBtnActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FI
      * Renderer for the map waypoints.
      */
     private class MapWaypointRenderer implements WaypointRenderer<MapWaypoint> {
+        private final Map<Color, BufferedImage> imageCache = new HashMap<>();
```
kellykelly3, line 692: This class is probably big enough for its own class file now.  Leaving it in MapPanel just bloats the file.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.form
```diff
@@ -116,7 +120,9 @@
                 <Property name="borderPainted" type="boolean" value="false"/>
                 <Property name="focusPainted" type="boolean" value="false"/>
                 <Property name="requestFocusEnabled" type="boolean" value="false"/>
-                <Property name="rolloverEnabled" type="boolean" value="false"/>
+                <Property name="rolloverEnabled" type="boolean" editor="org.netbeans.modules.form.RADConnectionPropertyEditor">
```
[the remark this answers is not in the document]
ethteck: I'm not. This is my way of keeping them "false", since my IDE is removing these lines unless I manually specify "false" as user code. I'll just manually restore this file.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
ethteck: I don't think Waypoint, which seems to be a data representation of a waypoint, should be storing information about its color, which is more of a visual feature and has nothing to do with the data inside the waypoint object.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
ethteck: Okay, I'll move it out.

ethteck: MapWaypoint is not public, so it would require making it so to move this class out from being an inner class. I don't really think it's that large. Thoughts @rcordovano ?

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
kellykelly3: @rcordovano and I were talking about this the other day.  I think the Waypoint class (and its subclasses) will need to become a bit more than they are.  Adding a color property to them does not seem out of line as it is a property of the Waypoint.  I would like to see the color go there so that we can avoid this type checking.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
kellykelly3: You would not have to make it public.  By specifying neither public nor private at the class definition the class will be visible to all other classes in the package.  

Please move it to its own class file.  I don't recall at the moment why I made MapPanel public that might be erroneous.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
```diff
@@ -64,6 +67,18 @@
     private static final Logger logger = Logger.getLogger(MapWaypoint.class.getName());
     private final static String HTML_PROP_FORMAT = "<b>%s: </b>%s<br>";
     static private final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX", Locale.US);
+    
+    private static final Map<Integer, Color> artifactTypesToColors = new HashMap();
```
kellykelly3, line 71: new HashMap<>();

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
```diff
@@ -64,6 +67,18 @@
...
+    
+    static {
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_BOOKMARK.getTypeID(), Color.BLUE);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_LAST_KNOWN_LOCATION.getTypeID(), Color.RED);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_ROUTE.getTypeID(), Color.YELLOW);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_SEARCH.getTypeID(), Color.GREEN);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACK.getTypeID(), Color.ORANGE);
+        artifactTypesToColors.put(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACKPOINT.getTypeID(), Color.ORANGE);
```
kellykelly3, line 79: You are getting a warning for the use of the deprecated value, add the annotation to ignore the warning.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
rcordovano: @ethteck and @kellykelly3, some thoughts...

None of this code is public API, i.e., none of the packages we are discussing are visible to code outside of this NetBeans module / JAR. That give us some welcome flexibility in refactoring and evolving the code over time, versus getting it absolutely right the first time. This is a good thing.

The MapWaypoint class does seem to be fairly large, so that is an argument in favor of giving it its own file. However, that is not the only criterion for judging whether a class should be an inner class, at least in my opinion. In fact, I have I have wrestled long and hard with these types of decisions, and have not come to a universal conclusion or reliable rule of thumb. I think as a general guideline, if something has no meaning outside of its use in another class, it is a good candidate for an inner class. From what I am seeing, it looks like MapWaypoint is a view class wrapped around a Waypoint model class with usage outside of a MapPanel (but please correct me if I am wrong, I am guilty of code skimming here), so it does seem to merit its own file. It looks that has already been done, so I think that's fine. We can change it later if we wish.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
rcordovano: @ethteck and @kellykelly3, working from the premise that I tentatively stated in another comment, i.e., that MapWaypoint is a view class wrapped around a Waypoint model class, it makes sense to me to put color properties here in the view class instead of in the data model class. It is my recollection that Waypoint imposes a data model on the weakly typed BlackboardArtifact construct beneath it, which is the justification for the class. I certainly am a fan of separating the model and the view. 

Regarding the future of the Waypoint class, in the cited conversation I was wondering about how the fact that the refactoring Brian requested at the SleuthKit level has given us a new concept of a Waypoint data model class. Specifically, a waypoint as a property of a route. At the time, the meaning of Waypoint in the context of the geolocation packages was no longer loaded into my brain, so I was starting to question whether or not TSK Waypoints could take the place of Autopsy geolocation Waypoints. That notion is not surviving the light of day. Even the cursory glance I have taken at this Autopsy code shows that TSK Waypoints and Autopsy Waypoints are orthogonal, despite the unfortunate similarity in their names. Like I said, a TSK Waypoint is a component of a route and exists to support Java serialization, while an Autopsy Waypoint imposes a welcome layer of abstraction over a family of related artifacts that have no mechanism for expressing that homogeneity in terms of the geolocation data display data model. I think that MapWaypoint and Waypopint are looking secure in their respective roles and are in accord with the Single Responsibility Principle (SRP).  So Kelly, I think what we hammered out together when this feature first appeared is holding up.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
ethteck: The annotation isn't playing well with the static block apparently. I think it's a fair warning that will remind us to remove this deprecated artifact type once it's removed.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
ethteck: @rcordovano I may have misunderstood your answer, but it seems like you're talking about whether MapWaypoint should have its own file. However, what Kelly is suggesting is that MapWaypointRenderer be its own file. Do you have an opinion on this? 

I tried to do so earlier, and for some reason I was getting errors saying MapWaypoint wasn't public despite both being classes being in the same package. Maybe that was just the IDE being buggy. I can try again if you think that's what we should do.

Core/src/org/sleuthkit/autopsy/geolocation/MapWaypoint.java
rcordovano: @ethteck and @kellykelly3, we will likely never remove the deprecated artifact type. At least not until there is a TSK 5, something which I currently have no reason to expect will happen soon. If we can't figure out where to put the suppress deprecation warnings annotation, we will just have to live with the warning. It will be annoying and more or less permanent, but we can only do our best within the constraints we have. I think that we shouldn't let this issue hold up the PR.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
rcordovano: @ethteck and @kellykelly3, I _was_ talking about the wrong class, which explains a lot. I apologize for any confusion caused by my confusion. I have to go with Ethan's argument for the location of MapWaypointRenderer. It is a small, private class which has no use outside of MapPanel. If we want to cut down the size of the MapPanel class, I think we need to look elsewhere.

PR #5789, opened 2020-04-08
6251 is supported NPE

gdicristofaro:


Core/src/org/sleuthkit/autopsy/textextractors/TextFileExtractor.java
```diff
@@ -73,7 +76,26 @@ private Reader getReader(Charset encoding) {
...
+        // get the MIME type
+        String mimeType = file.getMIMEType();
+        
+        // if it is not present, attempt to use the FileTypeDetector to determine
+        if (StringUtils.isEmpty(mimeType)) {
+            FileTypeDetector fileTypeDetector = null;
+            try {
+                fileTypeDetector = new FileTypeDetector();
+            } catch (FileTypeDetector.FileTypeDetectorInitException ex) {
+                logger.log(Level.WARNING, "Unable to create file type detector for determining MIME type.");
```
eugene7646, line 89: @rcordovano I'm wondering if we should log this as severe? I know that typically we try to reserve severe for errors that compromise the case, but I think if we are unable to instantiate FileTypeDetector class, then it definitely needs to be investigated.

rcordovano: I agree, something is very wrong if we can't do file type detection and it will greatly affect the case.

eugene7646: Yep. Everything else looks good.

gdicristofaro: Ok.  This should be resolved now.  Thanks.

Core/src/org/sleuthkit/autopsy/textextractors/TextFileExtractor.java
```diff
@@ -73,7 +76,26 @@ private Reader getReader(Charset encoding) {
...
+            } catch (FileTypeDetector.FileTypeDetectorInitException ex) {
+                logger.log(Level.SEVERE, "Unable to create file type detector for determining MIME type.");
+                return false;
+            }
+            mimeType = fileTypeDetector.getMIMEType(file);
+            
+            // if able to determine mime type, 
+            if (!StringUtils.isEmpty(mimeType))
+                file.setMIMEType(mimeType);
+        }
```
rcordovano, line 97: @eugene7646 and @gdicristofaro...

Do you gentlemen realize that AbstractFile.setMIMEType only stores the result in the object, not in the case database? I suspect that is not what was intended here. I'm no longer sure what our policy is: do we save or throw away such calculations? We should check what we do elsewhere, and if the intent is to save here, call AbstractFile.save here.

eugene7646: > Do you gentlemen realize that AbstractFile.setMIMEType only stores the result in the object, not in the case database?
The answer is "no". Personally I think we should save it in the database but I agree, we should check what we do in other places.

gdicristofaro: From what I can see, typically the result of FileTypeDetector.getMIMEType() is not persisted like FileViewer.isPreferred().  I don't believe I saw an example that the MIME type is set on a purely in-memory object either.  So, I can keep with this precedent if you would like and remove the file.setMIMEType all together or call save().  It appears that the save() method throws an exception on sql error, which may increase the complexity of the isSupported() method.

rcordovano: Let's just remove the call, thanks.

gdicristofaro: Got it.

Core/src/org/sleuthkit/autopsy/textextractors/TextFileExtractor.java
```diff
@@ -73,7 +76,22 @@ private Reader getReader(Charset encoding) {
...
+        // get the MIME type
+        String mimeType = file.getMIMEType();
+        
+        // if it is not present, attempt to use the FileTypeDetector to determine
+        if (StringUtils.isEmpty(mimeType)) {
+            FileTypeDetector fileTypeDetector = null;
+            try {
+                fileTypeDetector = new FileTypeDetector();
+            } catch (FileTypeDetector.FileTypeDetectorInitException ex) {
+                logger.log(Level.SEVERE, "Unable to create file type detector for determining MIME type.");
```
rcordovano, line 89: We should pass the exception to the logger as the third argument so that a stack trace is logged.

Issue #4259, opened 2018-11-05
Java-Problems on Mac OSX Mojave


chef2505:
Hi.
I tried to install autopsy 4.9 on MacOsX Mojave, following the instructions provided in the zip-file. Installation works well, but when starting, I only receive a blank window, without the possibility to open a case or work with autopsy. Trying to activate the autopsy-core-plugin, I receive the following message:

" Activation failed: StandardModule:org.sleuthkit.autopsy.core jarFile: /Users/Henry/Downloads/autopsy-4.9.0/autopsy/modules/org-sleuthkit-autopsy-core.jar: java.lang.UnsupportedClassVersionError: org/sleuthkit/datamodel/SleuthkitJNI has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0"

I have tried to install different java-versions, but without success.
Is there anyone with a solution for me?
Regards

apicciut:
Hey, same issue, let me load the GUI but couldn't begin an acquisition. 
Can anyone look into a patch, and let this thread know?

bcarrier:
My guess is that it has to do with the version of Java that brew uses.

apicciut:
Thanks for the quick response Brian, what would your suggested work around be?

chef2505:
Hi..
The mismatching java-version was also my first thought, hence I tried several java versions. Perhaps I missed the right one.
Thx

apicciut:
Chef2505, let me know if you are able to successfully roll back to a version that will compile the core plugins.

chef2505:
I have started with autopsy 4.8, which I was not able to run, respectively I was not able to activate the core plugins, the same problem like 4.9. So there was no version running on my mac. Sorry

bcarrier:
It looks like the brew package was compiled with JDK 11 (class version 55), but Autopsy uses JDK 8 (class version 52).  

Long-term, it would be good if brew shipped with the TSK Jar file complied with JDK 8. Short-term, you'll need to compile TSK yourself....   Download the TSK .tar.gz file and compile it. (./configure, make, make install, etc.).

I'll follow up with the brew people.

bcarrier:
I'm not entirely sure how to follow up with the brew people....

apicciut:
Thank you for putting so much care into this issue, take your time with the brew situation. For the meantime, could you explain your solution a little more? I’m not a very proficient Unix user. None of the students in my class are either.
[...]

PR #5803, opened 2020-04-12
Update installation instructions for macOS

Lazza:
On macOS one has to install the `liberica-jdk8-full` package in order to have OpenJFX.

The `liberica-jdk8` package **does not work correctly with Autopsy.**

### DIFF
--- a/Running_Linux_OSX.txt
+++ b/Running_Linux_OSX.txt
@@ -27,7 +27,7 @@ The following need to be done at least once. They do not need to be repeated for
 -- OS X:
     1. Install BellSoft Java 8.
         % brew tap bell-sw/liberica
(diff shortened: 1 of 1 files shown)


Lazza:
PS: please also check the caveat mentioned here.

https://github.com/sleuthkit/autopsy/issues/4259#issuecomment-612611286

Lazza:
@rcordovano thanks. I also tried to find which file must be changed to fix the `nbexec` script as outlined in #5804, but I was not able to find it.

Could you maybe help me with that?

markmckinnon:
That file is part of the NetBeans platform which Autopsy uses. This is why you cannot find it in the repo.

PR #5811, opened 2020-04-15
Update modDev.dox

HappyLer:
Add a hint, that you can deploy changes without rebuilding while debugging.


APriestman:
I would change it to "If you want to apply changes during debugging and have not changed any method signatures, you can use the "Apply Code Changes" function in NetBeans."

HappyLer:
@APriestman Yes you are right. Forgot to mention that the signatures should not be changed. Thanks.

PR #5819, opened 2020-04-16
6113 gpx file ingest

gdicristofaro:


InternalPythonModules/GPX_Module/GPX_Parser_Module.py
```diff
@@ -75,158 +80,178 @@ def getModuleDescription(self):
...
 
-    def createDataSourceIngestModule(self, ingestOptions):
+    def createFileIngestModule(self, ingestOptions):
         return GPXParserDataSourceIngestModule()
 
-    
+
 # Data Source-level ingest module. One gets created per data source.
-class GPXParserDataSourceIngestModule(DataSourceIngestModule):
+class GPXParserDataSourceIngestModule(FileIngestModule):
```
eugene7646, line 91: We should rename all class names - GPXParserDataSourceIngestModuleFactory, GPXParserDataSourceIngestModule - to reflect that it's a file ingest module, not data source ingest module. Also, there are comments that also talk about data source ingest.

gdicristofaro: Got it.  That should be integrated now.

PR #5827, opened 2020-04-17
Fix date parsing errors when system locale is not Locale.US

jonas-koeritz:
While running the Recent Activity module autopsy threw some errors on my non us-english machine. I explecitly added `Locale.US` to all `SimpleDateFormat` constructor calls to make the parsing work on any system locale, the RegRipper output seems to always use the US locale.

### DIFF
--- a/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
+++ b/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
@@ -33,6 +33,7 @@ import java.io.StringReader;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.logging.Level;
+import java.util.Locale;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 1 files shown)


globeone:
Odd, why is this not using ISO 8601 worldwide standard interchange date format?

YYYY-MM-DD HH:MM:SS 

https://en.wikipedia.org/wiki/ISO_8601

jonas-koeritz:
The RegRipper output does not use ISO8601 (that's what my logfile showed me). After making these changes the problem was solved. Other parts of the same Java file already had a US locale hard-coded so this looks like a usable fix. RegRipper could add an option to output ISO dates, which would be a much nicer variant.

globeone:
I've reported this upstream to RegRipper
https://github.com/keydet89/RegRipper2.8/issues/65

The RegRipper devs are going to cry when they find out the extremely outdated version of RegRipper used in Autopsy. (unless that was upgraded in the .15 release?)

keydet89:
Globeone stated, "...why is this not using ISO 8601 worldwide standard interchange date format?

YYYY-MM-DD HH:MM:SS"

However, going to the Wikipedia link:

"Separating date and time parts with other characters such as space is not allowed in ISO 8601..."

I've already closed the issue in the RegRipper Github (the "RegRipper devs" is singular), as the request was to have the "RegRipper devs" update all 389 plugins.  As I said in that issue, I can't promise anything soon.

jonas-koeritz:
I'll see what I can do.

rcordovano:
@jonas-koeritz, thanks for your work. To expedite the inclusion of this PR in the upcoming Autopsy 4.18.0 release while getting the change reviewer @markmckinnon requested and dealing with merge conflicts, we have brought your work in via https://github.com/sleuthkit/autopsy/pull/6615. Therefore, I'm closing this PR.

Thanks again!

PR #5844, opened 2020-04-22
6296 update machine translation links

APriestman:


Core/src/org/sleuthkit/autopsy/texttranslation/translators/Bundle.properties
```diff
@@ -13,5 +13,5 @@ GoogleTranslatorSettingsPanel.untranslatedLabel.text=Untranslated:
...
-GoogleTranslatorSettingsPanel.instructionsTextArea.text=You will need a JSON credentials file which contains your service account key for your Google Translate account. Information on how to create a service account key is available here: https://cloud.google.com/iam/docs/creating-managing-service-account-keys 
+GoogleTranslatorSettingsPanel.instructionsTextArea.text=<div style\\="font-size: 12pt; font-family: Verdana, 'Verdana CE',  Arial, 'Arial CE', 'Lucida Grande CE', lucida, 'Helvetica CE', sans-serif; ">You will need a JSON credentials file which contains your service account key for your Google Translate account. Information on how to create a service account key is available here: <a style="color: #1E2A60;" href="https://cloud.google.com/iam/docs/creating-managing-service-account-keys">https://cloud.google.com/iam/docs/creating-managing-service-account-keys</a>.</div>
```
dannysmyda, line 16: Is there anyway the html can be kept out of the bundle message? If so, I think the Japanese messages can be reused as is

APriestman: I don't think so - that's being created through the Netbeans UI designer.

dannysmyda: ok

PR #5847, opened 2020-04-24
6154 system resource usage database program run and network activity artifacts

markmckinnon:


Core/src/org/sleuthkit/autopsy/contentviewers/contextviewer/ContextViewer.java
```diff
@@ -416,6 +422,36 @@ private String recentDocArtifactToString(BlackboardArtifact artifact) throws Tsk
         return sb.toString();
     }
 
+    /**
+     * Returns a display string with Program Execution
+     * artifact.
+     *
+     * @param artifact artifact to get doc from.
```
esaunders, line 429: "display string" instead of "doc"?

Core/src/org/sleuthkit/autopsy/contentviewers/contextviewer/ContextViewer.java
```diff
@@ -416,6 +422,36 @@ private String recentDocArtifactToString(BlackboardArtifact artifact) throws Tsk
...
+     *
+     * @return Display string with download URL and date/time.
```
esaunders, line 431: Don't see where the download URL comes into it.

PR #5848, opened 2020-04-24
6253 update comment on change

gdicristofaro:


rcordovano:
We are going to stop using tags as an alternative way to make file instance comments in the CR.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepository.java
```diff
@@ -449,9 +449,10 @@ CorrelationAttributeInstance getCorrelationAttributeInstance(CorrelationAttribut
      *
      * @param eamArtifact Artifact containing exactly one (1) ArtifactInstance.
      * @param knownStatus The status to change the artifact to
+     * @param isTagDelete If the action is to delete a previously added tag.
      */
-    void setAttributeInstanceKnownStatus(CorrelationAttributeInstance eamArtifact, TskData.FileKnown knownStatus) throws CentralRepoException;
-
+    void setAttributeInstanceKnownStatus(CorrelationAttributeInstance eamArtifact, TskData.FileKnown knownStatus, boolean isTagDelete) throws CentralRepoException;
+
```
rcordovano, line 455: @gdicristofaro and @wschaeferB... 

It is not appropriate to add a flag about tag deletion to a method whose stated purpose is to update the known status of a correlation attribute instance (even acknowledging that the method has a completely undocumented side effect of setting comment, not just the known status).

The CentralRepository interface defines the operations that can be performed on the data store. Application / domain logic does not belong in the data store layer of an architecture, it belongs in the application / domain logic layer. In this case, the decision whether or not to change the known status and the comment belongs where it was, in the application logic implemented by the CaseEventListener.

Another way to think about this is like this: the CentralRepository interface exists because we have envisioned having an option to choose from multiple relational and non-relational data store technologies for the CR. Putting the application logic details in CentralRepository interface implementations means that some future implementer of a NoSQL CR will need to **_duplicate_** the complicated logic that has been put into RdbmsCentralRepo (as in Relational Database Management System Central Repo). This would introduce a maintenance problem that would be magnified with every new implementation of CentralRepository.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
+                return;
+            
+            setContentKnownStatus(af, tagAdded.getComment(), TskData.FileKnown.BAD, false);
+        }
+        
+        private AbstractFile getAbstractFile(Content content) {
+            if (content instanceof AbstractFile) {
+                return (AbstractFile) content;
+            } 
+            else {
```
rcordovano, line 279: Our coding style for this project is to use end braces, i.e., "...} else {..." Are you remembering to format all files with the IDE before submitting them?

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
+                handleTagDeleted((ContentTagDeletedEvent) event);
+            }
+        }
+        
+        private void handleTagDeleted(ContentTagDeletedEvent tagDeletedEvent) {
+            // CONTENT_TAG_DELETED
+            // For deleted tags, we want to set the file status to UNKNOWN if:
+            //   - The tag that was just removed is notable in central repo
+            //   - There are no remaining tags that are notable 
+            //   - If there are other notable tags, the comment should be updated to most recent tag
```
rcordovano, line 225: This is tricky code. Again, I think we want to be very careful in our comments...

This first comment describes what the method does. It belongs in a header comment.

The logic in the method does two things:

1) We are not setting the status of the file that has been tagged. We are setting the known status of the correlation attribute instance in the CR corresponding to the tagged file. We want to change that status from notable to unknown if the tag that was deleted was the only tag on the file that confers notability. 
2) If the deleted tag conferred notability on the file and also had a comment, we want to remove the comment from the correlation attribute instance, if and only if the comment on the correlation attribute instance and the tag match. Otherwise, we do not know the source of the comment (it could have been applied through the add/edit CR comment action independent of any tagging) and will not remove the comment.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
+
+            if (!isNotableTagName(tagDeletedEvent.getDeletedTagInfo().getName())) {
+                // If the tag that got removed isn't on the list of central repo tags, do nothing
```
rcordovano, line 228: Again, let's be careful with out comments. There is no such thing as a "central repo tag."

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
+                // Get the remaining tags on the content object
                 long contentID = tagDeletedEvent.getDeletedTagInfo().getContentID();
+                Content content = Case.getCurrentCaseThrows().getSleuthkitCase().getContentById(contentID);
+                TagsManager tagsManager = Case.getCurrentCaseThrows().getServices().getTagsManager();
+                List<ContentTag> tags = tagsManager.getContentTagsByContent(content);
 
-                String tagName = tagDeletedEvent.getDeletedTagInfo().getName().getDisplayName();
-                if (!TagsManager.getNotableTagDisplayNames().contains(tagName)) {
-                    // If the tag that got removed isn't on the list of central repo tags, do nothing
+                AbstractFile af = getAbstractFile(content);
```
rcordovano, line 239: getAbstractFile hides the fact that we are casting. It would be clearer if we used our knowledge that only AbstractFiles can be tagged, despite appearances to the contrary, by calling getAbstractFileById instead of getContentById.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
-                    } else {
-                        // There's still at least one bad tag, so leave the known status as is
-                        return;
-                    }
-                } catch (TskCoreException | NoCurrentCaseException ex) {
-                    LOGGER.log(Level.SEVERE, "Failed to find content", ex);
-                    return;
+                if (hasNotableTag(tags)) {
+                    // There's still at least one bad tag, so leave the known status but update the comment to most recent
+                    String comment = getMostRecentNotableComment(tags);
```
rcordovano, line 245: This logic is not quite correct. Comments in the CR can come from notable tags or directly from the user by way of the add/edit CR comment action. As mentioned in an earlier PR comment, we should clear the comment if and only if the comment on the correlation attribute instance and the tag match. The logic to get a replacement comment from another "notable tag" is a nice innovation - kudos for thinking of it. However, the replacement should only be attempted if we have decided to clear the comment. Otherwise, we could be wiping out a comment from the add/edit action.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -154,82 +206,90 @@ private ContentTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
-                        }
-                    } else {
-                        // There's still at least one bad tag, so leave the known status as is
-                        return;
-                    }
-                } catch (TskCoreException | NoCurrentCaseException ex) {
-                    LOGGER.log(Level.SEVERE, "Failed to find content", ex);
-                    return;
+                if (hasNotableTag(tags)) {
+                    // There's still at least one bad tag, so leave the known status but update the comment to most recent
```
rcordovano, line 244: "Bad tags" is an outdated concept. Tags either confer notability on a file, or they do not. 

We no longer label files as "bad" in Autopsy. Files are either notable, known or unknown.  There are leftover references to "bad" all over the place, but we should not add new references nor keep old references to "badness" unless it happens to be a matter of public API.

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -249,91 +309,87 @@ private BlackboardTagTask(CentralRepository db, PropertyChangeEvent evt) {
...
-                long contentID = tagDeletedEvent.getDeletedTagInfo().getContentID();
-                long artifactID = tagDeletedEvent.getDeletedTagInfo().getArtifactID();
+                handleTagAdded((BlackBoardArtifactTagAddedEvent) event);
+            } 
+            else { 
+                handleTagDeleted((BlackBoardArtifactTagDeletedEvent) event);
+            }
+        } // BLACKBOARD_ARTIFACT_TAG_ADDED, BLACKBOARD_ARTIFACT_TAG_DELETED
+        
+        private void handleTagDeleted(BlackBoardArtifactTagDeletedEvent tagDeletedEvent) {
```
rcordovano, line 323: See remarks on the file tag added/deleted event handling; artifact tag added/deleted event handling is similar.

PR #5867, opened 2020-05-01
1452 - Fixed getNotableTagDisplayName

kellykelly3:


Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -894,6 +923,25 @@ public BlackboardArtifactTag getBlackboardArtifactTagByTagID(long tagId) throws
...
+        List<TagSet> tagSets = caseDb.getTaggingManager().getTagSets();
+        
+        for(TagSet set: tagSets) {
+            if(set.getName().startsWith(PROJECT_VIC_TAG_SET_HEADER)) {
+                return set;
+            }
+        }
+        
+        return null;
+    }
```
rcordovano, line 944: See comment in Jira.

Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -146,13 +147,41 @@ public static String getNotableTagLabel() {
...
+                }
+            }
+        } catch(TskCoreException ex) {
+            LOGGER.log(Level.WARNING, "Failed to retrieve category TagSet from DB");
+        } catch (NoCurrentCaseException ignored) {
+            /*
+             * No current case, nothing more to add to the set.
+             */
+        }
+
```
rcordovano, line 184: What are the clients of this method? Given that it is a static method, there is an implication that it is called when there is no open case. If that is so, we may be breaking something that counted on all of the tag definitions to be in the Tag.properties file. Please look up the clients of this method and we'll move further discussion to Jira.

rcordovano:
I am closing this PR, since, per discussion in Jira, we are going to implement a different solution.

PR #5878, opened 2020-05-07
6314: Integrate Message viewer as a Results sub-viewer

raman-bt:


raman-bt:
Needs some reworking.

PR #5879, opened 2020-05-08
6336 Add UI elements to persona window

ethteck:


Core/src/org/sleuthkit/autopsy/persona/Bundle.properties
```diff
@@ -1,3 +1,14 @@
...
 CTL_PersonasTopComponentAction=PersonasTopComponent
-CTL_PersonasTopComponent=Personas
\ No newline at end of file
+CTL_PersonasTopComponent=Personas
+PersonasTopComponent.jTextField1.text=John Doe
+PersonasTopComponent.jRadioButton1.text=Name
+PersonasTopComponent.jRadioButton2.text=Account
+PersonasTopComponent.jLabel2.text=Name:
+PersonasTopComponent.jTextField2.text=Johnathan Dough
+PersonasTopComponent.jLabel3.text=Aliases:
```
raman-bt, line 9: Recommend that as far as possible we rename the Swing components to match their logical intent, so they are easier to search and follow in the code.  For e.g. jSearchButton, instead of jButton1, jAliasesLabel, instead of jLabel3.....

PR #5880, opened 2020-05-08
6333 Add max number of images to preload and an exception firewall.

APriestman:


APriestman:
Depends on https://github.com/sleuthkit/sleuthkit/pull/1915

PR #5883, opened 2020-05-08
1452 - Fixed getNotableTagDisplayName

kellykelly3:


Core/src/org/sleuthkit/autopsy/casemodule/services/TagsManager.java
```diff
@@ -146,13 +146,35 @@ public static String getNotableTagLabel() {
...
+                        !tagDisplayNames.contains(tagName.getDisplayName())) {
+                    tagDisplayNames.add(tagName.getDisplayName());
+                }
+            }
+        } catch (NoCurrentCaseException ignored) {
+            /*
+             * No current case, nothing more to add to the set.
+             */
+        } catch(TskCoreException ex) {
+            LOGGER.log(Level.SEVERE, "Faile to get list of TagNams from tagsManager.");
```
gdicristofaro, line 176: I think this was meant to be "Failed to get list of TagNames from TagsManager."

PR #5896, opened 2020-05-14
6359 remove commenting

gdicristofaro:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -235,15 +234,15 @@ private void handleTagDeleted(ContentTagDeletedEvent evt) {
                 // then handle the event 
                 handleTagChange(content);
             } catch (NoCurrentCaseException | TskCoreException ex) {
-                LOGGER.log(Level.WARNING, "Error updating non-file object.");
+                Long contentID = (evt != null && evt.getDeletedTagInfo() != null) ? evt.getDeletedTagInfo().getContentID() : null;
```
APriestman, line 237: Is this check necessary? We already know evt.getDeletedTagInfo() is not null, and the call to getContentID() can not throw an exception.

gdicristofaro: Sure.

PR #5904, opened 2020-05-19
6301 check reg ripper error logs and report any errors

markmckinnon:


markmckinnon:
This pr needs to go in after https://github.com/sleuthkit/autopsy/pull/5876

dannysmyda:
@markmckinnon Could you merge in develop?

markmckinnon:
Develop Merged in, you are all set.

thirdparty/rr-full/plugins/eventlogs.pl
```diff
@@ -57,9 +57,11 @@ sub pluginmain {
 			::rptMsg($win_path);
 			::rptMsg("LastWrite Time ".gmtime($win->get_timestamp())." (UTC)");
 			my $cn;
-			if ($cn = $win->get_value("ComputerName")->get_data()) {
-				::rptMsg("ComputerName = ".$cn);				
-			}
+            if (defined($win->get_value("ComputerName"))) {
```
dannysmyda, line 60: Are these and the changes in all other .pl files still part of the upgrade or are these yours?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -418,10 +418,31 @@ private RegOutputFiles ripRegistryFile(String regFilePath, String outFilePathBas
...
     }
 
+    private void scanErrorLogs(String errFilePath) throws IOException {
+        File regfile = new File(errFilePath);
+        try (BufferedReader reader = new BufferedReader(new FileReader(regfile))) {
+            String line = reader.readLine();
+            while (line != null) {
+                line = line.trim();
+                if (line.contains("Error") || line.contains("@INC")) {
+                   logger.log(Level.WARNING, "Regripper file {0} contains errors from run", errFilePath); //NON-NLS
```
dannysmyda, line 438: Would it be worth also sending a notification to the user to let them know? Or is the fact that we are running regripper not explicitly stated in the Recent Activity module description?

thirdparty/rr-full/plugins/eventlogs.pl
markmckinnon: These changes are fixes to undefined values that cause an error in the plugin and it will not completely execute.  These are separate from the upgrade PR.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
markmckinnon: I just put the warning in the log for debugging purposes if needed. Since not all plugins are parsed in Autopsy this will log the plugins that actually get an error that autopsy may not use.

thirdparty/rr-full/plugins/eventlogs.pl
dannysmyda: Okay, should these changes also be submitted to the RegRipper project? Reason I ask is that if we ever perform an upgrade these changes will get lost or need to be redone.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
dannysmyda: Okay, last thing, could we make the string checking case insensitive?

thirdparty/rr-full/plugins/eventlogs.pl
markmckinnon: Regripper 2.8 is no longer being supported as of 5/28/2020 according to this http://windowsir.blogspot.com/2020/05/regripper-v30.html so adding a PR to it will never happen as this is end of life for this version we are using.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
markmckinnon: changed to check to for case insensitive, good catch.

PR #5906, opened 2020-05-20
6380 Groundwork for persona editing / creating, viewing functional

ethteck:


ethteck:
Most of the remaining codacy issues will be resolved in future PRs.

ethteck:
Most of the remaining codacy issues are fixed in my next PR, since these unused things are now being actually used. The others are dealing with auto-generated code, as far as I can tell. @raman-bt are you against going with this as-is for now? I guess you could make the case that I could do some refactoring, but all of these files change so much in the next PR.

PR #5908, opened 2020-05-21
1457 Add TagSet submenus

kellykelly3:


Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
```diff
@@ -79,13 +76,13 @@ public void mousePressed(MouseEvent evt) {
                 JList<?> list = (JList) evt.getSource();
                 int index = list.locationToIndex(evt.getPoint());
                 if (index > -1) {
-                    String value = tagsNamesListModel.getElementAt(index);
-                    tagNameSelections.put(value, !tagNameSelections.get(value));
+                    TagName tagName = tagsNamesListModel.getElementAt(index);
+                    tagNameSelections.put(tagName, !tagNameSelections.get(tagName));
```
gdicristofaro, line 80: Could this be changed to `Objects.equal(tagNameSelections.get(tagName), true)` to avoid a possible NPE?

Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
```diff
@@ -187,7 +187,7 @@ private void populateHashSetComponents() {
     List<TagName> getSelectedTagNames() {
         List<TagName> selectedTagNames = new ArrayList<>();
         for (TagName tagName : tagNames) {
-            if (tagNameSelections.get(tagName.getDisplayName())) {
+            if (tagNameSelections.get(tagName)) {
```
gdicristofaro, line 190: `if (Objects.equals(tagNameSelections.get(tagName))) {`

Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
```diff
@@ -208,41 +208,20 @@ HashDb getSelectedHashDatabase() {
...
+    private class TagsNamesListCellRenderer extends JCheckBox implements ListCellRenderer<TagName> {
+
         private static final long serialVersionUID = 1L;
 
         @Override
-        public Component getListCellRendererComponent(JList<? extends String> list, String value, int index, boolean isSelected, boolean cellHasFocus) {
+        public Component getListCellRendererComponent(JList<? extends TagName> list, TagName value, int index, boolean isSelected, boolean cellHasFocus) {
             if (value != null) {
                 setEnabled(list.isEnabled());
                 setSelected(tagNameSelections.get(value));
```
gdicristofaro, line 220: Objects.equals...

Core/src/org/sleuthkit/autopsy/actions/GetTagNameAndCommentDialog.java
```diff
@@ -128,8 +128,7 @@ private void display() {
 
             @Override
             public Component getListCellRendererComponent(JList<?> list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
-                String status = ((TagName) value).getKnownStatus() == TskData.FileKnown.BAD ? TagsManager.getNotableTagLabel() : "";
-                String newValue = ((TagName) value).getDisplayName() + status;
+                String newValue = TagUtils.getDecoratedTagDisplayName((TagName) value);
```
gdicristofaro, line 131: It looks like these are decorated right, but they are not ordered.  Is that a problem?

kellykelly3: The rendered is not involved in order.

gdicristofaro: Sorry, I didn't have a better place to put the comment, but I don't this class is ordering the items that are are displayed in the drop down for the tag and comment dialog.  I'm thinking closer to 161 should probably be ordered if ordering needs to be applied to this list as well.  I'm also realizing that is probably more in reference to another PR where that was the issue at hand.  Sorry for the confusion on my part.

kellykelly3: Line 159, it gets the map and then puts it into a tree which puts the tags in alphabetical order.

Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
kellykelly3: This code is a little weird and needs more clean up than I did.  But its not possible to have an NPE, tagNameSelections and tagsNamesListModel are populated from the same list.  Eventually if someone has more time I think all the extra lists include TagNameSelections could be removed.

Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
kellykelly3: Same as above.

Core/src/org/sleuthkit/autopsy/report/modules/taggedhashes/SaveTaggedHashesToHashDbConfigPanel.java
kellykelly3: Same as above.

PR #5916, opened 2020-05-27
6397 Add Persona functionality

ethteck:


Core/src/org/sleuthkit/autopsy/centralrepository/persona/AddAccountDialog.java
```diff
@@ -0,0 +1,337 @@
...
+            JOptionPane.showMessageDialog(this,
+                    Bundle.AddAccountDialog_empty_msg(),
+                    Bundle.AddAccountDialog_empty_Title(),
+                    JOptionPane.ERROR_MESSAGE);
+            return;
+        }
+        Collection<PersonaAccount> candidates;
+        try {
+            candidates = 
+                    PersonaAccount.getPersonaAccountsForIdentifierLike(identifierTextField.getText());
```
raman-bt, line 271: We added these  apis to CentralRepoAccount:
 - getAccountsWithIdentifierLike()
 - getAccountsWithIdentifier()

I think when looking for an account to create a persona, those are the ones we should use here.  

Also it isn't obvious if the user is supposed to type in a substring or a full account identifier.  We are using the api getAccountsWithIdentifierLike().  What if that returns multiple hits  - does the user get to pick one?

PR #5919, opened 2020-05-28
6392 send files to ingest stream

APriestman:


Core/src/org/sleuthkit/autopsy/casemodule/AddImageTask.java
```diff
@@ -52,7 +53,9 @@
     private final String sha1;
     private final String sha256;
     private final DataSourceProcessorProgressMonitor progressMonitor;
-    private final DataSourceProcessorCallback callback;
+    //private final DataSourceProcessorCallback callback;
```
rcordovano, line 56: Commented out, not deleted?

Core/src/org/sleuthkit/autopsy/casemodule/AddImageTask.java
```diff
@@ -73,40 +76,26 @@
...
+        this.imagePath = imageDetails.imagePath;
+        this.sectorSize = imageDetails.sectorSize;
+        this.timeZone = imageDetails.timeZone;
+        this.ignoreFatOrphanFiles = imageDetails.ignoreFatOrphanFiles;
+        this.md5 = imageDetails.md5;
+        this.sha1 = imageDetails.sha1;
+        this.sha256 = imageDetails.sha256;
+        this.imageWriterSettings = imageDetails.imageWriterSettings;
+        this.addDataSourceCallbacks = addDataSourceCallbacks;
+        this.addImageTaskCallback = addImageTaskCallback;
```
rcordovano, line 98: Perhaps just have an ImageDetails instance variable instead of duplicating the ImageDetails class within the AddImageTask class?

Core/src/org/sleuthkit/autopsy/casemodule/DefaultIngestStream.java
```diff
@@ -0,0 +1,55 @@
...
+ * to access them through this class.
+ */
+class DefaultIngestStream implements IngestStream {
+
+    private volatile boolean isClosed = false;
+
+    @Override
+    public void addDataSource(long dataSourceObjectId) throws IngestStreamClosedException {
+        // Do nothing
+        System.out.println("### DefaultAddImageTaskIngestStream - addDataSource " + dataSourceObjectId);
```
rcordovano, line 27: No println's, please.

Core/src/org/sleuthkit/autopsy/casemodule/ImageDSProcessor.java
```diff
@@ -192,6 +198,60 @@ public void run(DataSourceProcessorProgressMonitor progressMonitor, DataSourcePr
...
+                sha1 = null;
+            }
+            sha256 = configPanel.getSha256();
+            if (sha256.isEmpty()) {
+                sha256 = null;
+            }
+        }
+                
+        run(deviceId, imagePath, sectorSize, timeZone, ignoreFatOrphanFiles, md5, sha1, sha256, progress, callBack);
+    }
```
rcordovano, line 244: I think we can reduce code duplication using DefaultIngestStream to call run() from run(DataSourceProcessorProgressMonitor progressMonitor, DataSourceProcessorCallback callback).

Core/src/org/sleuthkit/autopsy/casemodule/LocalDiskDSProcessor.java
```diff
@@ -221,6 +234,32 @@ public void reset() {
...
+        
+        @Override
+        public void onFilesAdded(List<Long> fileObjectIds) throws AddDataSourceCallbacksException {
+            try {
+                ingestStream.addFiles(fileObjectIds);
+            } catch (IngestStreamClosedException ex) {
+                throw new AddDataSourceCallbacksException("Error adding files to ingest stream - ingest stream is closed", ex);
+            }
+        }
+    }
```
rcordovano, line 262: It looks like this class is the same as the one in ImageDSProcessor. We should probably have just one version of the class that is used by both DSPs.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobInputStream.java
```diff
@@ -0,0 +1,130 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskDataException;
+
+/**
+ * Implementation of IngestStream. Will collect data from the data source
+ * processor to be sent to the ingest pipeline.
+ */
+public class IngestJobInputStream implements IngestStream {
```
rcordovano, line 36: I don't think this class needs to be public.

PR #5924, opened 2020-05-29
6392 Change ingest stream error handling

APriestman:


Core/src/org/sleuthkit/autopsy/ingest/IngestJobInputStream.java
```diff
@@ -82,8 +85,14 @@ public DataSource getDataSource() throws TskCoreException {
...
 
     @Override
-    public void addFiles(List<Long> fileObjectIds) throws IngestStreamClosedException {
+    public void addFiles(List<Long> fileObjectIds) throws IngestStreamClosedException, TskCoreException {
         synchronized(this) {
+            if (isClosed) {
+                throw new IngestStreamClosedException("Can not add files - ingest stream is closed");
+            }
+            if (dataSource == null) {
+                throw new TskCoreException("Files can not be added without a data source");
```
rcordovano, line 94: I'm thinking this should be a TskDataException instead of a TskCoreException. Or an IllegalStreamStateException. I like the latter.

PR #5926, opened 2020-06-01
6355 geolocation npe

gdicristofaro:


Core/src/org/sleuthkit/autopsy/geolocation/datamodel/GeoLocationParseResult.java
```diff
@@ -0,0 +1,166 @@
...
+        return new GeoLocationParseResult(artifact, false, exception, null);
+    }
+
+    /**
+     * The result of splitting an iterable of GeoLocationParseResults into a
+     * list of successfully parsed items and a list of failures.
+     *
+     * @param <T> The parsed item type.
+     */
+    public static class SeparationResult<T> {
```
dannysmyda, line 83: What is the need for separating the failed vs the parsed? From what I can tell, you just use the returned lists to answer boolean queries (not null, is empty, etc).

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
```diff
@@ -619,44 +622,88 @@ static private String getWaypointListQuery(List<DataSource> dataSources) {
...
-     * @throws GeoLocationDataException
      */
-    static private List<Waypoint> getWaypointForArtifact(BlackboardArtifact artifact, ARTIFACT_TYPE type) throws GeoLocationDataException {
-        List<Waypoint> waypoints = new ArrayList<>();
+    private static List<GeoLocationParseResult<Waypoint>> getWaypointForArtifact(BlackboardArtifact artifact, ARTIFACT_TYPE type) {
+        List<GeoLocationParseResult<Waypoint>> waypoints = new ArrayList<>();
         switch (type) {
             case TSK_METADATA_EXIF:
-                waypoints.add(new EXIFWaypoint(artifact));
+                waypoints.add(parseWaypoint((a) -> new EXIFWaypoint(a), artifact));
```
dannysmyda, line 685: This is very confusing. Could you instead go for an API like parseWaypoint(artifact, EXIFWaypoint.class)?

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Track.java
```diff
@@ -130,14 +133,17 @@ private void buildPath(GeoTrackPoints points, BlackboardArtifact artifact) throw
      */
     private GeoTrackPoints getPointsList(Map<BlackboardAttribute.ATTRIBUTE_TYPE, BlackboardAttribute> attributeMap) throws GeoLocationDataException {
         BlackboardAttribute attribute = attributeMap.get(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_GEO_TRACKPOINTS);
-        if (attribute != null) {
-            try {
-                return BlackboardJsonAttrUtil.fromAttribute(attribute, GeoTrackPoints.class);
-            } catch (InvalidJsonException ex) {
-                throw new GeoLocationDataException("Unable to parse track points in TSK_GEO_TRACKPOINTS attribute", ex);
-            }
+        if (attribute == null) {
```
dannysmyda, line 136: Can this become an illegal argument? And preferably out in the constructor of Track? I'm worried that GeoLocationDataException is becoming a 'god' exception.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/GeoLocationParseResult.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public boolean isSuccessfullyParsed() {
+        return successfullyParsed;
+    }
+
+    /**
+     * @return The exception caused in attempting to parse the GeoLocation
+     *         object if there was an exception.
+     */
+    public GeoLocationDataException getException() {
```
dannysmyda, line 147: Is getException() or getArtifact() used?

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/GeoLocationParseResult.java
```diff
@@ -0,0 +1,166 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.collections4.ListUtils;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+
+/**
+ * The result of attempting to parse a GeoLocation object.
+ */
+public class GeoLocationParseResult<T> {
```
dannysmyda, line 29: Since these are aggregated everywhere and then partition and managed in code outside of this class, what if the granularity of this class was not on a single ParseResult, but multiple? I think that might simplify things, what do you think?

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/GeoLocationParseResult.java
gdicristofaro: I initially created this because of a misunderstanding I had that I would need to report the track items that failed to parse to the user along with returning the items that successfully parsed.  Since the error that is reported to the user is generalized, this could potentially be simplified, but methods like `GeoPath.getTracks` and 'WaypointBuilder.getWaypointForArtifact` will need to have something that returns an object containing the list of successfully parsed items for display and some indication that there was an error.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/WaypointBuilder.java
gdicristofaro: How about something like `waypoints.add(parseWaypoint(EXIFWaypoint::new, artifact));`?

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/GeoLocationParseResult.java
gdicristofaro: Sure.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Track.java
gdicristofaro: I had a conversation where I think we decided on a `GeoLocationDataException`, because, similar to json parsing, the artifact attributes being passed were in a malformed state if they did not contain track information.

PR #5928, opened 2020-06-01
6299 Updated links on About page.

APriestman:
Open URLs in browser instead of Autopsy.


rcordovano:
There are some Codacy issues to look at.

PR #5934, opened 2020-06-02
Move Snapshot class.

APriestman:
Make AddImageTaskCallback an interface.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/casemodule/DefaultAddImageTaskCallback.java
+++ b/Core/src/org/sleuthkit/autopsy/casemodule/DefaultAddImageTaskCallback.java
@@ -1,0 +1,71 @@
+/*
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2020 Basis Technology Corp.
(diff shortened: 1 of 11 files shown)


Core/src/org/sleuthkit/autopsy/casemodule/DefaultAddImageTaskCallback.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import org.sleuthkit.autopsy.ingest.IngestStream;
+import org.sleuthkit.autopsy.corecomponentinterfaces.DataSourceProcessorCallback;
+import org.sleuthkit.autopsy.corecomponentinterfaces.DataSourceProcessorCallback.DataSourceProcessorResult;
+import org.sleuthkit.datamodel.Content;
+
+/**
+ * Called on completion of the add image task.
+ */
+class DefaultAddImageTaskCallback implements AddImageTaskCallback {
```
rcordovano, line 30: I'd like us to call this StreamingAddImageTaskCallback. We can get that in another round.

PR #5940, opened 2020-06-04
6398 Persona deletion functionality

ethteck:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/Persona.java
```diff
@@ -376,7 +385,8 @@ private static Persona getPersonaByUUID(String uuid) throws CentralRepoException
     public static Collection<Persona> getPersonaByName(String partialName) throws CentralRepoException {
 
         String queryClause = PERSONA_QUERY
-                + "WHERE LOWER(p.name) LIKE " + "LOWER('%" + partialName + "%')" ;
+                + "WHERE p.status_id != " + PersonaStatus.DELETED.status_id +
```
raman-bt, line 388: I think other methods using the PERSONA_QUERY also need to filter out deleted personas.  
We should probably just add the WHERE condition to PERSONA_QUERY. 

Similar check probably needs to be added to queries in PersonaAccount ?

Core/src/org/sleuthkit/autopsy/centralrepository/persona/PersonaDetailsPanel.java
```diff
@@ -712,6 +712,10 @@ void setMode(Component parent, PersonaDetailsMode mode, Persona persona) {
         }
         initializeFields();
     }
+    
+    public void setName(String name) {
```
raman-bt, line 716: setName is already a method defined in the JPanel class

We should rename this to setPersonaName()

PR #5941, opened 2020-06-04
Hide persona edit btns

ethteck:
Merge after #5940


raman-bt:
I've reviewed and tested this, and it's good to go.  But since this PR has changes from another PR that is still in progress, I wont approve it yet.

PR #5947, opened 2020-06-07
1477 - Added ArtifactViewers to CVT

kellykelly3:


rcordovano:
There is a minor Codacy issue top address, please. Thanks!

kellykelly3:
> There is a minor Codacy issue top address, please. Thanks!

@rcordovano I ignored that Codacy issue, but that's how the code is in the other files that do this little trick.  I guess what codacy is saying is correct, but I wonder what the motivation was to qualify this was in the first place.  If we change this here, we should change it in the other places that have similar code, but did not get a codacy warning.

rcordovano:
Looking at CallLogViewer more closely, the qualification of the this reference is being done because it is within the scope of a non-static inner class, in this case an anonymous inner class. I'm not sure why Codacy is flagging this for the reason it says it has for doing so. What should be flagged is that the constructor is leaking the "this" reference of an incompletely constructed object, which is playing with fire. We need to implement the "two-stage construction" idiom where we have a static method that first constructs the object, then in "stage two," creates the PropertyChangeListener.

kellykelly3:
> Looking at CallLogViewer more closely, the qualification of the "this" reference is being done because it is within the scope of a non-static inner class, in this case an anonymous inner class. I'm not sure why Codacy is flagging this for the reason it says it has for doing so. What should be flagged is that the constructor is leaking the "this" reference of an incompletely constructed object, which is playing with fire. We need to implement the "two-stage construction" idiom where we have a static method that first constructs the object, then in "stage two," creates the PropertyChangeListener.

@rcordovano This exact same code is in several places in the CVT window and in Timeline.  If you would like me to change it, may I suggest a new story focused on that so we change the code in all places, not just here?

rcordovano:
Let's do this...

1. Fix this class.
2. Make a story about the other places and reference the solution in this class.
3. Fix the merge conflict and I will take the PR.

Thanks!

PR #5949, opened 2020-06-08
6410 contactbook personas

raman-bt:


Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+        personasPanel.add(filler1, constraints);
+
+        constraints.gridx++;
+        javax.swing.JLabel primaryPersonaNameLabel = new javax.swing.JLabel();
+        primaryPersonaNameLabel.setText(Bundle.ContactArtifactViewer_persona_searching());
+        gridBagLayout.setConstraints(primaryPersonaNameLabel, constraints);
+        personasPanel.add(primaryPersonaNameLabel);
+        
+        personasPanel.setLayout(gridBagLayout);
+        personasPanel.revalidate();
```
kellykelly3, line 434: This should have a call to repaint(),

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+     * 
+     * @throws CentralRepoException 
+     */
+    @NbBundle.Messages({
+        "ContactArtifactViewer_persona_searching= Searching..."
+    })
+    private void initiatePersonasSearch(List<BlackboardAttribute> accountAttributesList) throws CentralRepoException {
+        
+        // create a Gridbag layout
+        // create a gridbag layout to show each participant on one line
```
kellykelly3, line 412: Fix the comment

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -231,6 +311,16 @@ public boolean isSupported(BlackboardArtifact artifact) {
         return artifact.getArtifactTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_CONTACT.getTypeID();
     }
 
+    /**
+     * Clear any artifact specific state.
+     */
+    private void reset() {
```
kellykelly3, line 317: Brian was looking for one resetComponent method that would reset the state of all of the components.

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -158,43 +194,81 @@ private void initComponents() {
...
     }// </editor-fold>//GEN-END:initComponents
 
     @Override
     public void setArtifact(BlackboardArtifact artifact) {
 
         // wipe the panel clean
         this.removeAll();
+        
+        reset();
         initComponents();
```
kellykelly3, line 248: initComponent should not be called in setArtifacts.  It should only be called in the constructor.

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
         sectionPanel.revalidate();
     }
 
+    /**
+     * Kicks of a search for personas, based in the list of attributes.
```
kellykelly3, line 400: Kicks off a search

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+     * 
+     * @param accountAttributesList a list of account identifying attributes.
```
kellykelly3, line 402: A list of account

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+        constraints.gridx++;
+        javax.swing.JLabel primaryPersonaNameLabel = new javax.swing.JLabel();
+        primaryPersonaNameLabel.setText(Bundle.ContactArtifactViewer_persona_searching());
+        gridBagLayout.setConstraints(primaryPersonaNameLabel, constraints);
+        personasPanel.add(primaryPersonaNameLabel);
+        
+        personasPanel.setLayout(gridBagLayout);
+        personasPanel.revalidate();
+        
+        // kick off a background task to serach for personas for the contact
```
kellykelly3, line 436: Kick off a

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+            constraints.gridx++;
+            javax.swing.JLabel missingAccountIdentifierLabel = new javax.swing.JLabel();
+            missingAccountIdentifierLabel.setText(missingAccount.getIdentifier());
+            gridBagLayout.setConstraints(missingAccountIdentifierLabel, constraints);
+            personasPanel.add(missingAccountIdentifierLabel);
+        }
+    }
+    
+    
+    private javax.swing.Box.Filler createFiller(int width, int height ) {
```
kellykelly3, line 581: Method header

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+            // Pre populate the persona name and accounts if we have them.
+            PersonaDetailsPanel personaPanel = createPersonaDialog.getDetailsPanel();
+            
+            if (contactName != null) {
+                personaPanel.setPersonaName(contactName);
+            }
+
+            // pass the list of accounts to the dialog
+            for (CentralRepoAccount account : contactUniqueAccountsList) {
+                personaPanel.addAccount(account, "Account found in Contact artifact.", Persona.Confidence.UNKNOWN);
```
kellykelly3, line 722: Should that string be in a @messages block?

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+   /**
+     * Thread to search for a personas for all account identifier attributes for a contact.
+     */
+    private class ContactPersonaSearcherTask extends SwingWorker<Map<Persona, ArrayList<CentralRepoAccount>>, Void> {
+
+        private final List<BlackboardAttribute> accountAttributesList;
+        private final List<CentralRepoAccount> uniqueAccountsList = new ArrayList<>();
+       
+        
+        ContactPersonaSearcherTask(List<BlackboardAttribute> accountAttributesList) {
```
kellykelly3, line 594: Method header

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
```diff
@@ -305,16 +396,410 @@ private void updateSection(List<BlackboardAttribute> sectionAttributesList, JLab
...
+            }
+        
+        }
+    } 
+    
+    /**
+     * A wrapper class that bags the UI components that need to be updated when 
+     * a persona search task or a create dialog returns.
+     */
+    private class PersonaUIComponents {
```
kellykelly3, line 677: This whole class is missing method headers.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoAccount.java
```diff
@@ -209,9 +245,17 @@ public void process(ResultSet rs) throws CentralRepoException, SQLException {
...
+
+        try {
+            normalizedAccountIdentifier = normalizeAccountIdentifier(accountIdentifier);
+        } catch (TskCoreException ex) {
+            throw new CentralRepoException("Failed to normalize account identifier.", ex);
+        }
+
         String queryClause = ACCOUNTS_QUERY_CLAUSE
-                + " WHERE LOWER(accounts.account_unique_identifier) = LOWER('" + accountIdentifier + "')";
+                + " WHERE LOWER(accounts.account_unique_identifier) = LOWER('" + normalizedAccountIdentifier + "')";
```
kellykelly3, line 258: Would it make sense for normalizeAccountIdentifier to take care of putting the  identifier into lowercase?  An email address passed into that method would already be in all lower case as CommunicationsUtils does that.

Core/src/org/sleuthkit/autopsy/centralrepository/persona/PersonaDetailsDialog.java
```diff
@@ -44,6 +44,10 @@
         "PersonaDetailsDialogEditTitle=Edit Persona",
         "PersonaDetailsDialogViewTitle=View Persona",})
     public PersonaDetailsDialog(Component parent, PersonaDetailsMode mode, Persona persona, PersonaDetailsDialogCallback callback) {
+        // by default, display the dialog right away
+        this(parent, mode, persona, callback, true);
+    }
+    public PersonaDetailsDialog(Component parent, PersonaDetailsMode mode, Persona persona, PersonaDetailsDialogCallback callback, boolean displayDialog) {
```
kellykelly3, line 50: Instead of passing in whether or not to display the dialog I would just have the calling code call display.  This will also help fix the lightbulb warnings in netbeans that you are not using your new instances.

Core/src/org/sleuthkit/autopsy/contentviewers/ContactArtifactViewer.java
raman-bt: I'm not sure if this should be in the messages. It's not a message being displayed to the user that needs to be localized - this is going to go in the database.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepoAccount.java
raman-bt: Normalization of other account types may do different things.  I don't know if we should assume it will convert to lower case.  We just had a discussion today to allow for different normalizations for Accounts  (as created by CommunicationsUtils)  Vs. CentralRepoAccounts

Core/src/org/sleuthkit/autopsy/centralrepository/persona/PersonaDetailsDialog.java
raman-bt: This dialog is created from several places.  (including persona manager)  Everywhere else its OK to immediately display the dialog.  So, far this is the first place where I needed to delay the display in order to pre-populate it.  Forcing every client to explicitly call display seems excessive.

PR #5956, opened 2020-06-09
5891 shutdown timeline once

gdicristofaro:


Core/src/org/sleuthkit/autopsy/timeline/TimeLineModule.java
```diff
@@ -97,14 +97,8 @@ public void propertyChange(PropertyChangeEvent evt) {
...
             if (Case.Events.valueOf(evt.getPropertyName()).equals(CURRENT_CASE)) {
-                // we care only about case closing here
-                if (evt.getNewValue() == null) {
-                    synchronized (controllerLock) {
-                        if (controller != null) {
-                            SwingUtilities.invokeLater(controller::shutDownTimeLine);
-                        }
-                        controller = null;
-                    }
+                synchronized (controllerLock) {
```
esaunders, line 100: What was the significance (if any) of the "if (evt.getNewValue() == null)" check in the original code?

rcordovano: When you see if "(evt.getNewValue() == null)" when handling a CURRENT_CASE event it is being done to determine whether the current case changed because a case was opened or closed by looking for which property has a reference to a case.

esaunders: Ok, does that mean that the check should not have been removed in the new code then?

gdicristofaro: Looking at `TimeLineModule.getController()` it appears that if the controller is null, a new controller is fetched based on the open case.  It seems to me that this should occur if a case is closed or if a different case is opened.  Was this an incorrect assumption?

esaunders: Beats me. @rcordovano ?

PR #5959, opened 2020-06-09
6444 Improve account add error message

ethteck:


Core/src/org/sleuthkit/autopsy/centralrepository/persona/AddAccountDialog.java
```diff
@@ -255,7 +255,7 @@ private void display() {
         "AddAccountDialog_search_failure_Title=Account add failure",
         "AddAccountDialog_search_failure_msg=Central Repository account search failed",
         "AddAccountDialog_search_empty_Title=Account not found",
-        "AddAccountDialog_search_empty_msg=Account not found for given identifier and type",})
+        "AddAccountDialog_search_empty_msg=Account not found in the Central Repository for the given identifier and type",})
```
raman-bt, line 258: Since this is message text that displays in the dialog, I think there should be a period at the end.
Same goes for the AddAccountDialog_search_failure_msg above.

PR #5962, opened 2020-06-09
6453 persona required fields

ethteck:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/Persona.java
```diff
@@ -392,6 +392,34 @@ private static Persona getPersonaByUUID(String uuid) throws CentralRepoException
...
+                + " FROM accounts"
+                + " JOIN persona_accounts as pa ON pa.account_id = accounts.id"
+                + " JOIN personas as p ON p.id = pa.persona_id"
+                + " JOIN examiners as e ON e.id = p.examiner_id"
+                + " WHERE LOWER(accounts.account_unique_identifier) LIKE LOWER('%" + partialName + "%')"
+                + " AND p.status_id != " + Persona.PersonaStatus.DELETED.getStatusId()
+                + " GROUP BY p.id";
+
+        PersonaQueryCallback queryCallback = new PersonaQueryCallback();
+        CentralRepository.getInstance().executeSelectSQL(queryClause, queryCallback);
```
raman-bt, line 418: Check for null.  This might be from another PR.

PR #5966, opened 2020-06-10
6454 Added confirmation before deleting persona

ethteck:


ethteck:
We decided the codacy change won't work for this use case. Keeping it as-is.

PR #5975, opened 2020-06-15
6456 redo calllogviewer

raman-bt:


kellykelly3:
@raman-bt Make sure all of the files are formatted.

PR #5976, opened 2020-06-15
6415 Launch streaming ingest from wizard.

APriestman:


Core/src/org/sleuthkit/autopsy/corecomponentinterfaces/DataSourceProcessor.java
```diff
@@ -122,16 +122,18 @@
      * return results.
      *
      * This method should not be called unless isPanelValid returns true, and 
-     * should only be called for DSPs that support ingest streams.
+     * should only be called for DSPs that support ingest streams. The ingest settings
+     * must be complete before calling this method.
      * 
+     * @param settings        The ingest job settings
```
rcordovano, line 128: Missing the period.

PR #5979, opened 2020-06-15
6473 diff script

gdicristofaro:


release_scripts/diffscript.py
```diff
@@ -0,0 +1,219 @@
+"""This script determines the updated, added, and deleted properties from the '.properties-MERGED' files
+and generates a csv file containing the items changed.  This script requires gitpython and jproperties.
+"""
```
markmckinnon, line 3: Can you also add that it requires Python 3+ also

PR #5980, opened 2020-06-16
6469: Check if CR is disabled in Persona data model class apis.

raman-bt:


kellykelly3:
@raman-bt Please remove the throwing of the IllegalStateException from all of the method headers.

raman-bt:
Ditching this PR. A new approach has been defined to handle this.

PR #5985, opened 2020-06-17
6381 mbox files greater then 2gb will not be processed in email parser

markmckinnon:


rcordovano:
Please address the issue flagged by Codacy, thanks.

PR #5986, opened 2020-06-17
1481 integrate persona message viewer

kellykelly3:


Core/src/org/sleuthkit/autopsy/contentviewers/MessageAccountPanel.java
```diff
@@ -0,0 +1,408 @@
+/*
```
raman-bt, line 1: Update header.

Core/src/org/sleuthkit/autopsy/contentviewers/MessageAccountPanel.java
```diff
@@ -0,0 +1,408 @@
...
+            List<AccountContainer> dataList = new ArrayList<>();
+
+            CommunicationsManager commManager = Case.getCurrentCase().getSleuthkitCase().getCommunicationsManager();
+            List<Account> accountList = commManager.getAccountsRelatedToArtifact(artifact);
+            for (Account account : accountList) {
+                if (isCancelled()) {
+                    return new ArrayList<>();
+                }
+
+                Collection<PersonaAccount> personAccounts = PersonaAccount.getPersonaAccountsForAccount(account);
```
raman-bt, line 102: I think some where there should be a check for CR.isEnabled(), and if so, no persona serches should be made and persona labels/buttons should be disabled...?

kellykelly3: I disabled the tab if Cr.isEnabled is fasle.  See MessageArtifactViewer setArtifact.

PR #5994, opened 2020-06-19
6415 Call IngestManager.startIngestJob instead of IngestJob.startIngest fr…

APriestman:
…om the stream.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/ingest/IngestJob.java
+++ b/Core/src/org/sleuthkit/autopsy/ingest/IngestJob.java
@@ -106,7 +106,7 @@ public final class IngestJob {
      * ingest modules specified via ingest job settings. Either all of the files
      * in the data source or a given subset of the files will be analyzed.
      *
-     * @param dataSource The data source to be analyzed
+     * @param dataSource The data source to be analyzed.
      * @param files      A subset of the files for the data source.
      * @param settings   The ingest job settings.
      */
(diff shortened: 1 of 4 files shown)


Core/src/org/sleuthkit/autopsy/ingest/IngestJob.java
```diff
@@ -204,8 +203,13 @@ private IngestJobPipeline getStreamingIngestPipeline() {
 	 * Set up the pipeline(s)
 	 */
 	if (ingestMode == Mode.STREAMING) {
-	    IngestJobPipeline ingestJobPipeline = new IngestJobPipeline(this, streamingIngestDataSource, settings);
-	    this.ingestJobPipelines.put(ingestJobPipeline.getId(), ingestJobPipeline);
+            if (streamingIngestDataSource == null) {
+                logger.log(Level.SEVERE, "Error starting streaming ingest - data source not set");
+                return new ArrayList<>();
```
rcordovano, line 208: I think that we need to return a non-empty list or otherwise signal that there was an error and the job won't be executing. Then we can leaving logging to the higher level client code, as has been the case before. Maybe we need a start(DataSource) overload after all?

APriestman: This code will change once the DSP gets moved (it won't be possible for the data source to be missing), so I don't think it's worth doing any major changes right now.

rcordovano: OK

PR #5996, opened 2020-06-19
1484 - Fixed SQL error in PersonaAccount add

kellykelly3:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/PersonaAccount.java
```diff
@@ -379,7 +379,7 @@ static void modifyPersonaAccount(long id, Persona.Confidence confidence, String
             throw new CentralRepoException("Failed to modify persona account, Central Repo is not enabled");
         }
         
-        String updateClause = "UPDATE persona_accounts SET confidence_id = " + confidence.getLevelId() + ", justification = \"" + justification + "\" WHERE id = " + id;
+        String updateClause = "UPDATE persona_accounts SET confidence_id = " + confidence.getLevelId() + ", justification = '" + justification + "' WHERE id = " + id;
```
raman-bt, line 382: The justification itself could have a single quote in it.
I think we may need to check and escape those.

kellykelly3: Do we generally do that?  Do we look for other characters that should be escaped from a string?  That doesn't seem like a just here type of issue.  Like what prevents the justification from being an SQL statement?

raman-bt: This is something an end user will type in UI. There's a good possibility that it could have a single quote :)

I'm not saying to check it in every string, or check of all kinds of characters.  But a single quote in the justification string will cause the same kind of SQL error you're fixing.  If you search, you will find calls to escapeSingleQuotes() method in Persona API and elsewhere.

kellykelly3: I get that, my suggestion is that escapeSingleQuotes is not good enough.  What do we do in other places where the user can supply a string field like this?

kellykelly3: Since it sounds like there is more work in this area might be coming.  I will just remove the other instances of double quotes for this Jira\PR

rcordovano:
This PR does not auto-merge. Please address this, thanks.

raman-bt:
@rcordovano: can this be merged?

kellykelly3:
@rcordovano & @raman-bt  I keep fixing the merge conflicts and more are made before this can be merged.  It's getting a little fustrating.

raman-bt:
@rcordovano: could you merge it please?

PR #6001, opened 2020-06-22
1486 - Fixed getOrCreateAccount race condition

kellykelly3:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/RdbmsCentralRepo.java
```diff
@@ -1082,28 +1082,20 @@ public void addArtifactInstance(CorrelationAttributeInstance eamArtifact) throws
      */
     @Override
     public CentralRepoAccount getOrCreateAccount(CentralRepoAccountType crAccountType, String accountUniqueID) throws CentralRepoException {
-        // Get the account fom the accounts table
-        CentralRepoAccount account = getAccount(crAccountType, accountUniqueID);
+        String query = "INSERT INTO accounts (account_type_id, account_unique_identifier) "
+                + "VALUES ( " + crAccountType.getAccountTypeId() + ", '"
+                + accountUniqueID + "' ) ON CONFLICT DO NOTHING";
```
raman-bt, line 1087: The syntax for ON CONFLICT DO NOTHING is different for SQLite vs Postgres.  We need to be able to handle both. 

You would need to use the  method getPlatformSpecificInsertSQL(String sql)

kellykelly3:
SOmethign is screwy with my branches.  They are not updating from my push.  I am closing this PR and have opened PR https://github.com/sleuthkit/autopsy/pull/6087 in its place.

PR #6003, opened 2020-06-22
6155 create background activity moderator tsk prog run artifacts

markmckinnon:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Bundle.properties-MERGED
```diff
@@ -5,10 +5,15 @@ ChromeCacheExtract_adding_artifacts_msg=Chrome Cache: Adding %d artifacts for an
 ChromeCacheExtract_adding_extracted_files_msg=Chrome Cache: Adding %d extracted files for analysis.
 ChromeCacheExtract_loading_files_msg=Chrome Cache: Loading files from %s.
 ChromeCacheExtractor.moduleName=ChromeCacheExtractor
+# {0} - module name
```
esaunders, line 8: Where do these additional lines come from?

markmckinnon: Good question. I normally see them appear after I compile it and I usually dismiss them since I do not touch the file but this time they got in. I can get rid of them as I do not know where they come from.

rcordovano: I recently made a commit that was intended to bring the MERGED properties files in line with the bundle.properties files and the message annotations in the code. Please look into those sources of what we see in the MERGED file to determine if the checked in file is out of sync or Mark's branch is out of sync, thanks.

markmckinnon: Out of sync, did a merge upstream of develop and they all went away except the change I made.

PR #6004, opened 2020-06-22
KWS/Solr UI and logic changes (6447)

eugene7646:


eugene7646:
@rcordovano bumping this up. We should merge this PR before you and I go on vacations, especially since this is not going into the current release.

PR #6007, opened 2020-06-23
6529 Added scroll panes to Personas window

ethteck:


ethteck:
Please wait for #6002 to be merged before reviewing.

kellykelly3:
@ethteck Are all these files changed for this fix or does it include changes from 6521?  I only ask because I notice that a change that was in 6521 in Person to fix the delete method is also in this PR.

ethteck:
This includes the changes from #6002 per my comment above, but the diff is disappointingly not cleaner after the merge. I'll try to fix it..

ethteck:
There we go. Should be better now @kellykelly3

Core/src/org/sleuthkit/autopsy/centralrepository/persona/PersonasTopComponent.java
```diff
@@ -276,6 +277,7 @@ private void initComponents() {
         editBtn = new javax.swing.JButton();
         deleteBtn = new javax.swing.JButton();
         jSeparator1 = new javax.swing.JSeparator();
+        jScrollPane2 = new javax.swing.JScrollPane();
```
kellykelly3, line 280: Please give the scroll panes more descriptive names.  Please fix jSplitePane1 while you are at it as well.

kellykelly3:
@ethteck please merge develop into this branch so I can do a clean build.  There is a bug that I fixed yesterday.

PR #6018, opened 2020-06-25
6532 - Auto resize content viewer image

kellykelly3:


Core/src/org/sleuthkit/autopsy/contentviewers/MediaViewImagePanel.java
```diff
@@ -131,6 +131,8 @@
     private double zoomRatio;
     private double rotation; // Can be 0, 90, 180, and 270.
 
+    boolean autoResize = true; // Auto resize when the user changes the size
```
wschaeferB, line 134: can be private

PR #6022, opened 2020-06-26
6565 better interesting files set messages

gdicristofaro:


Core/src/org/sleuthkit/autopsy/modules/interestingitems/InterestingItemsFilesSetSettings.java
```diff
@@ -184,6 +201,9 @@ private static ParentPathCondition readPathCondition(Element ruleElement) throws
      * @throws
      * org.sleuthkit.autopsy.modules.interestingitems.FilesSetsManager.FilesSetsManagerException
      */
+    @Messages({
+        "# {0} - regex",
+        "InterestingItemsFilesSetSettings.readDateCondition.failedCompiledRegex=Error detmining ''{0}'' number",})
```
dannysmyda, line 206: Typo in determining

PR #6034, opened 2020-06-29
6528 seven zip initialization to core installer

gdicristofaro:


Core/src/org/sleuthkit/autopsy/core/Installer.java
```diff
@@ -207,6 +209,13 @@ private static void loadDynLibraries() {
             } catch (UnsatisfiedLinkError e) {
                 logger.log(Level.SEVERE, "Error loading LIBPQ library, ", e); //NON-NLS
             }
+            
+            try {
```
esaunders, line 213: I don't think loadDynLibraries() is the best spot for this. Everything else in this method is being loaded with a call to System.loadLibrary().
I would recommend putting this into it's own method (like we do for JavaFX and Jython) and adding a comment on why this is being done in the installer.

Core/src/org/sleuthkit/autopsy/core/Installer.java
```diff
@@ -395,6 +398,18 @@ public void restored() {
     }
     
     
+    /**
+     * Initializes 7zip-java bindings.
```
esaunders, line 402: How about an explanation as to why this is being done in the installer?

esaunders:
@gdicristofaro I just noticed the Codacy issue. Can you take care of it please?

PR #6039, opened 2020-06-30
6411 - Added refresh throttler to FileTypesByMimeType

dannysmyda:


Core/src/org/sleuthkit/autopsy/datamodel/FileTypesByMimeType.java
```diff
@@ -212,6 +207,37 @@ public static boolean isEmptyMimeTypeNode(Node node) {
...
+
+    private class FileTypesByMimeTypeRefresher implements RefreshThrottler.Refresher {
+
+        @Override
+        public void refresh() {
+            refreshMimeTypes();
+        }
+
+        @Override
+        public boolean isRefreshRequired(PropertyChangeEvent evt) {
```
esaunders, line 235: Did we come to the conclusion that there was no point in checking to see if the file actually has a mime type before refreshing?

dannysmyda: Yes

PR #6044, opened 2020-07-02
6491 geolocation doc

APriestman:


docs/doxygen-user/geolocation.dox
```diff
@@ -8,15 +8,17 @@ The Geolocation window shows artifacts that have longitude and latitude attribut
...
+To open the Geolocation window, click on the "Geolocation" button near the top of the main window, or go to "Tools" and then select "Geolocation".
 
 \subsection geo_navigation General Usage
 
-You can move the map by clicking and dragging, and zoom using either the mouse wheel or the slider in the bottom left of the map. If a map tile is not available the tile will appear grey but the waypoints will still be displayed. This is more likely to happen when changing the default \ref geo_map_options.
+You can move the map by clicking and dragging, and zoom using either the mouse wheel or the slider in the bottom left of the map. If a map tile is not available the tile will appear grey but the waypoints will still be displayed. This is more likely to happen when changing the default \ref geo_map_options. Different types of waypoint will be displayed different colors. You can use the key in the lower left to easily identify the type of each waypoint. Some types will also use different icons on the map. For example, individual track points will be displayed as smaller circles. The entire track will be highlighted when selecting an individual track point.
```
gdicristofaro, line 15: I think this sentence:
"If a map tile is not available the tile will appear grey but the waypoints will still be displayed."

should be
"If a map tile is not available, the tile will appear grey, but the waypoints will still be displayed."

gdicristofaro, line 15: I think this:
"Different types of waypoint will be displayed different colors."
probably needs to be:
"Different types of waypoints will be displayed in different colors."

PR #6045, opened 2020-07-02
6575-Error_parsing-cached-GPS-locations-to-blackboard

markmckinnon:
Check number of entries of cache files to see if it is greater then the file size, if it is then skip processing of the file, otherwise process the file.


eugene7646:
@rcordovano FYI, I won't be back until next week.

PR #6048, opened 2020-07-06
6425: CommunicationsManager is validating account identifiers too rig…

raman-bt:
…orously

6516: CentralRepo accounts should be created with a normalized account identifier.
6507: CR validation does not include a length check


wschaeferB:
Requires sleuthkit PR: https://github.com/sleuthkit/sleuthkit/pull/1966

PR #6049, opened 2020-07-06
6516 take2 normalize cr account

raman-bt:


raman-bt:
DEPENDS ON: https://github.com/sleuthkit/autopsy/pull/6048 

review and merge AFTER the above PR is merged.

wschaeferB:
Conflicts need to be resolved

wschaeferB:
I don't know what is going on here how that file still has conflicts when so little is changed in it, but it does.

PR #6055, opened 2020-07-07
6527: Cleaned up MessageNode createSheet to remove dup values

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/relationships/MessageNode.java
```diff
@@ -89,42 +89,45 @@ protected Sheet createSheet() {
 
         final BlackboardArtifact artifact = getArtifact();
         BlackboardArtifact.ARTIFACT_TYPE fromID = BlackboardArtifact.ARTIFACT_TYPE.fromID(artifact.getArtifactTypeID());
```
raman-bt, line 91: I know this is not part of your change, but 'fromID" is a terrible name in this context to denote the "type id of the artifact from which to get the attributes. 

I've tripped on it multiple times.

PR #6057, opened 2020-07-08
6579: Input from user can cause SQL exceptions in Personas

raman-bt:
Modified persona classes to run SQL as prepared statements.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepository.java
+++ b/Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepository.java
@@ -64,7 +64,8 @@ public interface CentralRepository {
      * It will not close active/in-use connections. Thus, it is vital that there
      * are no in-use connections when you call this method.
      *
-     * @throws CentralRepoException if there is a problem closing the connection pool.
+     * @throws CentralRepoException if there is a problem closing the connection
+     * pool.
      */
     void shutdownConnections() throws CentralRepoException;
 
@@ -101,7 +102,7 @@ public interface CentralRepository {
     /**
      * Add a new name/value pair in the db_info table.
      *
-     * @param name  Key to set
+     * @param name Key to set
      * @param value Value to set
      *
      * @throws CentralRepoException
@@ -112,9 +113,9 @@ public interface CentralRepository {
      * Set the data source object id for a specific entry in the data_sources
      * table
      *
-     * @param rowId              - the row id for the data_sources table entry
+     * @param rowId - the row id for the data_sources table entry
      * @param dataSourceObjectId - the object id for the data source from the
-     *                           caseDb
+     * caseDb
      */
     void addDataSourceObjectId(int rowId, long dataSourceObjectId) throws CentralRepoException;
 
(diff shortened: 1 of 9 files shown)


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/RdbmsCentralRepo.java
```diff
@@ -2543,88 +2547,55 @@ public void processSelectClause(String selectClause, InstanceTableCallback insta
...
-        try (Connection conn = connect();
-                PreparedStatement preparedStatement = conn.prepareStatement(sqlSb.toString());) {
-            preparedStatement.executeUpdate();
-        } catch (SQLException ex) {
-            throw new CentralRepoException(String.format("Error running SQL %s, exception = %s", deleteSQL, ex.getMessage()), ex);
+            throw new CentralRepoException(String.format("Error executing prepared statement for SQL query %s", sql), ex);
+        } finally {
+            if (resultSet != null) {
+                CentralRepoDbUtil.closeResultSet(resultSet);
+            }
```
rcordovano, line 2598: @raman-bt and @gdicristofaro, I think we can hadnle closing of the ResultSet more cleanly using another try with resources.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/SqliteCentralRepo.java
```diff
@@ -832,48 +832,27 @@ public void processSelectClause(String selectClause, InstanceTableCallback insta
...
-        }
-    }
     
     @Override
-    public void executeSelectSQL(String selectSQL, CentralRepositoryDbQueryCallback queryCallback) throws CentralRepoException {
+    public void executeSqlAsPreparedStatement(String sql, List<Object> params) throws CentralRepoException {
         try {
             acquireSharedLock();
-            super.executeSelectSQL(selectSQL, queryCallback);
+            super.executeSqlAsPreparedStatement(sql, params);
```
rcordovano, line 840: @raman-bt and @gdicristofaro, this method is documented as executing an UPDATE/INSERT/DELETE statement. We need to acquire the exclusive (write) lock, not the shared (read) lock.

Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CentralRepository.java
```diff
@@ -827,78 +825,64 @@ CorrelationAttributeInstance getCorrelationAttributeInstance(CorrelationAttribut
...
+     *
+     * @param sql sql to execute.
+     * @param params List of query params to use, may be empty.
+     * @param queryCallback Query callback to handle the result of the query.
+     *
      * @throws CentralRepoException If there is an error.
      */
-    void executeDeleteSQL(String sql) throws CentralRepoException;
-    
+    void executeQueryAsPreparedStatement(String sql, List<Object> params, CentralRepositoryDbQueryCallback queryCallback) throws CentralRepoException;
```
rcordovano, line 856: Perhaps we want to call the new methods:

executeCommand
executeQuery

?

PR #6086, opened 2020-07-15
1490 & 1402 cvt move filter db calls to thread and clean up device list

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/Bundle.properties
```diff
@@ -14,7 +14,7 @@ FiltersPanel.endCheckBox.text=End:
 FiltersPanel.refreshButton.text=Refresh
 FiltersPanel.deviceRequiredLabel.text=Select at least one.
 FiltersPanel.accountTypeRequiredLabel.text=Select at least one.
-FiltersPanel.needsRefreshLabel.text=Displayed data is out of date. Press Refresh.
+FiltersPanel.needsRefreshLabel.text=Displayed data maybe out of date. Press Refresh to update.
```
raman-bt, line 17: I think this should be "may be" ?

PR #6097, opened 2020-07-17
6624: Incorrect SQL for inserting accounts in SQLite CentralRepo

raman-bt:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/RdbmsCentralRepo.java
```diff
@@ -1085,8 +1085,19 @@ public CentralRepoAccount getOrCreateAccount(CentralRepoAccountType crAccountTyp
...
 
-        String insertSQL = "INSERT INTO accounts (account_type_id, account_unique_identifier) "
-                + "VALUES (?, ?) " + getConflictClause();
+        // insert the account.  If there is a conflict, ignore it.
+        String insertSQL;
+        switch (CentralRepoDbManager.getSavedDbChoice().getDbPlatform()) {
+            case POSTGRESQL:
+                insertSQL =  "INSERT INTO accounts (account_type_id, account_unique_identifier) VALUES (?, ?) " + getConflictClause();  //NON-NLS
+                break;
+            case SQLITE:
```
kellykelly3, line 1094: @raman-bt This change should not be needed.  You do not need that switch case.  The overload of getConflictCause for each of the subclass should take care of this.  I checked before I made the change.

PR #6107, opened 2020-07-21
6590 refresh discovery filters when opened

wschaeferB:


APriestman:
Please fix the conflicts.

PR #6147, opened 2020-07-28
6634 add chromium based browsers to chrome module

markmckinnon:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Chrome.java
```diff
@@ -84,14 +85,25 @@
     private Content dataSource;
     private IngestJobContext context;
     
+    private static final Map<String, String> BROWSERS_MAP = ImmutableMap.<String, String>builder()
```
dannysmyda, line 88: Should the module named be changed from Chrome to Chromium?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Chrome.java
```diff
@@ -223,8 +241,7 @@ private void getHistory() {
                         RecentActivityExtracterModuleFactory.getModuleName(),
                         ((result.get("title").toString() != null) ? result.get("title").toString() : ""))); //NON-NLS
                 bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_PROG_NAME,
-                        RecentActivityExtracterModuleFactory.getModuleName(),
-                        NbBundle.getMessage(this.getClass(), "Chrome.moduleName")));
+                        RecentActivityExtracterModuleFactory.getModuleName(), browser));
```
dannysmyda, line 244: If these browser names are going to be displayed to the user, I think they should be put into a Bundle so they can be translated.

markmckinnon: Done

PR #6149, opened 2020-07-29
1498 make method to map account to contact book

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/ContactCache.java
```diff
@@ -0,0 +1,160 @@
...
+            List<BlackboardAttribute> contactAttributes = contactArtifact.getAttributes();
+            for(BlackboardAttribute attribute: contactAttributes) {
+                String typeName = attribute.getAttributeType().getTypeName();
+                
+                if(typeName.startsWith("TSK_EMAIL")
+                || typeName.startsWith("TSK_PHONE")
+                || typeName.startsWith("TSK_NAME")
+                || typeName.startsWith("TSK_ID")) {
+                    String accountID = attribute.getValueString();
+                    List<BlackboardArtifact> artifactList = acctMap.get(accountID);
```
dannysmyda, line 145: You may find acctMap.getOrDefault(accountID, new ArrayList<>()) useful here.

Core/src/org/sleuthkit/autopsy/communications/ContactCache.java
```diff
@@ -0,0 +1,160 @@
...
+                try {
+                    return  buildMap();
+                } catch (SQLException | TskCoreException ex) {
+                    logger.log(Level.WARNING, "Failed to build account to contact map", ex);
+                } 
+                return new HashMap<>();  // Return an empty map if there is an exception to avoid NPE and continual trying.
+            }
+        });
+
+        PropertyChangeListener ingestListener = pce -> {
```
dannysmyda, line 97: In its current state, switching between cases may return old artifacts. Could you invalidate the cache when a case is closed?

Core/src/org/sleuthkit/autopsy/communications/ContactCache.java
```diff
@@ -0,0 +1,160 @@
...
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * A singleton cache of the Contact artifacts for accounts. The map of account
+ * unique ids to list of contact artifacts is stored in a LoadingCache which
+ * expires after 10 of non-use. 
+ *
+ */
+final class ContactCache {
```
dannysmyda, line 50: I was originally going to post this on the story but I can't find it. But, this class seems to be filling a void in SleuthkitCase, @rcordovano would it make sense to add queries to be able to select artifacts by value and type? I'd imagine this functionality will be needed elsewhere.

kellykelly3: @dannysmyda & @rcordovano Brian originally for this story told me to put this class in communications.  It moves to guiutils in a later story.

Core/src/org/sleuthkit/autopsy/communications/ContactCache.java
kellykelly3: Good idea.

Core/src/org/sleuthkit/autopsy/communications/ContactCache.java
dannysmyda: Richard sent me a link to the story. Since the requirement is to keep these queries out of public API, we can just ignore this comment.

kellykelly3:
@dannysmyda & @rcordovano I have made the changes based on Danny's comments. IS this story PR all set or are more changes required?

dannysmyda:
@kellykelly3 Let me test it quickly and we can get it through. I don't have any other suggestions.

dannysmyda:
@kellykelly3 Actually, is there anything to test?

kellykelly3:
Sorry about that @dannysmyda I guess you test that it compile.  There is nothing specific to test.

PR #6157, opened 2020-08-04
6660 - HEIC Ingest Module

dannysmyda:


dannysmyda:
@markmckinnon I'm not going to do anything to address the EXIF codacy comments, this code was taken from an existing module.

Core/src/org/sleuthkit/autopsy/modules/pictureanalysis/impls/HEICProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+        final Path moduleOutputFolder = getModuleOutputFolder(heicFile);
+
+        final String baseFileName = FilenameUtils.getBaseName(FileUtil.escapeFileName(heicFile.getName()));
+        final Path outputFile = moduleOutputFolder.resolve(baseFileName + ".jpg");
+
+        // ImageMagick will write the primary image to the output file.
+        // Any additional images found within the HEIC container will be
+        // formatted as fileName-1.jpg, fileName-2.jpg, etc.
+        final ProcessBuilder processBuilder = new ProcessBuilder()
+                .command(IMAGE_MAGICK_PATH.toString(),
```
markmckinnon, line 186: Could you add a redirectError to processBuilder so that it will redirect the error to the temp directory for the file.  This will make debugging easier in the future when issues happen.

PR #6158, opened 2020-08-05
6654 summary pie chart

gdicristofaro:


Core/src/org/sleuthkit/autopsy/casemodule/datasourcesummary/DataSourceSummaryCountsPanel.java
```diff
@@ -96,29 +96,20 @@ public DataSource getDataSource() {
...
 
     /**
      * Specify the DataSource to display file information for.
      *
-     * @param mimeTypeDataModel     The mime type data model.
      * @param fileCategoryDataModel The file category data model.
      * @param artifactDataModel     The artifact type data model.
      */
-    private void updateCountsTableData(Object[][] mimeTypeDataModel, Object[][] fileCategoryDataModel, Object[][] artifactDataModel) {
-        fileCountsByMimeTypeTable.setModel(new NonEditableTableModel(mimeTypeDataModel, MIME_TYPE_COLUMN_HEADERS));
```
wschaeferB, line 118: The MIME_TYPE_COLUMN_HEADERS is no longer used, and should be removed.

Core/src/org/sleuthkit/autopsy/casemodule/datasourcesummary/DataSourceSummaryCountsPanel.java
```diff
@@ -39,7 +38,8 @@
     "DataSourceSummaryCountsPanel.FilesByCategoryTableModel.count.header=Count"
 })
 class DataSourceSummaryCountsPanel extends javax.swing.JPanel {
```
wschaeferB, line 40: we can make this class final to prevent the setDataSource(null) from complaining about being overridable as well as remove the logger I will add these changes to PR #6191

PR #6162, opened 2020-08-06
6687 add tooltip long hash name

kellykelly3:


kellykelly3:
This PR includes files it shouldn't.  I am going to fix my branch and then make a new PR

PR #6165, opened 2020-08-07
new URL's mapping

dcharit:


dcharit:
Per Brian's request, the XML was updated to reflect needed changes.

PR #6171, opened 2020-08-11
6682 - Added support for new command -runUI

kellykelly3:


Core/src/org/sleuthkit/autopsy/commandlineingest/CommandLineOptionProcessor.java
```diff
@@ -52,8 +52,10 @@
     private final Option ingestProfileOption = Option.requiredArgument('p', "ingestProfile");
     private final Option listAllDataSourcesCommandOption = Option.withoutArgument('l', "listAllDataSources");
     private final Option generateReportsOption = Option.withoutArgument('g', "generateReports");
+    private final Option runUICommandOption = Option.requiredArgument('u', "runUI");
```
eugene7646, line 55: Seems like this should be "withoutArgument" instead?

Core/src/org/sleuthkit/autopsy/report/modules/portablecase/PortableCaseReportModule.java
```diff
@@ -1142,6 +1154,76 @@ private String getExportSubfolder(AbstractFile abstractFile) {
...
+    
+    /**
+     * Create a bat file at destBaseFolder that will launch the portable case.
+     * 
+     * @param destBaseFolder
+     * @throws IOException 
+     */
+    private void createAppLaunchBatFile(String destBaseFolder) throws IOException {
+        Path filePath = Paths.get(destBaseFolder, "open.bat");
+        String exePath = "\"%~dp0autopsy\\bin\\" + getAutopsyExeName() + "\"";
```
eugene7646, line 1221: Will this algorithm work if we are running Viking? I haven't tested that.

kellykelly3:
Closing PR so there is no confusion about the current state of the code.

PR #6174, opened 2020-08-12
6682 autopsy in portable case

kellykelly3:


kellykelly3:
I have addressed some of the codacy issues, but not all.  The use of System.err was consistent with existing code.  I didn't change compressCase much but I didn't not affect the complexity of the existing method.  I looked at it for a bit and tried to decide how it could be broken down and I kind of liked the method the way it is.

PR #6177, opened 2020-08-13
6689 - Added command line support for using report profiles

kellykelly3:


kellykelly3:
The one codacy issue about the use of the System.err is consistent with the reset of the code and therefore can be ignored.

Core/src/org/sleuthkit/autopsy/report/infrastructure/ReportingConfigLoader.java
```diff
@@ -219,5 +221,27 @@ static synchronized void saveConfig(ReportingConfig reportConfig) throws ReportC
...
     }
+    
+    /**
+     * Return a list of the names of the report profiles in the
+     * REPORT_CONFIG_FOLDER_PATH.
+     *
+     * @return Naturally ordered list of report profile names. If none were found
+     *         the list will be empty.
+     */
+    static synchronized Set<String> getListOfReportConfigs() {
```
eugene7646, line 232: I'm wondering why does this method have to be synchronized. Or mostly I'm wondering why is loadConfig(String configName) in the same class synchronized? Was the original intent to make sure that we don't read the configuration from disk, while it is being written to disk in saveConfig(ReportingConfig reportConfig) method? If so, then obviously synchronizing on the methods will not work, we need to have a separate lock. @rcordovano thoughts?

eugene7646: Ooooohhhhhh, nevermind, these are static methods, so they lock the class, not just the methods. Whoops!

eugene7646:
@kellykelly3 Code looks good. The only minor note is to update the copyright text to 2020.

kellykelly3:
@rcordovano merge conflicts are fixed.

PR #6178, opened 2020-08-17
6669 use local drive

gdicristofaro:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1352,7 +1328,15 @@ public String getOutputDirectory() {
      * @return The temp subdirectory path.
      */
     public String getTempDirectory() {
-        return getOrCreateSubdirectory(TEMP_FOLDER);
+        // get temp folder scoped to the uuid of the case
```
APriestman, line 1331: Is this comment accurate?

gdicristofaro: I think `getName()` is a unique identifier.  I believe it is the case name and a time stamp.

APriestman: Yes it's the case name + timestamp, but I don't exactly consider that a UUID? Especially since you use an actual UUID in one of the other temp paths.

gdicristofaro: Sure.  Change it from 'uuid' to 'unique identifier'?

APriestman: I think just "case name and timestamp" so there's no confusion

gdicristofaro: Ok.  Should be pushed up now.

PR #6181, opened 2020-08-17
6659 - Upgrade Tika to 1.24.1

dannysmyda:


KeywordSearch/ivy.xml
```diff
@@ -46,7 +38,11 @@
         <dependency conf="slf4j-libs->default" org="org.slf4j" name="jcl-over-slf4j" rev="1.7.10"/>
         <dependency conf="slf4j-libs->default" org="org.slf4j" name="jul-to-slf4j" rev="1.7.10"/>
         
+        <!-- Remnants of a past Tika dependency. These jars were being exported from Tika, which has since
```
esaunders, line 41: Are you sure about this? KWS talks to Solr via the SolrJ HttpSolrServer class which in turn uses the Apache HTTP client.

dannysmyda: Yes, and it does declare it as a dep in Maven so Ivy pulls it down. I need clarify this in the comment, but the ones pulled down for SolrJ are much earlier editions. When Tika was here, it needed later versions which were in turn exported. Those newer versions added packages that don't exist in the ones Solr pulls down. To keep this PR to only removing Tika stuff from the public exports of KWS and Core, I decided to explicitly pull down the newer versions.

dannysmyda:
@esaunders Yes they do, check the build-impl.xml (https://github.com/sleuthkit/sleuthkit/blob/develop/case-uco/java/nbproject/build-impl.xml) for that project. It's way more fleshed out then our existing projects. You can Ctrl-F for -pre-compile.

PR #6203, opened 2020-08-21
6732 - Added getArtifact method to DataSourceInfoUtilities

kellykelly3:


Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/DataSourceInfoUtilities.java
```diff
@@ -133,9 +139,126 @@ static String getMetaFlagsContainsStatement(TSK_FS_META_FLAG_ENUM flag) {
...
+     *         found an empty list will be returned.
+     *
+     * @throws TskCoreException
+     */
+    static List<BlackboardArtifact> getArtifacts(SleuthkitCase skCase, BlackboardArtifact.Type artifactType, DataSource dataSource, BlackboardAttribute.Type attributeType, SortOrder sortOrder, int maxCount) throws TskCoreException {
+        if (maxCount < 1 && maxCount != -1) {
+            throw new IllegalArgumentException("Invalid maxCount passed to getArtifacts, value must be at greater 0");
+        }
+
+        TreeMap<BlackboardAttribute, BlackboardArtifact> sortedMap = new TreeMap<>(new AttributeComparator(sortOrder));
```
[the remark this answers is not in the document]
kellykelly3: If an artifact does not have the attribute being sorted on it is not included.  Why would artifacts that doesn't have the attribute be of interest?  

You are correct, the TreeMap doesn't allow for multiple of the same key.  I will fix that.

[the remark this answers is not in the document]
gdicristofaro: Ok.  I think it would be good to document that behavior in the comment.  I initially thought it was going to behave comparable to a sql sort where rows with null for the sort value are still included.  It would be good to update the comment for the 'maxCount' parameter for the same reason.

kellykelly3:
@rcordovano & @gdicristofaro updated with fix and requested change.

Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/DataSourceInfoUtilities.java
```diff
@@ -20,13 +20,21 @@
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
```
gdicristofaro, line 26: I think this is an unused import causing a codacy issue.

Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/DataSourceInfoUtilities.java
```diff
@@ -133,9 +141,204 @@ static String getMetaFlagsContainsStatement(TSK_FS_META_FLAG_ENUM flag) {
...
+     *
+     * @return A list of artifacts of type artifactType sorted by the attribute
+     *         of attributeType in the given sortOrder. If no artifacts are
+     *         found an empty list will be returned.
+     *
+     * @throws TskCoreException
+     */
+    static List<BlackboardArtifact> getArtifacts(SleuthkitCase skCase, BlackboardArtifact.Type artifactType, DataSource dataSource, BlackboardAttribute.Type attributeType, SortOrder sortOrder, int maxCount) throws TskCoreException {
+        if (maxCount < 0) {
+            throw new IllegalArgumentException("Invalid maxCount passed to getArtifacts, value must be at greater 0");
```
gdicristofaro, line 200: I think this is supposed to be "Invalid maxCount passed to getArtifacts; value must be greater than or equal to 0."

Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/DataSourceInfoUtilities.java
```diff
@@ -133,9 +141,204 @@ static String getMetaFlagsContainsStatement(TSK_FS_META_FLAG_ENUM flag) {
...
+     * @param maxCount  Maximum number of artifacts to return.
+     *
+     * @return List of artifacts, list will be empty if none were found.
+     */
+    static private List<BlackboardArtifact> createListFromMap(SortedMap<BlackboardAttribute, List<BlackboardArtifact>> sortedMap, int maxCount) {
+        List<BlackboardArtifact> artifactList = new ArrayList<>();
+
+        for (List<BlackboardArtifact> mapArtifactList : sortedMap.values()) {
+
+            if (maxCount == artifactList.size()) {
```
gdicristofaro, line 263: I believe the intent is that if maxCount is zero, return all records.  It looks like this line causes no records to be returned in that event.

PR #6204, opened 2020-08-24
6657 - Removed the use of group layout from cvt visualization pane

kellykelly3:


Core/src/org/sleuthkit/autopsy/communications/Bundle_ja.properties
```diff
@@ -56,7 +56,6 @@ CVTTopComponent.accountsBrowser.TabConstraints.tabTitle_1=\u53c2\u7167
 CVTTopComponent.browseVisualizeTabPane.AccessibleContext.accessibleName=\u53ef\u8996\u5316
 CVTTopComponent.vizPanel.TabConstraints.tabTitle_1=\u53ef\u8996\u5316
 VisualizationPanel.fitGraphButton.text=
-VisualizationPanel.jTextArea1.text=[\u30a2\u30ab\u30a6\u30f3\u30c8\u53c2\u7167] \u30c6\u30fc\u30d6\u30eb\u3067\u30a2\u30ab\u30a6\u30f3\u30c8\u3092\u53f3\u30af\u30ea\u30c3\u30af\u3057\u3001[\u53ef\u8996\u5316] \u3092\u9078\u629e\u3057\u3066\u958b\u59cb\u3057\u307e\u3059\u3002
```
esaunders, line 59: Given that the text hasn't changed, is it simply a matter of changing jTextArea1 to jTextPane1 to keep the existing Japanese translation?

kellykelly3: Maybe?

esaunders: It would seem to me that, given the text hasn't changed,  keeping the existing Japanese translation would be a good thing so that it doesn't have to go through the translation process again down the road.
@rcordovano What's your take on this?

rcordovano: We should preserve the translation by changing the syncing the property keys, if I understand the issue correctly.

APriestman: The translation script looks for all new keys and keys where the English text has changed. It does not currently copy the existing Japanese text into the spreadsheet (it's just an empty box to fill in for every new/changed key), so I don't believe manually moving the translation will save any work.

kellykelly3: How about I do this.  I can change jTextPane1 to jTextArea1 (even though its a pane).  Then I will put the original translation back?

esaunders: I could live with that.

rcordovano: OK

PR #6230, opened 2020-09-03
6761 Don't display device ID under accounts.

APriestman:
Fix bug related to call logs with no direction.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/contentviewers/artifactviewers/CallLogArtifactViewer.java
+++ b/Core/src/org/sleuthkit/autopsy/contentviewers/artifactviewers/CallLogArtifactViewer.java
@@ -171,17 +171,17 @@ public class CallLogArtifactViewer extends javax.swing.JPanel implements Artifac
                 fromAccountAttr = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM));
                 localAccountAttr = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM));
             }
-        }
-
-        // if direction isn't known, check all the usual attributes that may have the number/address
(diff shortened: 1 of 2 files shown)


Core/src/org/sleuthkit/autopsy/contentviewers/artifactviewers/CallLogArtifactViewer.java
```diff
@@ -171,17 +171,17 @@ private CallLogViewData getCallLogViewData(BlackboardArtifact artifact) throws T
...
+            // in the absence of sufficent data, any number available will be displayed as a From address.
+            if (fromAccountAttr == null) {
+                fromAccountAttr = ObjectUtils.firstNonNull(
+                        artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM)),
+                        artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO)),
+                        artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER)),
+                        artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ID))
+                );
+            }
         }
```
kellykelly3, line 185: Should we be getting all attributes and looking for any that start with TSK_PHONE_* ? There is some Perl module that parses the db of phone numbers which could cause, for example, TSK_PHONE_HOME , TSK_PHONE_OFFICE.

APriestman: We could. I'm not sure those would ever realistically show up in a call log but it might be good to check for that case.

kellykelly3: Oh you are correct, that is probably more of a contact thing.  Strike my comment.  :-)

kellykelly3 (approved):

PR #6242, opened 2020-09-08
6713 implement domain discovery

dannysmyda:


Core/src/org/sleuthkit/autopsy/discovery/search/DomainSearchCacheLoader.java
```diff
@@ -0,0 +1,282 @@
...
+                havingClause.toString()
+        );
+    }
+    
+    /**
+     * Callback to handle the result set of the domain query. This callback
+     * is responsible for mapping result set rows into ResultDomain objects
+     * for display.
+     */
+    private class DomainCallback implements CaseDbAccessQueryCallback {
```
wschaeferB, line 223: access level of constructor and some methods in this class do not match, it looks like they should all be private I think

wschaeferB (approved):

PR #6243, opened 2020-09-08
6777 types summary tab

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datasourcesummary/ui/DataSourceSummaryTabbedPane.java
```diff
@@ -33,25 +33,85 @@
...
+            this.onDataSource = onDataSource;
+        }
+
+        /**
+         * Main constructor.
+         *
+         * @param tabTitle  The title of the tab.
+         * @param component The component to be displayed.
+         *                  BaseDataSourceSummaryPanel.setDataSource method is
+         *                  used on new data source.
```
APriestman, line 73: What is this comment talking about?

gdicristofaro: A `BaseDataSourceSummaryPanel` has a method 'setDataSource' that should be called when a new data source should be displayed in the tabbed pane.  I can rephrase.

APriestman (approved):

PR #6245, opened 2020-09-08
6778 analysis summary tab

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/DataSourceAnalysisSummary.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Gets counts for hashset hits.
+     *
+     * @param dataSource The datasource for which to identify hashset hits.
+     *
+     * @return The hashset set name with the number of hits in descending order.
+     *
+     * @throws
+     * org.sleuthkit.autopsy.datasourcesummary.datamodel.SleuthkitCaseProvider.SleuthkitCaseProviderException
```
APriestman, line 81: Remove path? (There are several of these)

gdicristofaro: Sure.  So just `@throws SleuthkitCaseProviderException`?

APriestman (approved):

PR #6257, opened 2020-09-14
6768 Extract role from URL

APriestman:
Depends on https://github.com/sleuthkit/sleuthkit/pull/2022


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractWebAccountType.java
```diff
@@ -0,0 +1,369 @@
...
+        if (urlAttr == null) {
+            return;
+        }
+
+        BlackboardAttribute domainAttr = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DOMAIN));
+        if (domainAttr == null) {
+            return;
+        }
+
+        String url = urlAttr.getValueString();
```
dannysmyda, line 98: Could this be lower-cased since the contains checks are looking for lower-cased strings? Unless there is a reason this must be case sensitive.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractWebAccountType.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Extract myBB role.
+     * 
+     * @param url       The full URL.
+     * @param domain    The domain.
+     * @param artifact  The original artifact.
+     * @param roleProcessor Object to collect and process domain roles.
+     */    
+    private void findMyBbRole(String url, String domain, BlackboardArtifact artifact, RoleProcessor roleProcessor) {    
+        String platformName = "myBB platform"; // NON-NLS
```
dannysmyda, line 116: Very minor point, but should the platform suffix be a bundle message? Or is that part of the proper noun? Same for all instances of platformName variable in the various methods.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractWebAccountType.java
```diff
@@ -0,0 +1,369 @@
...
+                logger.log(Level.SEVERE, "Error creating web accounts", ex);
+            }
+        }
+    }
+    
+    /**
+     * Possible roles with rank and display name.
+     */
+    private enum Role {
+        USER("User", 0),
```
dannysmyda, line 278: I think these should be bundle messages.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractWebAccountType.java
APriestman: I don't think so, since the combination of name + translated "platform" may not make sense in other languages due to word ordering

dannysmyda: Oh right, good point.

dannysmyda (approved):

PR #6258, opened 2020-09-14
6838 Don't store localhost or 127.0.0.1 in CR

APriestman:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -148,8 +150,10 @@ private static String getEmailAddressAttrDisplayName() {
                         || artifactTypeID == ARTIFACT_TYPE.TSK_WEB_COOKIE.getTypeID()
                         || artifactTypeID == ARTIFACT_TYPE.TSK_WEB_DOWNLOAD.getTypeID()
                         || artifactTypeID == ARTIFACT_TYPE.TSK_WEB_HISTORY.getTypeID()) {
-                    makeCorrAttrFromArtifactAttr(correlationAttrs, sourceArtifact, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DOMAIN, CorrelationAttributeInstance.DOMAIN_TYPE_ID);
-
+                    String domain = sourceArtifact.getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_DOMAIN)).getValueString();
+                    if (! domainsToSkip.contains(domain)) {
+                        makeCorrAttrFromArtifactAttr(correlationAttrs, sourceArtifact, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DOMAIN, CorrelationAttributeInstance.DOMAIN_TYPE_ID);
+                    }
```
rcordovano, line 156: makeCorrAttrFromArtifactAttr() handles the condition where there is no TSK_DOMAIN attribute attached to the artifact, but we lose that with this change. BlackboardArtifact.getAttribute() can return null, so there is a potential NPE here now.

rcordovano (changes requested):

rcordovano (approved):

PR #6276, opened 2020-09-16
6847 Hook up Domain Discovery UI to Thumbnail API

dannysmyda:


Core/src/org/sleuthkit/autopsy/discovery/search/DomainSearchThumbnailLoader.java
```diff
@@ -24,6 +24,7 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import org.sleuthkit.autopsy.coreutils.ImageUtils;
+import org.openide.util.ImageUtilities;
```
wschaeferB, line 27: unused import

wschaeferB (approved):

PR #6277, opened 2020-09-16
6808 replace all exif with pa

kellykelly3:


Core/src/org/sleuthkit/autopsy/modules/pictureanalyzer/impls/EXIFProcessor.java
```diff
@@ -69,12 +70,11 @@
 
     @Override
     @NbBundle.Messages({
-        "ExifProcessor.indexError.message=Failed to post EXIF Metadata artifact(s).",
-        "ExifProcessor.userContent.description=EXIF metadata exists for this file.",
-        "ExifProcessor.module_name=Exif Parser"
+        "ExifProcessor.indexError.message=Failed to post Picture Analzyer artifact(s).",
+        "ExifProcessor.userContent.description=Picture Analyzer data exists for this file."
```
rcordovano, line 74: These two seem like they shouldn't change because I think they are referring to EXIF artifacts/data rather than to the ingest module.

rcordovano (changes requested):

rcordovano (approved):

PR #6330, opened 2020-09-23
6886 - Discovery UI Background Task Cancellation

dannysmyda:
This PR builds off of https://github.com/sleuthkit/autopsy/pull/6328. The last commit include the relevant changes to review.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/discovery/search/DiscoveryEventUtils.java
+++ b/Core/src/org/sleuthkit/autopsy/discovery/search/DiscoveryEventUtils.java
@@ -95,6 +95,9 @@ public final class DiscoveryEventUtils {
      */
     public static final class CancelBackgroundTasksEvent {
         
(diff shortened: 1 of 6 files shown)


Core/src/org/sleuthkit/autopsy/discovery/search/DiscoveryEventUtils.java
```diff
@@ -88,6 +88,14 @@ public ClearInstanceSelectionEvent() {
             //no arg constructor
         }
     }
+    
+    /**
+     * Event to signal that any background tasks currently running should
+     * be cancelled.
+     */
+    public static final class CancelBackgroundTasksEvent {
+
```
wschaeferB, line 97: Can we add an empty no arg constructor similar to the ClearInstanceSelectionEvent for this so that netbeans doesn't think it can improve things and tell us 'this may be an interface'?

wschaeferB (approved):

PR #6383, opened 2020-10-09
6945 Enable OCR in translation tab (Release)

APriestman:


APriestman:
Included in https://github.com/sleuthkit/autopsy/pull/6386

PR #6393, opened 2020-10-15
6951 - Fix for Nightly TIFF failures

dannysmyda:


Core/src/org/sleuthkit/autopsy/ingest/IngestManager.java
```diff
@@ -887,6 +897,44 @@ long getFreeDiskSpace() {
...
+    private <T> void sortPluginsInCategory(IIORegistry pluginRegistry, Class<T> category) {
+        Iterator<T> serviceProviderIter = pluginRegistry.getServiceProviders(category, false);
+        ArrayList<T> providers = new ArrayList<>();
+        while (serviceProviderIter.hasNext()) {
+            providers.add(serviceProviderIter.next());
+        }
+        Collections.sort(providers, (first, second) -> {
+            return first.getClass().getCanonicalName().compareToIgnoreCase(second.getClass().getCanonicalName());
+        });
+        for(int i = 0; i < providers.size(); i++) {
```
APriestman, line 928: Could change to size - 1

Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/EmbeddedFileExtractorIngestModule.java
```diff
@@ -122,8 +127,53 @@ public void startUp(IngestJobContext context) throws IngestModuleException {
         } catch (NoCurrentCaseException ex) {
             throw new IngestModuleException(Bundle.EmbeddedFileExtractorIngestModule_UnableToGetMSOfficeExtractor_errMsg(), ex);
         }
+        
+        /**
+         * Initialize Java's Image I/O API so that image reading and writing 
+         * (needed for image extraction) happens consistently through the 
+         * same providers. See JIRA-6951 for more details.
+         */
+        initializeImageIO();
```
rcordovano, line 136: We don't need to do this on every thread, i.e., in every ingest module instance startUp() call in an ingest job, do we? In fact, we should only need to do this once per application session, right?

rcordovano (changes requested):

Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/EmbeddedFileExtractorIngestModule.java
dannysmyda: I don't think it hurts. Besides, being a global registry means any libraries can rearrange ordering. For that reason, it'd be most correct to do it before every module instance run.

APriestman (approved):

Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/EmbeddedFileExtractorIngestModule.java
rcordovano: Assuming there is no significant performance cost, then no, it shouldn't hurt. There is certainly no need for premature optimization, so I will withdraw my objection. 

For the sake of our understanding, the registry is not global to the application, it's global to the thread group, right? I now that we are not explicitly creating the ThreadGroup that the file ingest threads belong to and the doc at https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html says this about Executors.defaultThreadFactory(): "This factory creates all new threads used by an Executor in the same ThreadGroup" while a comment in the code for com.google.common.util.concurrent.ThreadFactoryBuilder indicates that this is the ThreadGroup we are getting because we do not supply one: "If no backing thread factory is provided, a default backing thread factory is used as if by calling {@code setThreadFactory(}{@linkExecutors#defaultThreadFactory()}{@code )}." I conclude from this that the meddlers would have to be ingest modules that did their meddling after the EFE module instances all execute their startUp() methods, right?

dannysmyda: Yes, I withdraw my global comment. It would be local to the ingest modules since they're in the same thread group. And yes, exactly.

PR #6395, opened 2020-10-16
Java11 timeline filter tree fix

kellykelly3:


Core/src/org/sleuthkit/autopsy/timeline/ui/AbstractFXCellFactory.java
```diff
@@ -64,6 +65,8 @@ protected void updateItem(Y item, boolean empty) {
         @Override
         protected void updateItem(Y item, boolean empty) {
             super.updateItem(item, empty);
+            TreeTableView<X> treeTableView = this.treeTableViewProperty().get();
+            this.setTranslateX(treeTableView.getTreeItemLevel(treeTableView.getTreeItem(getIndex())) << 4);
```
esaunders, line 69: This (and the CSS additions below) could probably do with a comment explaining what we are doing and why we are doing it (possibly with a link to the JavaFX bug).

esaunders (approved):

PR #6404, opened 2020-10-21
6949 STIX Report Module Errors

dannysmyda:


esaunders (commented):
Rather than having to repeat the ContextClassLoader code in each of these static methods, would it be a worthwhile trade-off to statically initialize all of the necessary factory objects? 
@dannysmyda Thoughts?

dannysmyda:
@esaunders Yeah, it would be cleaner. I'll do the refactor.

PR #6408, opened 2020-10-23
6774 6780 downloads and cache discovery details viewer

wschaeferB:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -284,17 +284,50 @@ public BlackboardArtifactNode(BlackboardArtifact artifact) {
...
+            return Lookups.fixed(artifact, content);
+        }
+
+    }
+
+    /**
+     * Private helper method to allow content specified in a path id attribute
+     * to be retrieved.
+     *
+     * @param artifact The artifact which content may be specified as a tsk path
```
rcordovano, line 311: "which content" => "for which content"

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -284,17 +284,50 @@ public BlackboardArtifactNode(BlackboardArtifact artifact) {
...
+     */
+    private static Content getPathIdFile(BlackboardArtifact artifact) throws ExecutionException {
+        try {
+            for (BlackboardAttribute attribute : artifact.getAttributes()) {
+                if (attribute.getAttributeType().getTypeID() == BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PATH_ID.getTypeID()) {
+                    return contentCache.get(attribute.getValueLong(), () -> artifact.getSleuthkitCase().getContentById(attribute.getValueLong()));
+                }
+            }
+        } catch (TskCoreException ex) {
+            logger.log(Level.WARNING, MessageFormat.format("Error getting content for path id attrbiute for artifact: ", artifact.getId()), ex); //NON-NLS
```
rcordovano, line 328: I think that we can simplify this code slightly by taking advantage of this BlackboardArtifact method:
BlackboardAttribute getAttribute(BlackboardAttribute.Type attributeType).

rcordovano (changes requested):

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
wschaeferB: Thank you.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
wschaeferB: Good catch. Thank you.

rcordovano (approved):

PR #6443, opened 2020-11-05
Updated deprecated Solr8 schema fields (6990)

eugene7646:
* Updated deprecated schema fields. 
* Improved check for existing collections


eugene7646:
@rcordovano can you please take this PR into solr8 branch? I would like to have it before I build a new Solr8 installer.

rcordovano (approved):

PR #6453, opened 2020-11-11
6910 Initial file repository work

APriestman:


Core/src/org/sleuthkit/autopsy/filerepository/Bundle.properties
```diff
@@ -0,0 +1,9 @@
+OptionsCategory_Name_File_Repository_Options=File Repository
+OptionsCategory_Keywords_File_Repository_Options=File Repository Settings
+FileRepositoryOptionsPanel.addressLabel.text=Hostname or IP Address
+FileRepositoryOptionsPanel.addressTextField.text=
+FileRepositoryOptionsPanel.portLabel.text=Port
+FileRepositoryOptionsPanel.portTextField.text=
+FileRepositoryOptionsPanel.headingLabel.text=Configure the settings for the file repository. Settings can not be changed while ingest is in progress.
```
dannysmyda, line 7: Given that the panel is disabled while the case is open, can this message be updated again (sorry) to mention that?

APriestman: Ok wow yeah that is just incorrect.

dannysmyda (approved):

PR #6470, opened 2020-11-16
6989 - Added new KeywordSearch ingest OCR process terminator

kellykelly3:


KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/KeywordSearchIngestModule.java
```diff
@@ -807,4 +814,27 @@ private boolean indexTextFile(AbstractFile aFile) {
             return false;
         }
     }
+    
+    /**
+     * This class takes a list of ProcessTerminators checking all of them
+     * during shouldTerminateProcess.
+     */
+    private class HybridTerminator implements ProcessTerminator {
```
dannysmyda, line 822: Very nice use of composition :)

kellykelly3: It might make sense to move this to the ProcessTerminator for better reuse, but I was keeping the scope small for this story.

dannysmyda (approved):

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/KeywordSearchIngestModule.java
```diff
@@ -515,7 +518,11 @@ private void postIndexSummary() {
         private boolean extractTextAndIndex(AbstractFile aFile, Map<String, String> extractedMetadata) throws IngesterException {
             ImageConfig imageConfig = new ImageConfig();
             imageConfig.setOCREnabled(KeywordSearchSettings.getOcrOption());
-            ProcessTerminator terminator = () -> context.fileIngestIsCancelled();
+            
+            List<ProcessTerminator> terminators = new ArrayList<>();
+            terminators.add(() -> context.fileIngestIsCancelled()); 
+            terminators.add(new TimedProcessTerminator(OCR_TIMEOUT_SECONDS));
+            HybridTerminator terminator = new HybridTerminator(terminators);
             Lookup extractionContext = Lookups.fixed(imageConfig, terminator);
```
dannysmyda, line 526: This will actually cancel any extractor that is listening to the terminator, instead of just the OCR task.
Could you instead add the TimedProcessTerminator (and the 30 minute constant) to the ImageConfig class instead?
You could then construct a HybridTerminator before the OCR task is started in TikaTextExtractor.

dannysmyda (changes requested):

dannysmyda (approved):

PR #6479, opened 2020-11-20
7034 data source nav

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datasourcesummary/ui/DataSourceSummaryTabbedPane.java
```diff
@@ -138,10 +132,46 @@ public Runnable getOnClose() {
...
+     * Sets the listener for parent close events.
+     *
+     * @param parentCloseAction The observer.
+     */
+    void setParentCloseListener(Runnable parentCloseAction) {
+        notifyParentClose = parentCloseAction;
+    }
+
+    private List<DataSourceTab> getTabs() {
+        List<DataSourceTab> tabs = Arrays.asList(
```
wschaeferB, line 159: If this method is intended to improve readability or for some other reason it is certainly fine, but we should avoid hiding the tabs field with a local variable named tabs.

Core/src/org/sleuthkit/autopsy/datasourcesummary/ui/Bundle.properties-MERGED
```diff
@@ -103,6 +103,17 @@ PastCasesPanel.notableFileLabel.text=Cases with Common Items That Were Tagged as
...
+RecentFilesPanel.rightClickForMoreOptions2.text=Right click on row for more options
+RecentFilesPanel.rightClickForMoreOptions3.text=Right click on row for more options
+UserActivityPanel.rightClickForMoreOptions1.text=Right click on row for more options
+UserActivityPanel.rightClickForMoreOptions2.text=Right click on row for more options
+UserActivityPanel.rightClickForMoreOptions3.text=Right click on row for more options
+UserActivityPanel.rightClickForMoreOptions4.text=Right click on row for more options
+UserActivityPanel.rightClickForMoreOptions5.text=Right click on row for more options
+TimelinePanel.viewInTimelineBtn.text=View in Timeline
+UserActivityPanel_goToArtifact=Go to Artifact
+UserActivityPanel_goToFile=Go to File
```
wschaeferB, line 116: We want to use the term Result in place of Artifact in text the end user will read.

Also we may want to consider the following phrasing instead for consistency with the rest of the application.
"View Source Result" and
"View Source File in Directory"

wschaeferB (approved):

PR #6493, opened 2020-11-24
6984 Add support for GPS area artifacts

APriestman:


Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
```diff
@@ -903,4 +943,66 @@ public void paint(Graphics2D g, JXMapViewer map, int w, int h) {
...
+     */
+    private class MapAreaRenderer implements Painter<JXMapViewer> {
+
+        private final List<Set<MapWaypoint>> areas;
+
+        MapAreaRenderer(List<Set<MapWaypoint>> areas) {
+            this.areas = areas;
+        }
+
+        private void drawArea(Set<MapWaypoint> area, Graphics2D g, JXMapViewer map) {
```
kellykelly3, line 958: Missing method header.

Core/src/org/sleuthkit/autopsy/geolocation/MapPanel.java
```diff
@@ -903,4 +943,66 @@ public void paint(Graphics2D g, JXMapViewer map, int w, int h) {
...
+                    polygon.moveTo(thisX, thisY);
+                    first = false;
+                } else {
+                    polygon.lineTo(thisX, thisY);
+                }
+            }
+            polygon.closePath();
+            
+            Color areaColor = area.iterator().next().getColor();
+            g.setPaint(new Color((float)(areaColor.getRed() / 255.0),
```
kellykelly3, line 981: Put 255.0 into constant.

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Area.java
```diff
@@ -0,0 +1,177 @@
...
+import org.sleuthkit.datamodel.blackboardutils.attributes.BlackboardJsonAttrUtil;
+import org.sleuthkit.datamodel.blackboardutils.attributes.BlackboardJsonAttrUtil.InvalidJsonException;
+import org.sleuthkit.datamodel.blackboardutils.attributes.GeoAreaPoints;
+import org.sleuthkit.autopsy.coreutils.Logger;
+
+/**
+ * A GPS track with which wraps the TSK_GPS_AREA artifact.
+ */
+public final class Area extends GeoPath {
+    private static final Logger LOGGER = Logger.getLogger(Track.class.getName());
```
kellykelly3, line 38: Should the class for the logger be Track or Area?

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Area.java
```diff
@@ -0,0 +1,177 @@
...
+     * @param attributeMap Map of artifact attributes.
+     *
+     * @return GeoTrackPoint list empty list if the attribute was not found.
+     *
+     * @throws GeoLocationDataException
+     */
+    private GeoAreaPoints getPointsList(Map<BlackboardAttribute.ATTRIBUTE_TYPE, BlackboardAttribute> attributeMap) throws GeoLocationDataException {
+        BlackboardAttribute attribute = attributeMap.get(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_GEO_AREAPOINTS);
+        if (attribute == null) {
+            LOGGER.log(Level.SEVERE, "No TSK_GEO_AREAPOINTS attribute was present on the artifact.");
```
kellykelly3, line 111: Do we need to log the error and throw an exception?  Will the calling code write an error too?

kellykelly3 (changes requested):

Core/src/org/sleuthkit/autopsy/geolocation/datamodel/Area.java
APriestman: So that's copied from the Track class, and the error actually isn't logged in the one place I see it being caught in GeoPath.
} catch (GeoLocationDataException e) {
   allParsedSuccessfully = false;
}
But I think it is better behavior to throw the exception and then log it up there where we have access to the artifactID, so I'll update my code and the Track code.

kellykelly3 (approved):

PR #6494, opened 2020-11-24
7002 domain names

gdicristofaro:


Core/src/org/sleuthkit/autopsy/coreutils/DomainCategorizer.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Loads a DomainCategorizer instance using the public suffix list.
+     *
+     * @return The DomainCategorizer instance.
+     * @throws IOException If there is an error reading the file.
+     */
+    private static DomainCategorizer load() throws IOException {
+        InputStream is = DomainCategorizer.class.getResourceAsStream(DOMAIN_LIST);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(is));
```
dannysmyda, line 92: Reader should be closed.

Core/src/org/sleuthkit/autopsy/coreutils/public_suffix_list.dat
```diff
@@ -0,0 +1,13494 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
dannysmyda, line 1: This file will become public API. Could we move it someplace else?

dannysmyda (changes requested):

Core/src/org/sleuthkit/autopsy/coreutils/DomainCategorizer.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Attempts to get the domain from a url/domain provided removing the
+ * subdomain(s).
+ */
+class DomainCategorizer {
```
bcarrier, line 35: This isn't really categorization and will be confusing with some of our other uses of categorization. We should call this something like DomainTokenizer.

Core/src/org/sleuthkit/autopsy/coreutils/DomainCategorizer.java
gdicristofaro: Will do.

Core/src/org/sleuthkit/autopsy/coreutils/DomainCategorizer.java
gdicristofaro: Sounds good.

Core/src/org/sleuthkit/autopsy/coreutils/NetworkUtils.java
```diff
@@ -18,21 +18,25 @@
  */
 package org.sleuthkit.autopsy.coreutils;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.UnknownHostException;
-import java.util.StringTokenizer;
+import java.util.logging.Level;
```
dannysmyda, line 25: Oh, just noticed this. This is not the right logger, could be the coreutils one.

dannysmyda: This is the logging level, not logger. My mistake.

dannysmyda (approved):

PR #6502, opened 2020-11-25
6985 mini timeline discovery

wschaeferB:


Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineArtifactListPanel.java
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * Panel to display list of dates and counts.
+ */
+class MiniTimelineArtifactListPanel extends JPanel {
+
+    private static final long serialVersionUID = 1L;
+    private final TypeDescriptionTableModel tableModel;
+    private static final Logger logger = Logger.getLogger(MiniTimelineArtifactListPanel.class.getName());
+    private static final BlackboardAttribute.ATTRIBUTE_TYPE[] DESCRIPTION_TYPES = {BlackboardAttribute.ATTRIBUTE_TYPE.TSK_TITLE, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_NAME, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PROG_NAME, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_TEXT, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_URL};
```
gdicristofaro, line 44: It would help readability if this was broken up into multiple lines.

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineArtifactListPanel.java
```diff
@@ -0,0 +1,265 @@
+/*
```
gdicristofaro, line 1: It looks like this file was generated with the gui designer.  Is there are corresponding .form file?

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineDateListPanel.java
```diff
@@ -0,0 +1,242 @@
+/*
```
gdicristofaro, line 1: It looks like this file was generated with the gui designer.  Is there are corresponding .form file?

gdicristofaro (changes requested):

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineArtifactListPanel.java
```diff
@@ -0,0 +1,265 @@
...
+    void clearDates() {
+        tableModel.setContents(new ArrayList<>());
+        tableModel.fireTableDataChanged();
+    }
+
+    /**
+     * Initialize the UI components.
+     */
+    private void initComponents() {
+        //This class is a refactored copy of ArtifactsListPanel so lacks the form however the init method still constructs the proper UI elements.
```
wschaeferB, line 144: It was a copy refactor so there is not a form file. I did put this comment in the initComponents file to reflect the nature of this file.

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineDateListPanel.java
wschaeferB: Same as the other list no form file because it was a refactored copy

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineArtifactListPanel.java
gdicristofaro: Yeah, I missed that.

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineWorker.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    protected void done() {
+        List<MiniTimelineResult> results = null;
+        if (!isCancelled()) {
+            try {
+                results = get();
+            } catch (InterruptedException | ExecutionException ex) {
+                logger.log(Level.SEVERE, "Exception while trying to get list of artifacts for Domain details for mini timeline view for domain: " + domain, ex);
```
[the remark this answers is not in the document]
wschaeferB: I don't think it should at this point, since we only want to entirely ignore things if the swing worker is cancelled in which case there will be a cancellation exception instead of an interruptedexception.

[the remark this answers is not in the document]
gdicristofaro: I think the CancellationException is thrown if the get method is called after the swing worker has been cancelled and the InterruptedException is thrown if a waiting thread receives an interrupt during cancellation based on https://docs.oracle.com/javase/tutorial/uiswing/concurrency/cancel.html and https://docs.oracle.com/javase/7/docs/api/javax/swing/SwingWorker.html#get().  I'm not seeing anything in your get method that would directly throw an InterruptedException so it is probably okay, but if that get method was changed to use, say the cache or a sql query directly where an InterruptedException could potentially be thrown, it might be good to treat as a cancellation.

Core/src/org/sleuthkit/autopsy/discovery/ui/MiniTimelineWorker.java
```diff
@@ -0,0 +1,87 @@
...
```
gdicristofaro, line 77: I think the CancellationException is thrown if the get method is called after the swing worker has been cancelled and the InterruptedException is thrown if a waiting thread receives an interrupt during cancellation based on https://docs.oracle.com/javase/tutorial/uiswing/concurrency/cancel.html and https://docs.oracle.com/javase/7/docs/api/javax/swing/SwingWorker.html#get(). I'm not seeing anything in your get method that would directly throw an InterruptedException so it is probably okay, but if that get method was changed to use, say the cache or a sql query directly where an InterruptedException could potentially be thrown, it might be good to treat as a cancellation.

gdicristofaro (approved):

PR #6511, opened 2020-12-01
6912 Add option to not copy files that are present in file repository.

APriestman:


Core/src/org/sleuthkit/autopsy/report/modules/portablecase/PortableCaseReportModule.java
```diff
@@ -1230,6 +1317,123 @@ private void createAppLaunchBatFile(String destBaseFolder) throws IOException {
...
+     * @param files File to look up.
+     * 
+     * @throws TskCoreException 
+     */
+    private void doBulkFileRepositoryLookup(List<AbstractFile> files) throws TskCoreException {
+        try {
+            BulkExistenceResult results = FileRepository.exists(files);
+            for (AbstractFile file : files) {
+                BulkExistenceEnum result = results.getResult(file);
+                if (result != null && result.equals(BulkExistenceEnum.TRUE)) {
```
dannysmyda, line 1427: Hmm, maybe my BulkExistenceEnum should have EXISTS, DOES_NOT_EXIST, INVALID_HASH, instead of True, False, Invalid.

dannysmyda: And maybe it should be renamed FileRepoExistenceEnum and FileRepoExistenceResults

Core/src/org/sleuthkit/autopsy/report/modules/portablecase/PortableCaseReportModule.java
```diff
@@ -103,6 +111,12 @@
     private String caseName = "";
     private File caseFolder = null;
     private File copiedFilesFolder = null;
+    
+    // Stores whether files are present in the file repository
+    private enum FileRepoState {
+        EXISTS, DOES_NOT_EXIST;
+    }
+    private final Map<Long, FileRepoState> fileRepositoryLookupCache = new HashMap<>();
```
dannysmyda, line 119: Are we not going to deal with if the state changes? Also, if the FileRepo is redirected to another server in the settings, this cache needs to be invalidated. Will this PortableCase object be GC'd between Case close and open?

APriestman: I just forgot to add clearing the cache to the cleanup() method. That will take care of it.

Core/src/org/sleuthkit/autopsy/report/modules/portablecase/PortableCaseReportModule.java
APriestman: That might be better. I found the TRUE/FALSE awkward to use

dannysmyda (changes requested):

dannysmyda (approved):

PR #6520, opened 2020-12-03
7060 webmail domains

gdicristofaro:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/MessageURLAnalyzer.java
```diff
@@ -0,0 +1,488 @@
...
+                }
+
+                // make sure there is attached file
+                AbstractFile file = tskCase.getAbstractFileById(artifact.getObjectID());
+                if (file == null) {
+                    continue;
+                }
+
+                // get the url string from the artifact
+                BlackboardAttribute urlAttr = artifact.getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_URL));
```
dannysmyda, line 410: Why not get the TSK_DOMAIN attribute that will be on this artifact? Then, you could skip all the business about getting the host.

dannysmyda: Reviewing some code, this extractor will be run after the browser extractors.

dannysmyda (approved):

PR #6538, opened 2020-12-09
[MultiUserTestTool] fix for wrong created temporary file

mammo0:
Hello,

after setting up a multi user environment in Linux the test fails and tells me that I have insufficient privileges to create a file.

I tracked the error down when the `MultiUserTestTool` class is creating a temporary file: https://github.com/sleuthkit/autopsy/blob/4c66d9b4e610c3d462dd3a16ba8f4fba36fc4220/Experimental/src/org/sleuthkit/autopsy/experimental/configuration/MultiUserTestTool.java#L156
The problem is that at least in Linux the environment variable `java.io.tmpdir` contains by default the string: `"/tmp"`
That means the concatenation will create a string like that:
`String tempFilePath = "/tmpAutopsyTempFile_<timestamp>.txt"`
So autopsy tries to create a file in the root directory `/`. But the file should be created in the `/tmp` directory.

With my fix there's a `java.nio.file.Path` object is created, which is more platform independent. An object of this type is created later anyway.

### DIFF
--- a/Experimental/src/org/sleuthkit/autopsy/experimental/configuration/MultiUserTestTool.java
+++ b/Experimental/src/org/sleuthkit/autopsy/experimental/configuration/MultiUserTestTool.java
@@ -153,7 +153,7 @@ class MultiUserTestTool {
                 return Bundle.MultiUserTestTool_unableToReadDatabase() + ". " + ex.getMessage();
             }
 
-            // Make a text file in Windows TEMP folder 
+            // Make a text file in TEMP folder
             Path tempFilePath = Paths.get(System.getProperty("java.io.tmpdir"), TEST_FILE_NAME + "_" + TimeStampUtils.createTimeStamp() + ".txt");
             try {
                 FileUtils.writeStringToFile(tempFilePath.toFile(), "Test", Charset.forName("UTF-8"));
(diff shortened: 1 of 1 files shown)


Experimental/src/org/sleuthkit/autopsy/experimental/configuration/MultiUserTestTool.java
```diff
@@ -153,16 +154,16 @@ static String runTest(String rootOutputDirectory) {
             }
 
             // Make a text file in Windows TEMP folder 
-            String tempFilePath = System.getProperty("java.io.tmpdir") + TEST_FILE_NAME + "_" + TimeStampUtils.createTimeStamp() + ".txt";
+            Path tempFilePath = Paths.get(System.getProperty("java.io.tmpdir"), TEST_FILE_NAME + "_" + TimeStampUtils.createTimeStamp() + ".txt");
```
esaunders, line 157: You might want to update the comment since the code is no longer Windows specific.

esaunders (approved):

Experimental/src/org/sleuthkit/autopsy/experimental/configuration/MultiUserTestTool.java
mammo0: Done ;)

PR #6547, opened 2020-12-11
7105 domain categories

gdicristofaro:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/DomainCategoryRunner.java
```diff
@@ -0,0 +1,396 @@
...
+
+        for (DomainCategorizer provider : foundProviders) {
+            try {
+                provider.initialize();
+            } catch (DomainCategorizerException ex) {
+                throw new IngestModule.IngestModuleException("There was an error instantiating the provider: " + provider.getClass().getSimpleName(), ex);
+            }
+        }
+
+        this.domainProviders = foundProviders == null
```
dannysmyda, line 377: Can foundProviders be null? I don't think so, but even if it could, you would get an NPE on line 367.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/DomainCategoryRunner.java
```diff
@@ -0,0 +1,396 @@
...
+                // get the pertinent details for this artifact.
+                ArtifactHost curArtHost = getDomainAndHost(artifact);
+                if (curArtHost == null || isDuplicateOrAdd(hostsSeen, curArtHost.getHost())) {
+                    continue;
+                }
+
+                // if we reached this point, we are at least analyzing this item
+                artifactsAnalyzed++;
+
+                // attempt to get the domain type for the host using the suffix trie
```
dannysmyda, line 296: Might want to update this comment

dannysmyda (approved):

PR #6568, opened 2020-12-18
7084 feedback while loading artifacts

wschaeferB:
This PR is branched off of #6556


Core/src/org/sleuthkit/autopsy/discovery/ui/Bundle.properties-MERGED
```diff
@@ -66,12 +68,18 @@ ImageThumbnailPanel.isDeleted.text=All instances of file are deleted.
...
+LoadingPanel.loading.text=Loading, please wait.
+# {0} - resultInfo
+LoadingPanel.retrieving.text=Retrieving results for {0}.
 MiniTimelineArtifactListPanel.descriptionColumn.name=\ Description
 MiniTimelineArtifactListPanel.typeColumn.name=Result Type
 MiniTimelineArtifactListPanel.value.noValue=No value available.
 MiniTimelineDateListPanel.countColumn.name=Count
 MiniTimelineDateListPanel.dateColumn.name=Date
 MiniTimelineDateListPanel.value.noValue=No value available.
+<<<<<<< HEAD
```
gdicristofaro, line 80: I think this might need to be fixed along with line 82 and 95.

wschaeferB: Yes, good catch. I needed a clean and build to get the correct version of the file, and it also reminded me that we wanted to change the display name to just Timeline, I have left the classes and developer facing items named with mini to minimize the effect of further overloading of terms. Thank you.

Core/src/org/sleuthkit/autopsy/discovery/ui/DiscoveryTopComponent.java
```diff
@@ -267,6 +276,25 @@ private void newSearchButtonActionPerformed(java.awt.event.ActionEvent evt) {//G
                 .collect(Collectors.toList());
     }
 
+//    /**
```
dannysmyda, line 279: Is this here for a later time or should it be deleted?

wschaeferB: Will delete, it was part of an earlier attempt to make the panels reset correctly when nothing was selected that I forgot to remove.

dannysmyda (approved):

Issue #3829, opened 2018-06-07
Linux: Library not found in jar (libtsk_jni)


obilodeau:
I'm trying to package autopsy for Arch Linux and I'm facing a strange problem.

If I run it inside my home folder in let's say: `~/opt/autopsy-4.7.0/` everything runs fine.

If I install it system-wide in `/opt/autopsy/` it no longer works. Loading plugins result in a popup with:

```
org.netbeans.InvalidException:
  StandardModule:org.sleuthkit.autopsy.core jarFile: /opt/autopsy/autopsy/modules/org-sleuthkit-autopsy-core.jar:
    java.lang.UnsatisfiedLinkError: org.sleuthkit.datamodel.SleuthkitJNI.getVersionNat()Ljava/lang/String;
```

Command line outputs:
```
$ ./bin/autopsy 
Library not found in jar (libtsk_jni)
SleuthkitJNI: failed to load libtsk_jni
```

Functioning setup command line outputs:

```
$ ./bin/autopsy 
Temp Folder for Libraries: /tmp
SleuthkitJNI: loaded libtsk_jni
log4j: reset attribute= "false".
[...]
```

The difference between the two is that when I run it from my home folder, a file is created in `/tmp/` called `libtsk_jni_olivier.so` (olivier is my username). Looking at the source code, I can't find where is the logic in autopsy to generate that file or to attempt to use it. Is it a netbeans thing?

braytac:
The same to me!!

NgoHuy:
You must cp this line to PKGBUILD and rebuild
`cp /usr/share/java/sleuthkit-4.6.1.jar $pkgdir/usr/autopsy/modules/ext/sleuthkit-postgresql-4.6.1.jar`

obilodeau:
Thanks @NgoHuy, with your help I was able to fix the PKGBUILD: https://gist.github.com/obilodeau/a2f50b9ddd66f16d1b01e61d0c7eb221

vdun:
Can push your updates to BlackArch/ArchStrike ?

obilodeau:
* https://github.com/BlackArch/blackarch/pull/1976
* https://github.com/ArchStrike/ArchStrike/pull/416

sammysaita:
Does this still solve same problem in autopsy 4.12.0

obilodeau:
I think so but the sleuthkit version is probably going to be more recent.

Look at the BlackArch PKGBUILD there it is probably correct.

[...]

PR #6578, opened 2020-12-27
Added JAVA_HOME variable

Parveshdhull:
Script is not detecting JAVA_HOME on own and showing error 
`Checking for Java...ERROR: JAVA_HOME environment variable must be defined.`

And running autopsy without setting JAVA_HOME leading to issue 
[Issue #3829](https://github.com/sleuthkit/autopsy/issues/3829)


Parveshdhull:
Looks like hard coding JAVA path creates problem for OSX

PR #6658, opened 2021-01-20
7244 - Binary email fix

kellykelly3:


thunderbirdparser/src/org/sleuthkit/autopsy/thunderbirdparser/MimeJ4MessageParser.java
```diff
@@ -164,7 +165,13 @@ EmailMessage extractEmail(Message msg, String localPath, long sourceFileID) {
         if (msg.isMultipart()) {
             handleMultipart(email, (Multipart) msg.getBody(), sourceFileID);
         } else {
-            handleTextBody(email, (TextBody) msg.getBody(), msg.getMimeType(), msg.getHeader().getFields());
+            if(msg.getBody() instanceof TextBody) {
+                handleTextBody(email, (TextBody) msg.getBody(), msg.getMimeType(), msg.getHeader().getFields());
+            } else if(msg.getBody() instanceof BinaryBody) {
+               if(msg.getDispositionType().equals("attachment")) {
+                   handleAttachment(email, msg, sourceFileID, 1);
+               }
```
bcarrier, line 173: We should put some log message here if the message is not binarybody or attachment.

kellykelly3: I debated whether or not to check the disposition type or just to send it to handleAttachement.  I can put a message just in case, but I don't believe we will hit that point.  You know what I think I am going to take the disposition check completely out.  It doesn't again us anything.

eugene7646 (approved):

PR #6675, opened 2021-01-25
6736 - Add more cancellation to RA

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Chromium.java
```diff
@@ -32,11 +32,15 @@
 import org.openide.util.NbBundle;
 import org.sleuthkit.autopsy.datamodel.ContentUtils;
 import java.util.logging.Level;
-import java.util.*;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.Collection;
```
dannysmyda, line 39: These seem unused

kellykelly3: No? They are very much used.

dannysmyda: Oh I see, you removed the import of all util classes. I just saw the disconnect between the new imports vs the changes. Ignore.

dannysmyda (approved):

PR #6699, opened 2021-02-03
7284 Cleaned up OpenAction threads

kellykelly3:


kellykelly3:
Replaced by https://github.com/sleuthkit/autopsy/pull/6704

PR #6722, opened 2021-02-11
update brew cask install to brew install --cask

sanjitk7:
For MacOS in 2021:
brew cask install is no longer valid and has since been replaced with brew install --cask 

check https://brew.sh


markmckinnon (commented):
If the change is only for Big Sur, you should keep the old command and note that it is for releases prior to Big Sur, and this command is for Big Sur or something to that effect.

sanjitk7:
> If the change is only for Big Sur, you should keep the old command and note that it is for releases prior to Big Sur, and this command is for Big Sur or something to that effect.

made changes stating the same.

markmckinnon (approved):

PR #6734, opened 2021-02-18
7275 os account in ra

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1116,6 +1142,34 @@ private boolean parseSamPluginOutput(String regFilePath, AbstractFile regAbstrac
             for (Map<String, String> userInfo : userSet) {
                 userInfoMap.put(userInfo.get(SID_KEY), userInfo);
             }
+            
+            // New OsAccount Code 
+            OsAccountManager accountMgr = tskCase.getOsAccountManager();
+            HostManager hostMrg = tskCase.getHostManager();
+            Host host = hostMrg.getHost(tskCase.getDataSource(regAbstractFile.getDataSourceObjectId()));
```
bcarrier, line 1149: We should be able to get the host from the DataSource that was passed into RecentActivity. It should have a getHost() method on it.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -2138,4 +2192,277 @@ public void process(Content dataSource, IngestJobContext context, DataSourceInge
...
+     *
+     * @param file     Registry file
+     * @param sid      Account sid
+     * @param userName Login name
+     * @param homeDir  Account home Directory
+     *
+     * @throws TskCoreException
+     * @throws TskDataException
+     */
+    private void updateOsAccount(AbstractFile file, String sid, String userName, String homeDir) throws TskCoreException, TskDataException {
```
bcarrier, line 2207: I'd rename this createOrUpdateOsAccount to make it more clear.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
kellykelly3: I am sorry, I am not following.  Content doesn't seem to have a getHost method.

bcarrier: Right, but DataSource does and it should be a DataSource object that is passed into RecentActivity.

APriestman (approved):

PR #6738, opened 2021-02-19
4138: HostManager returns a list instead of a set.

raman-bt:


raman-bt:
Pairs with TSK PR: https://github.com/sleuthkit/sleuthkit/pull/2182

APriestman (approved):

PR #6745, opened 2021-02-23
7278 os account event firing

kellykelly3:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -129,6 +131,10 @@
 import org.sleuthkit.datamodel.DataSource;
 import org.sleuthkit.datamodel.FileSystem;
 import org.sleuthkit.datamodel.Image;
+import org.sleuthkit.datamodel.OsAccount;
+import org.sleuthkit.datamodel.OsAccountManager;
+import org.sleuthkit.datamodel.OsAccountManager.OsAccountsAddedEvent;
+import org.sleuthkit.datamodel.OsAccountManager.OsAccountsChangedEvent;
```
wschaeferB, line 137: These two datamodel.OsAccountManager.OsAccounts..... imports wind up being unused because their usage is specified as OsAccountManager.OsAccountsAddedEvent and OsAccountManager.OsAccountsChangedEvent below when used as parameters for the methods.

Do we really want these public classes in the sleuthkint.datamodel level to have the same class name as the classes in the casemodule.events package? It seems like mistakes using the wrong one would be easy to make.

wschaeferB: I guess this is probably a question for @bcarrier unless it was covered in the design doc.

kellykelly3: You are right @wschaeferB.  Let me see if I can think of better names for the data model classes.

wschaeferB (approved):

PR #6771, opened 2021-03-04
7323 dss excel export

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datasourcesummary/ui/ExcelExportAction.java
```diff
@@ -0,0 +1,310 @@
...
+        int returnVal = fc.showSaveDialog(dialogParent);
+
+        if (returnVal != JFileChooser.APPROVE_OPTION || fc.getSelectedFile() == null) {
+            return null;
+        }
+
+        File file = fc.getSelectedFile();
+
+        // insert xlsx if not at the end.
+        if (!file.getAbsolutePath().endsWith("." + expectedExtension)) {
```
eugene7646, line 160: Strictly as FYI, FilenameUtils package has convenient methods to get file extension, name, etc. See example:
https://github.com/sleuthkit/autopsy/blob/develop/Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/SupportedDataSources.java#L43

PR #6780, opened 2021-03-10
7376 - Fixed OsAccountFactoryNode exception

kellykelly3:


Core/src/org/sleuthkit/autopsy/datamodel/OsAccounts.java
```diff
@@ -111,34 +111,46 @@ public String getItemType() {
...
+                        skCase = null;
+                    }
+                }
             }
         };
         
         @Override
         protected void addNotify() {
             Case.addEventTypeSubscriber(Collections.singleton(Case.Events.OS_ACCOUNT_ADDED), listener);
+            Case.addEventTypeSubscriber(EnumSet.of(Case.Events.CURRENT_CASE), listener);
```
gdicristofaro, line 130: I needed to import EnumSet

gdicristofaro (changes requested):

Core/src/org/sleuthkit/autopsy/datamodel/OsAccounts.java
kellykelly3: Line 25 had the EnumSet import.

gdicristofaro: I pulled in develop just to make sure it would build with what I have for sleuthkit and that removed the import.  I guess some other work has been happening in this file that is overwriting.  Would you please pull in latest from develop?

kellykelly3: Seems like something happens when you merged in develop.  It should be all set now.

gdicristofaro: Great.  Yes, that's working for me now.  Thanks!

gdicristofaro (approved):

PR #6812, opened 2021-03-22
Handle NotUserSIDException.

raman-bt:


raman-bt:
Goes with TSK PR: https://github.com/sleuthkit/sleuthkit/pull/2251

APriestman (approved):

PR #6824, opened 2021-03-25
7455 - Fixed content panel and RA

kellykelly3:


Core/src/org/sleuthkit/autopsy/contentviewers/osaccount/OsAccountViewer.java
```diff
@@ -53,26 +53,29 @@ public OsAccountViewer() {
...
+                    osAccountId = optional.get();
                 }
             }
 
-            if (osAccount == null) {
+            if (osAccountId == null) {
                 DataArtifact dataArtifact = node.getLookup().lookup(DataArtifact.class);
                 if (dataArtifact != null) {
-                    Optional<OsAccount> optional = dataArtifact.getOsAccount();
+                    optional = dataArtifact.getOsAccountObjectId();
```
wschaeferB, line 76: I think this method went away or something.

wschaeferB: Oh nevermind the story that causes this story is still open.

kellykelly3: Yup, you need the PR for sleuthkit from the story.

wschaeferB (approved):

PR #6847, opened 2021-03-31
4251: Reconcile different OsAccount update philosophies.

raman-bt:


raman-bt:
Depends on: https://github.com/sleuthkit/sleuthkit/pull/2283

branding/core/core.jar/org/netbeans/core/startup/Bundle.properties
```diff
@@ -1,5 +1,5 @@
 #Updated by build script
-#Mon, 25 Jan 2021 12:41:22 -0500
+#Thu, 01 Apr 2021 23:53:50 -0400
```
APriestman, line 2: Please remove changes to this file (and the next)

APriestman (approved):

PR #6855, opened 2021-03-31
7490 - Fixed OsAccountDataPanel NPE

kellykelly3:


Core/src/org/sleuthkit/autopsy/contentviewers/osaccount/OsAccountDataPanel.java
```diff
@@ -50,6 +50,7 @@
 import org.sleuthkit.datamodel.OsAccountManager;
 import org.sleuthkit.datamodel.OsAccountRealm;
 import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
```
APriestman, line 53: This does not appear necessary

APriestman (approved):

PR #6866, opened 2021-04-01
7474 limited ocr option

wschaeferB:


Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -158,7 +160,7 @@
                     .map(mt -> mt.getType() + "/" + mt.getSubtype())
                     .collect(Collectors.toList());
 
-    public TikaTextExtractor(Content content) {
+    TikaTextExtractor(Content content) {
```
eugene7646, line 163: @wschaeferB @rcordovano I think this might technically be a public API change, since org.sleuthkit.autopsy.textextractors is a publicly exported package.

Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -528,28 +546,34 @@ private static int getTimeout(long size) {
...
             ImageConfig configInstance = context.lookup(ImageConfig.class);
             if (configInstance != null) {
                 if (Objects.nonNull(configInstance.getOCREnabled())) {
                     this.tesseractOCREnabled = configInstance.getOCREnabled();
                 }
+                if (Objects.nonNull(configInstance.getLimitedOCREnabled())) {
+                    this.limitedOCREnabled = configInstance.getLimitedOCREnabled();
+                }
+                else {
+                    System.out.println("NULL SETTING");
```
eugene7646, line 561: @wschaeferB What is the idea with System.out.println("NULL SETTING") ?

wschaeferB: @eugene7646 me messing up and leaving in conditionals I was using to debug when I didn't have it working because I had the file size mistakenly set to 1 GB

Thank you for catching it and sorry you had to, I will get it removed asap.

eugene7646 (approved):

Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
wschaeferB: My bad we @eugene7646, @rcordovano and I discussed something related but not the removal of the public keywords in these instances, that is a mistake as well caused by me for whatever reason thinking because there was no public constructor in a final class no one would be able to call any of those methods, I forgot that other classes in the package could still return TikaTextExtractor or a lookup could still find one. I will add the public keywords back and I am sorry for that mistake.

wschaeferB: @eugene7646 @rcordovano wait no the class itself isn't public but it had a public constructor, can that constructor ever even be used publicly if the class isn't public?

PR #6887, opened 2021-04-08
7500-prefetchChanges

gdicristofaro:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractPrefetch.java
```diff
@@ -282,11 +306,13 @@ private void createAppExecArtifacts(String prefetchDb, Content dataSource) {
                                         BlackboardAttribute.ATTRIBUTE_TYPE.TSK_COMMENT, getName(), PREFETCH_TSK_COMMENT));
 
                         try {
-                            BlackboardArtifact blkBrdArt = createArtifactWithAttributes(BlackboardArtifact.ARTIFACT_TYPE.TSK_PROG_RUN, pfAbstractFile, blkBrdAttributes);
-                            blkBrdArtList.add(blkBrdArt);
-                            BlackboardArtifact associatedBbArtifact = createAssociatedArtifact(applicationName.toLowerCase(), filePath, blkBrdArt, dataSource);
-                            if (associatedBbArtifact != null) {
-                                blkBrdArtList.add(associatedBbArtifact);
+                            if (!blackboard.artifactExists(dataSource, BlackboardArtifact.ARTIFACT_TYPE.TSK_PROG_RUN, blkBrdAttributes)) {
```
APriestman, line 309: I would rather not add artifact deduplication as part of this PR (assuming I'm reading this right and it wasn't there before). I believe this would also be the one place in RA that does it which would add a strange inconsistency where every artifact from RA except program run would appear multiple times.

APriestman (approved):

APriestman (approved):

APriestman (approved):

PR #6901, opened 2021-04-12
7505 temp path changes

gdicristofaro:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -1468,15 +1471,104 @@ public String getOutputDirectory() {
...
+            List<String> remainingList = pathPieces.subList(1, pathPieces.size());
+            dirFile = Paths.get(first, remainingList.toArray(new String[remainingList.size()])).toFile();
+        }
+        
+        // ensure directory exists
+        if (!dirFile.exists()) {
+            dirFile.mkdirs();
+        }
+        
+        return dirFile.getAbsolutePath();
```
rcordovano, line 1571: I think that a somewhat different algorithm will allow us to separate concerns a little bit more. I think the key is to think of the application temp directory and the temp directory choice as two separate things. If there is no current case, the client can call Path UserMachinePreferences.getAppTempDirPath(); otherwise most clients can call Case.getTempDirectory(). 

We can also use the Path and Paths classes to work at a higher level of abstraction than String arrays and perhaps make the code easier to follow.  Here is what I'm thinking can go here:

```
if UserMachinePreferences.getTempDirectoryType() == CASE
	if getCaseType() == SINGLE_USER_CASE
		return Paths.get(getCaseDirectory(), TEMP_FOLDER).toString()
	else
		return Paths.get(getCaseDirectory(), getHostName(), TEMP_FOLDER).toString()
else
	Path appTempDirPath = UserMachinePreferences.getAppTempDirPath()
	Path caseTempDirPath = appTempDirPath.subpath(0, appTempDirPath.getNameCount() - 1) 
	caseTempDirPath = Paths.get(caseTempDirPath, getName());
	if getCaseType() == SINGLE_USER_CASE
		return Paths.get(caseTempDirPath.toString(), TEMP_FOLDER).toString()
	else
		return Paths.get(caseTempDirPath.toString(), getHostName(), TEMP_FOLDER)
```

Core/src/org/sleuthkit/autopsy/machinesettings/UserMachinePreferences.java
```diff
@@ -175,7 +141,7 @@ public static String getTempDirectory() {
     /**
      * @return The user-specified custom temp directory path or empty string.
      */
-    public static String getCustomTempDirectory() {
+    public static String getCustomTempDirectoryProperty() {
```
rcordovano, line 144: I know that the use of "Property" has been proposed as a way to make the intent of these methods more clear, but that is not consistent with our usage to this point. I think that we can achieve the same clarity without changing our style with Path getCustomTempDirectoryPath(). Note the return type.

Core/src/org/sleuthkit/autopsy/machinesettings/UserMachinePreferences.java
```diff
@@ -80,89 +79,56 @@
...
+     *
+     * NOTE: The Case class also handles providing a temp directory. When
+     * altering code in this class, also look at the Case class as well.
+     * See JIRA-7505 for more information.
      *
-     * @return The directory.
+     * @return The base user-specified temporary directory.
      */
-    private static File getTempDirFile() {
+    public static String getTempDirectory() {
```
rcordovano, line 117: See comments to Case class changes. I think we should make this method Path getAppTempDirectoryPath(). Note return type.

rcordovano (changes requested):

rcordovano (approved):

PR #6929, opened 2021-04-19
7537-IllegalArgumentException-in-Chromium-RA-module

markmckinnon:
Check for buffer capacity of zero (0) and return if it is, nothing to look at.


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -1039,6 +1039,9 @@ void extract() throws TskCoreException, IngestModuleException {
                 this.data = new byte [length];
                 ByteBuffer buf = cacheFileCopy.getByteBuffer();
                 int dataOffset = DATAFILE_HDR_SIZE + cacheAddress.getStartBlock() * cacheAddress.getBlockSize();
+                if (buf.capacity() == 0) {
+                    return;
+                }
```
APriestman, line 1044: Would it make more sense to check that the capacity is greater than what we're about to read? (I think dataOffset + length)

markmckinnon: Yes it would, good catch, thanks, done.

APriestman (approved):

PR #6946, opened 2021-04-29
7553 data model event changes

rcordovano:
- Make new data model event classes capable of batching data model objects
- Make data model object deletion events work in terms of IDs, not objects


Core/src/org/sleuthkit/autopsy/casemodule/events/HostsAddedToPersonEvent.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @return The hosts.
+     */
+    public List<Host> getHosts() {
+        return getNewValue();
+    }
+
+    @Override
+    protected List<Person> getOldValueObjects(SleuthkitCase caseDb, List<Long> ids) throws TskCoreException {
+        List<Person> persons = new ArrayList<>();
```
gdicristofaro, line 70: For this class, `HostsAddedToPersonEvent`, and the `HostsRemovedFromPersonEvent` class, would it make sense for the old and new values to relate to the list of `Host` objects (ids or objects)?  I initially had a little confusion that the 'new values' were the child `Host`s and the 'old value' was the parent `Person` object.

gdicristofaro (changes requested):

PR #6948, opened 2021-05-03
Fix Java error on MacOS

seaturtle15:
After following the installation guide to the letter, you still get an error saying that it can't find Java when trying to run autopsy.
If you look at bin/autopsy, the variable being used is $jdkhome, not $JAVA_HOME. Fixing this allows the app to open.

### DIFF
--- a/Running_Linux_OSX.txt
+++ b/Running_Linux_OSX.txt
@@ -46,7 +46,7 @@ The following need to be done at least once. They do not need to be repeated for
 
      % sudo ldconfig  
 
-- Install the BellSoft Java 8 JRE and JavaFX 8 distribution and set JAVA_HOME.
+- Install the BellSoft Java 8 JRE and JavaFX 8 distribution and set jdkhome.
   * The BellSoft distribution bundles OpenJDK and OpenJFX. Other distributions we have tried either don't
     bundle OpenJFX (AdoptOpenJDK) or don't include all necessary binaries (Amazon Corretto).
 -- Linux:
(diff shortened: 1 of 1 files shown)


gdicristofaro:
Hi @seaturtle15 ,
Thanks for reaching out.  Could you share more about the error that you are hitting with just JAVA_HOME set?  Also, which version of OS X are you using and would you mind including the output of these commands?
```
sh
/usr/libexec/java_home
/usr/libexec/java_home --version 1.8.0+ --failfast
```

seaturtle15:
Thanks for the prompt reply! I'm super excited to use autopsy and I really appreciate all the work you do to maintain it :)

1. I'm using Big Sur.
```
seaturtle15@MacBookAir ~ $ sw_vers                                                            
ProductName:    macOS
ProductVersion: 11.3
BuildVersion:   20E232
```
2. Output of the commands:
```
seaturtle15@MacBookAir ~ $ sh                                                                                                                                        
sh-3.2$ /usr/libexec/java_home
/Library/Java/JavaVirtualMachines/jdk-15.0.1.jdk/Contents/Home
sh-3.2$ /usr/libexec/java_home --version 1.8.0+ --failfast
The operation couldn’t be completed. Unable to locate a Java Runtime that supports (null).
Please visit http://www.java.com for information on installing Java.

```
3. With just JAVA_HOME set:
```
seaturtle15@MacBookAir ~/Other/autopsy-4.18.0/bin $ ./autopsy                                                                                                        
The operation couldn’t be completed. Unable to locate a Java Runtime that supports (null).
Please visit http://www.java.com for information on installing Java.

Cannot find java. Please use the --jdkhome switch.
```
4. Making the following edit:
```
export jdkhome=$(/usr/libexec/java_home -v 1.8) # in ~/.bashrc

# the following does not need to be executed again after following the installation guide obviously, 
# but if I were starting from scratch, I would have made sure I used this command:
seaturtle15@MacBookAir ~ $ ln -s $jdkhome /usr/local/opt/openjdk
```
After sourcing ~/.bashrc and executing bin/autopsy, the app launches as expected.

Also, as a side note, in the linked installation guide it details how to make an app launcher that appears in Launchpad. I just want to point out that you need to add
```
 export jdkhome=$(/usr/libexec/java_home -v 1.8)
```
before the next line with the executable in order to set the variable.

gdicristofaro:
Hi @seaturtle15 ,

I played around with this on OS X machine, and I am seeing a similar problem.  The `$jdkhome` solution looks like a good one, but I think it would be an improvement to set `$jdkhome` in the `/etc/autopsy.conf` file, which would mean we wouldn't need to specify environmental variables in `.bashrc`.  So, if you want to put in a PR changing the OS X directions to use the `/etc/autopsy.conf` file, that would be great.  Otherwise, we can make that change.  Thanks for catching this!

PR #6952, opened 2021-05-05
7555-modified updateguiforcaseopened to get db calls off of edt

kellykelly3:


Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -63,6 +64,7 @@
 import javax.swing.JOptionPane;
 import javax.swing.SwingUtilities;
 import org.apache.commons.lang3.StringUtils;
+import org.openide.util.Exceptions;
```
wschaeferB, line 67: unused import

wschaeferB (approved):

PR #6957, opened 2021-05-06
7557 move directory tree queries off edt

wschaeferB:


Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeTopComponent.java
```diff
@@ -1056,6 +1073,24 @@ protected void done() {
...
     }
 
+    /**
+     * Identify if the specified case has data.
+     *
+     * @param currentCase The case you are checking for data.
+     *
+     * @return True if the case exists and has data, false otherwise.
+     */
+    private static boolean caseHasData(Case currentCase) {
```
kellykelly3, line 1083: This is fine for now but after my hasData story we might want to swing back and see if we need to modify

kellykelly3 (approved):

PR #6961, opened 2021-05-10
Add Registry Key Modification Time To Registry Content Viewer

markmckinnon:
Added registry key modification time to registry content viewer for a registry key when it is selected.  Date/time is in GMT + 0 (UTC)


Core/src/org/sleuthkit/autopsy/rejview/RejTreeKeyView.java
```diff
@@ -95,7 +102,12 @@ public RejTreeKeyView(RejTreeKeyNode node) {
             numValues = -1;
         }
 
-        JLabel metadataLabel = new JLabel(String.format(metadataTemplate, keyName, numSubkeys, numValues), JLabel.LEFT);
+        Date date = new java.util.Date(node.getKey().getTimestamp().getTimeInMillis());
```
APriestman, line 105: Is there any chance getTimestamp() could return null? It doesn't look like it to me but I'm not sure.

APriestman (approved):

PR #6979, opened 2021-05-18
7589 is preferred for artifact categories

gdicristofaro:


Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -344,25 +354,49 @@ public boolean isSupported(Node node) {
...
-        } else {
-            return 6;
+        
+        // if there is an artifact, get the type
+        BlackboardArtifact.Type artifactType = null;
+        if (artifact != null) {
+            try {
+                artifactType = artifact.getType();
+            } catch (TskCoreException ex) {
+                logger.log(Level.WARNING,
```
rcordovano, line 370: If we have a problem executing a case database query, we should log it at the SEVERE level. An example of an appropriate use of the WARNING level would be reading the contents of a file, where the file might be corrupted

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -63,6 +68,11 @@
     private final static Logger logger = Logger.getLogger(DataContentViewerArtifact.class.getName());
     private final static String WAIT_TEXT = NbBundle.getMessage(DataContentViewerArtifact.class, "DataContentViewerArtifact.waitText");
     private final static String ERROR_TEXT = NbBundle.getMessage(DataContentViewerArtifact.class, "DataContentViewerArtifact.errorText");
+    
+    // Value to return in isPreferred if this viewer is less preferred.
+    private static final int LESS_PREFERRED = 3;
+    // Value to return in isPreferred if this viewer is more preferred.
+    private static final int MORE_PREFERRED = 6;
```
rcordovano, line 75: Nice use of named constants!

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -344,25 +354,49 @@ public boolean isSupported(Node node) {
...
+        }
+        
+        switch (category) {
+            // data artifacts should be more preferred
+            case DATA_ARTIFACT:
+                return MORE_PREFERRED;
+            // everything else is less preferred
+            case ANALYSIS_RESULT:
+            default:
+                return LESS_PREFERRED;
```
rcordovano, line 399: I am finding this logic a bit hard to follow with some unnecessary null checks and setting of local variables to null to act as flow of control. Consequently, it's not clear to me that we are avoiding all NPEs. If there is no artifact in the lookup, we need not proceed any further.  BlackboardArtifact.getType() will always return a value or an exception will be thrown, so we don't need to null check that. The Type will always have a category, so we don't need to null check that. In any case, let's try to make this easier to read, please. Also, we should comment on why we are treating web download and cache artifacts differently. It's because for these artifacts, the important thing is the content that was downloaded or cached, not the artifact itself.

rcordovano (changes requested):

Core/src/org/sleuthkit/autopsy/corecomponents/DataContentViewerArtifact.java
```diff
@@ -347,22 +351,39 @@ public boolean isSupported(Node node) {
...
+        }
+
+        // get the type of the artifact
+        BlackboardArtifact.Type artifactType = null;
+        try {
+            artifactType = artifact.getType();
+        } catch (TskCoreException ex) {
+            logger.log(Level.SEVERE,
+                    String.format("There was an error getting the artifact type for artifact with id: %d", artifact.getId()),
+                    ex);
```
APriestman, line 367: I think the logic would flow better if we returned LESS_PREFERRED here immediately after the error getting the type.

APriestman (approved):

PR #6989, opened 2021-05-25
7581 Remove name from extracted files

APriestman:


Core/src/org/sleuthkit/autopsy/modules/embeddedfileextractor/SevenZipExtractor.java
```diff
@@ -724,7 +724,7 @@ boolean unpack(AbstractFile archiveFile, ConcurrentHashMap<Long, Archive> depthM
                 if (checkForIngestCancellation(archiveFile)) {
                     return false;
                 }
-                final String uniqueExtractedName = FileUtil.escapeFileName(uniqueArchiveFileName + File.separator + (inArchiveItemIndex / 1000) + File.separator + inArchiveItemIndex + "_" + new File(pathInArchive).getName());
+                final String uniqueExtractedName = FileUtil.escapeFileName(uniqueArchiveFileName + File.separator + (inArchiveItemIndex / 1000) + File.separator + inArchiveItemIndex);
```
eugene7646, line 727: @APriestman perhaps we can take this a bit further and truncate other parts of the path as well and make sure that it is as short as possible. Like uniqueArchiveFileName for example, that one can be quite long as well? In getUniqueName() we do file.getName() + "_" + file.getId(). Perhaps we can just do file.getId() instead?

APriestman: Seems like that should work - either way it should correspond to exactly one file in the case. I'll do some testing.

eugene7646: Great. The problem in production is that the case names are typically VERY long, so that doesn't leave much space for path elements inside the case output folder. We should reduce the path length as much as possible.

eugene7646:
@APriestman Are we going to have backwards compatibility problems if we change EmbeddedFileExtractorModuleFactory Ln 34
"EmbeddedFileExtractorIngestModule.ArchiveExtractor.moduleName" from "Embedded File Extractor" to "EFE"? At first glance I don't think so. All of the paths to previously extracted files are already stored in the case database, right?

APriestman:
I think it's fine to shorted the folder name and get rid of another 20 characters. The path is indeed just stored in the database, so it worked fine to have some files in the old folder and some in the new.

eugene7646 (approved):

PR #6992, opened 2021-05-25
7631-Use-Signatures-for-VHD-detection-in-VM-extractor-ingest-module

markmckinnon:
Check if valid VHD files are in the list of files to extract.

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/modules/vmextractor/VMExtractorIngestModule.java
+++ b/Core/src/org/sleuthkit/autopsy/modules/vmextractor/VMExtractorIngestModule.java
@@ -49,6 +49,7 @@ import org.sleuthkit.autopsy.ingest.IngestManager;
 import org.sleuthkit.autopsy.ingest.IngestMessage;
 import org.sleuthkit.autopsy.ingest.IngestModule;
 import org.sleuthkit.autopsy.ingest.IngestServices;
+import org.sleuthkit.autopsy.modules.filetypeid.FileTypeDetector;
 import org.sleuthkit.datamodel.AbstractFile;
 import org.sleuthkit.datamodel.Content;
 import org.sleuthkit.datamodel.DataSource;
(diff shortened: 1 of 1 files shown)


Core/src/org/sleuthkit/autopsy/modules/vmextractor/VMExtractorIngestModule.java
```diff
@@ -237,6 +239,46 @@ public ProcessResult process(Content dataSource, DataSourceIngestModuleProgress
...
+
+        for (AbstractFile vmFile : vmFiles) {
+            if (vmFile.getNameExtension().equalsIgnoreCase("vhd")) {
+                String fileMimeType = vmFile.getMIMEType();
+                if (fileMimeType == null) {
+                    FileTypeDetector fileTypeDetector = null;
+                    try {
+                        fileTypeDetector = new FileTypeDetector();
+                    } catch (FileTypeDetector.FileTypeDetectorInitException ex) {
+                        logger.log(Level.WARNING, String.format("Unable to create file type detector for determining MIME type for file %s with id of %d", vmFile.getName(), vmFile.getId()));
```
APriestman, line 262: If this error occurs we will get an NPE on the next line.

Core/src/org/sleuthkit/autopsy/modules/vmextractor/VMExtractorIngestModule.java
```diff
@@ -237,6 +239,46 @@ public ProcessResult process(Content dataSource, DataSourceIngestModuleProgress
...
+    private static List<AbstractFile> removeNonVMFiles(List<AbstractFile> vmFiles) {
+        List<AbstractFile> vFile = new ArrayList<>();
+
+        for (AbstractFile vmFile : vmFiles) {
+            if (vmFile.getNameExtension().equalsIgnoreCase("vhd")) {
+                String fileMimeType = vmFile.getMIMEType();
+                if (fileMimeType == null) {
+                    FileTypeDetector fileTypeDetector = null;
+                    try {
+                        fileTypeDetector = new FileTypeDetector();
```
APriestman, line 260: Can we make one FileTypeDetector at the beginning and reuse it?

markmckinnon: Yes, done.

Core/src/org/sleuthkit/autopsy/modules/vmextractor/VMExtractorIngestModule.java
markmckinnon: Good catch, added abstract file to list and then continue on with loop to avoid the NPE.

APriestman (approved):

PR #6997, opened 2021-05-26
7611 Make listeners in ChildFactory.Detachable weak

kellykelly3:


kellykelly3:
I screwed something up on my end so there is a new PR for this story
https://github.com/sleuthkit/autopsy/pull/7041

PR #7019, opened 2021-06-04
7645 Added the call to getUniquePath to the makeKey method

kellykelly3:


Core/src/org/sleuthkit/autopsy/datamodel/ContentChildren.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.logging.Level;
+import org.openide.util.Exceptions;
```
wschaeferB, line 25: Left over unused import from using Netbean's surround with try/catch option, but the PR works and seems good.

wschaeferB (approved):

PR #7042, opened 2021-06-14
7696 view context fix

gdicristofaro:


Core/src/org/sleuthkit/autopsy/directorytree/ViewContextAction.java
```diff
@@ -168,176 +169,223 @@ public ViewContextAction(String displayName, Content content) {
...
-            }
-            
-            if ((parentContent != null) 
-                    && (parentContent instanceof UnsupportedContent)) {
+            Content parentContent = getParentContent(this.content);
+
+            if ((parentContent != null) && (parentContent instanceof UnsupportedContent)) {
                 MessageNotifyUtil.Message.error(Bundle.ViewContextAction_errorMessage_unsupportedParent());
                 logger.log(Level.WARNING, String.format("Could not navigate to unsupported content with id: %d", parentContent.getId())); //NON-NLS
-                return;
```
wschaeferB, line 191: This seems like a return we would want since creating the parentTreeNode and settingNodeSelection does not seem to have much purpose if can't navigate to the node, or am I missing something?

wschaeferB (approved):

PR #7043, opened 2021-06-15
7680 ocr only

gdicristofaro:


Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -171,9 +169,29 @@
...
+     * to see if the limited OCR setting is enabled. If it is it will also check
+     * that one of the limiting factors is true.
+     *
+     * @param file    The AbstractFile which OCR might be performed on.
+     * @param boolean The configuration setting which indicates if limited OCR
+     *                is enabled in Keyword Search.
+     *
+     * @return True if limited OCR is not enabled or the image is greater than
+     *         100KB in size or the image is a derived file.
+     */
```
rcordovano, line 183: I don't think this comment is accurate at this point. One of the goals that emerged for for this story was to insulate the text extractors from the "limited OCR" workaround.

Core/src/org/sleuthkit/autopsy/textextractors/TikaTextExtractor.java
```diff
@@ -171,9 +169,29 @@
...
+    public boolean willUseOCR() {
+        // If Tesseract has been installed and is set to be used through
+        // configuration, then ocr is enabled. OCR can only currently be run on 64
+        // bit Windows OS.
+        return TESSERACT_PATH != null
+                && tesseractOCREnabled
+                && PlatformUtil.isWindowsOS()
+                && PlatformUtil.is64BitOS()
+                // if limited OCR is not enabled or the image is greater than 100KB in size or the image is a derived file.
+                && content instanceof AbstractFile;
```
rcordovano, line 194: Don't we need to refactor the MIME type check into this code to answer the question "will use OCR?" It seems like this implementation is only affirming that the text extractor might use OCR, not that it will use OCR.

KeywordSearch/src/org/sleuthkit/autopsy/keywordsearch/KeywordSearchIngestModule.java
```diff
@@ -547,17 +591,17 @@ public Reader openStream() throws IOException {
                 }
                 //divide into chunks and index
                 return Ingester.getDefault().indexText(finalReader, aFile.getId(), aFile.getName(), aFile, context);
-            } catch (TextExtractorFactory.NoTextExtractorFound | TextExtractor.InitReaderException ex) {
+            } catch (TextExtractor.InitReaderException ex) {
```
rcordovano, line 594: I think the comment that follows here needs to be relocated. We did find a text extractor.

rcordovano (changes requested):

rcordovano (changes requested):

rcordovano (approved):

PR #7047, opened 2021-06-16
7708 preExpandNodes fix

gdicristofaro:


Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeTopComponent.java
```diff
@@ -201,30 +178,28 @@ public void popupMenuCanceled(PopupMenuEvent e) {
...
-                        .filter((n) -> n != null)
-                        .forEach((n) -> tree.expandNode(n));
-            }
+            Stream.of(rootChildrenNodes)
+                    .flatMap((n) -> getHostNodesAndParents(n).stream())
+                    .filter((n) -> n != null)
+                    .forEach(tree::expandNode);
+        } else {
+            // nodes to be opened if present at top level
+            Set<String> resultArtifactNames
```
kellykelly3, line 196: Is there a particular reason you are using a set instead of a list?  List are more efficent when adding items.  This method is called on the EDT so we need to be thoughtful about performance.

gdicristofaro: I was trying to optimize for the `contains` operation on line 201 for O(1) performance.

kellykelly3: If it doesn't need to be created more than once I would move it, however static class fields for DirectoryTreeTopComponent will be constructed on the EDT.

gdicristofaro: Ok.  I just moved it to a static field.

kellykelly3 (approved):

PR #7051, opened 2021-06-17
7663 artifact unique path

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -650,12 +640,19 @@ protected Sheet createSheet() {
                     NO_DESCR,
                     actualMimeType));
         }
+        
+        BlackboardArtifact.Type artifactType = null;
+        try {
+            artifactType = artifact.getType();
```
kellykelly3, line 646: For custom artifact types this is adding a DB call onto the EDT.  I know that its an edge case but it is a potential performance issue.  While we are trying to sort out other issues it doesn't make sense to add extra DB calls.

gdicristofaro: Sure.  Did you have something in mind for how to fix this?  Do you want to move the call into the constructor with `srcContent.getUniquePath()` and have an instance field with the category?

kellykelly3: The constructor is also on the EDT.

gdicristofaro: Sure.  I believe the call to `srcContent.getUniquePath()` currently in the constructor could potentially be making a database call (at least for `AbstractFile`) as well.  We could move the potential db call for the artifact type into the constructor so that whenever we fix the db/EDT issue for `BlackboardArtifactNode`, these two calls will be in the same place.  Did you want to do that or did you have something else in mind?

kellykelly3: The artifact type is not going to change so having this call in createSheet doesn't make sense.  Your best option, for now, is to cache the type on construction.

gdicristofaro: Sounds good.  Will do.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -237,6 +228,15 @@ public void propertyChange(PropertyChangeEvent evt) {
     public BlackboardArtifactNode(BlackboardArtifact artifact, String iconPath) {
         super(artifact, createLookup(artifact, false));
         this.artifact = artifact;
+        
+        BlackboardArtifact.Type artType = null;
```
kellykelly3, line 232: This temp var seems unnecessary, just set this.artifactType directly.

gdicristofaro: I made `BlackboardArtifactNode.artifactType` a final variable since it shouldn't change for the lifetime of the object similar to `BlackboardArtifactNode.artifact`.  I think assigning directly within the try/catch won't compile because it could potentially be unassigned if there was an exception.  Is it okay to leave as is given that circumstance or did you have something else in mind?

kellykelly3: Make a method, that takes a blackboard artifact and returns the type.  The BlackboardArtifactNode constructor already has a lot going on, moving this bit to its own method will be cleaner then all this, plus you can assign the class var without the need of a temp var.

gdicristofaro: Ok.  Good to go.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -655,7 +664,7 @@ protected Sheet createSheet() {
          * If the type of the artifact represented by this node dictates the
          * addition of the source content's unique path, add it to the sheet.
          */
-        if (Arrays.asList(SHOW_UNIQUE_PATH).contains(artifactTypeId)) {
+        if (artifactType != null && artifactType.getCategory() == Category.ANALYSIS_RESULT) {
```
kellykelly3, line 667: I just had another though.  Would all artifacts of Category type ANALYSIS_RESULT be AnalysisResult objects?  Can't you just do an instanceof instead?

gdicristofaro: Unfortunately, I don't think so.  I think `getBlackboardArtifacts` returns a `BlackboardArtifact`.  That would be nice if we could.

kellykelly3 (approved):

PR #7052, opened 2021-06-17
installer is created with RELEASE build.type

gdicristofaro:


build.xml
```diff
@@ -343,7 +343,11 @@
     
     <target name="versioning-script" depends="check-release, versioning-script-if-release, versioning-script-if-not-release"/>
      
-    <target name="build-installer" depends="getProps, doxygen, build-zip" description="Builds Autopsy installer.">    
+    <target name="set-release-build">
+        <property name="build.type" value="RELEASE"/>
+    </target>
+    
+    <target name="build-installer" depends="set-release-build, getProps, doxygen, build-zip" description="Builds Autopsy installer.">
```
eugene7646, line 350: @gdicristofaro We should call "set-release-build" in "build-zip" instead of "build-installer". This way if we are making an Autopsy ZIP for Viking installer, then the flag is also set to "release"

eugene7646 (approved):

PR #7056, opened 2021-06-18
7579 Use transaction for adding derived files

APriestman:


Experimental/src/org/sleuthkit/autopsy/experimental/autoingest/Bundle.properties-MERGED
```diff
@@ -10,6 +10,10 @@ AinStatusNode.status.title=Status
 AinStatusNode.status.unknown=Unknown
 AutoIngestAdminActions.cancelJobAction.title=Cancel Job
 AutoIngestAdminActions.cancelModuleAction.title=Cancel Module
+AutoIngestAdminActions.disableOCR.error=Failed to disable OCR for case "%s".
```
eugene7646, line 13: @APriestman This file probably doesn't belong in this PR?

APriestman: It doesn't, but I think it's just something we haven't gotten around to merging in.

eugene7646 (approved):

PR #7107, opened 2021-07-06
7695-Move-Recyclebin-Extractor-in-Recent-Activity-to-Run-Before-Recent-Documents

markmckinnon:
move run of recycle bin to beginning of the recent activity run


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/RAImageIngestModule.java
```diff
@@ -85,6 +85,7 @@ public void startUp(IngestJobContext context) throws IngestModuleException {
         Extract webAccountType = new ExtractWebAccountType();
         Extract messageDomainType = new DomainCategoryRunner();
 
+        extractors.add(recycleBin); // this needs to run after ExtractRegistry and ExtractOS
```
kellykelly3, line 88: @markmckinnon the comment says that it needs to run after ExtractRegistry.  Is that no longer true?

kellykelly3 (approved):

PR #7117, opened 2021-07-09
7804 initial attempt to make previously seen users obey CR settings

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
```diff
@@ -678,7 +676,8 @@ private OsAccountInstancesAddedTask(CentralRepository db, PropertyChangeEvent ev
 
         @Override
         public void run() {
-            if (!CentralRepository.isEnabled()) {
+            //Nothing to do here if the central repo is not enabled or the ingest is running and the setting to flag previously seen devices and users is not set to true
+            if (!CentralRepository.isEnabled() || (IngestManager.getInstance().isIngestRunning() && !IngestEventsListener.isFlagSeenDevices())) {
```
rcordovano, line 680: So what if ingest is not running? I believe that we want to just go with whatever the IngestEventsListener currently believes, whether ingest is running or not.

rcordovano (changes requested):

Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/CaseEventListener.java
wschaeferB: The decision to make them when ingest is not running is a decision based on previous decisions which were made on 'what to do until we have an artifacts pipeline. and was made for consistency,  I will link the relevant discussions in the related stories in JIRA.

rcordovano (approved):

PR #7139, opened 2021-07-15
User Documentation - Updated complete French translation

Seb2lyon:
Based on the original English version, this updated French translation is totally complete and is the continuation of my previous Pull Request.


bcarrier:
Hi @Seb2lyon. Thanks for this. It might be easiest to maintain if you also checked in the commit number that your translation was updated to. 

For example, 
- If HEAD is currently at B and you find out what changed in the English version. 
- Then more commits happen and we end up at C. 
- Then you do a PR with your changes (which were about the changes up until B). 
- It would be good to know that the French version is accurate up until B, but not C. 

It could be as simple as adding a file into the doxygen_user-fr folder with the commit in it (B in the example above). 

We do something similar for our UI translations. 

https://github.com/sleuthkit/autopsy/blob/develop/release_scripts/localization_scripts/lastupdated.properties

Seb2lyon:
Hi @bcarrier. Thanks for your explanations. I'll take good note of this process for my futur updates.

bcarrier:
This should now be updated each night and published to the site. You can see the latest here: 

https://sleuthkit.org/autopsy/docs/user-docs_fr/4.19.0/

Thanks

PR #7164, opened 2021-07-27
5316 preserve settings between searches

wschaeferB:

PR #7165, opened 2021-07-27
3849 pre select data source for file search

wschaeferB:
This PR includes the code from #7164

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/directorytree/FileSearchProvider.java
+++ b/Core/src/org/sleuthkit/autopsy/directorytree/FileSearchProvider.java
@@ -1,7 +1,7 @@
 /*
  * Autopsy Forensic Browser
  * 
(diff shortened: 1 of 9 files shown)


Core/src/org/sleuthkit/autopsy/directorytree/FileSearchProvider.java
```diff
@@ -23,5 +23,7 @@
  */
 public interface FileSearchProvider {
 
+    public void showDialog(long dataSourceID);
```
kellykelly3, line 26: Do we need both showDialog?  Can we deprecate the old one and have it call the new one with a null value?

wschaeferB: Good idea, I will make that change. Thanks

PR #7169, opened 2021-07-28
Flag unique apps (7862)

eugene7646:
Requires TSK PR: https://github.com/sleuthkit/sleuthkit/pull/2502


Core/src/org/sleuthkit/autopsy/datasourcesummary/datamodel/AnalysisSummary.java
```diff
@@ -49,8 +49,7 @@
     private static final Set<String> EXCLUDED_KEYWORD_SEARCH_ITEMS = new HashSet<>();
 
     private static final Set<Integer> ARTIFACT_UPDATE_TYPE_IDS = new HashSet<>(Arrays.asList(
-            ARTIFACT_TYPE.TSK_INTERESTING_FILE_HIT.getTypeID(),
-            ARTIFACT_TYPE.TSK_INTERESTING_ARTIFACT_HIT.getTypeID(),
+            ARTIFACT_TYPE.TSK_PREVIOUSLY_SEEN.getTypeID(),
```
APriestman, line 52: I'm not sure this change is correct - I think we're still displaying interesting files/results on this panel.

eugene7646: Yep, you're right. I just noticed that. I'll change it along with the PastCasesSummary fix.

APriestman (approved):

PR #7174, opened 2021-07-30
7848 Temporary code to flag matching personas

APriestman:
Depends on https://github.com/sleuthkit/sleuthkit/pull/2506


Core/src/org/sleuthkit/autopsy/centralrepository/eventlisteners/IngestEventsListener.java
```diff
@@ -313,7 +317,72 @@ static private void makeAndPostPreviouslyUnseenArtifact(BlackboardArtifact origi
...
+     * @param attributesForNewArtifact Attributes to assign to the new artifact
+     * @param configuration            The configuration to be specified for the new artifact hit
+     * @param score                    sleuthkit.datamodel.Score to be assigned to this artifact
+     * @param justification            Justification string
+     */
+    private static void makeAndPostPersonaArtifact(BlackboardArtifact.Type newArtifactType, BlackboardArtifact originalArtifact, Collection<BlackboardAttribute> attributesForNewArtifact, String configuration,
+            Score score, String justification) {
+        try {
+            SleuthkitCase tskCase = originalArtifact.getSleuthkitCase();
+            Content originalContent = originalArtifact.getParent(); // Associate artifact with file instead of artifact
```
eugene7646, line 362: @APriestman We don't have to change this PR but just FYI, we have a story to re-write the Artifact.getParent() method because it does 2 database trips when same information can be obtained in 1 database call.

eugene7646 (approved):

PR #7178, opened 2021-08-02
Support processing of multiple reports (7887)

eugene7646:


Core/src/org/sleuthkit/autopsy/commandlineingest/CommandLineIngestSettingsPanel.java
```diff
@@ -288,6 +288,10 @@ private void bnEditReportSettingsActionPerformed(java.awt.event.ActionEvent evt)
         String reportName = getReportName();
         if (reportName.equals(Bundle.CommandListIngestSettingsPanel_Make_Config())) {
             reportName = JOptionPane.showInputDialog(this, Bundle.CommandListIngestSettingsPanel_Report_Name_Msg());
+            
+            // sanitize report name. Remove all commas because in CommandLineOptionProcessor we use commas
+            // to separate multiple report names
+            reportName = reportName.replaceAll(",", "");
```
bcarrier, line 294: I think we should tell the user they can't use commas rather than silently removing them.

eugene7646: @bcarrier I was thinking about that but I think we will need to create a new panel if we want to sanitize the text as the users types it. We are currently using JOptionPane.showInputDialog() to display that window and get user's input. It doesn't allow for listeners, I don't think. But we could just add text that says commas are not allowed.

bcarrier: Documentation is better than nothing.

eugene7646: @bcarrier 
image

gdicristofaro (approved):

PR #7188, opened 2021-08-06
7707 - Moved the content cache up to DisplayableItemNode

kellykelly3:
the dataModel bundle file is included with the PR as the one in the tree is incorrect.


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -96,7 +92,7 @@
      * database to retrieve Content objects that are the source of multiple
      * artifacts.
      */
-    private static final Cache<Long, Content> contentCache = CacheBuilder.newBuilder().expireAfterWrite(1, TimeUnit.MINUTES).build();
+   // private static final Cache<Long, Content> contentCache = CacheBuilder.newBuilder().expireAfterWrite(1, TimeUnit.MINUTES).build();
```
wschaeferB, line 95: commented code

kellykelly3: Good catch. All set.

wschaeferB (approved):
Seems good.

PR #7200, opened 2021-08-16
7879 remove unnecessary MD5 validation and check against empty string

wschaeferB:


wschaeferB:
wrong story number remade with it corrected

PR #7203, opened 2021-08-18
7877 improved discovery search cancellation

wschaeferB:


Core/src/org/sleuthkit/autopsy/discovery/ui/DiscoveryTopComponent.java
```diff
@@ -260,7 +260,6 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
     private void newSearchButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_newSearchButtonActionPerformed
         close();
         final DiscoveryDialog discDialog = DiscoveryDialog.getDiscoveryDialogInstance();
-        discDialog.cancelSearch();
```
wschaeferB, line 263: Removing a call to cancelsearch in a pr to improve cancellation might seem odd, but this one was redundant as the componentClosed method which is called in response to the above call to close is already cancelling the search and covers the use case of the user clicking the X as well.

wschaeferB:
Adding cancellation to further down in the search code.

PR #7235, opened 2021-09-03
7943 view source x

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -447,10 +464,10 @@ public BlackboardArtifact getArtifact() {
          * action to view it in the timeline.
          */
         try {
-            if (ViewArtifactInTimelineAction.hasSupportedTimeStamp(artifact) &&
-                    // don't show ViewArtifactInTimelineAction for AnalysisResults.
+            if (ViewArtifactInTimelineAction.hasSupportedTimeStamp(artifact)
+                    && // don't show ViewArtifactInTimelineAction for AnalysisResults.
                     (!(this.artifact instanceof AnalysisResult))) {
```
kellykelly3, line 469: Switch the checks around. We should only do the hasSupportedTimeStamp check if the artifact is not an AnalysisResult.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -476,10 +493,17 @@ public BlackboardArtifact getArtifact() {
...
          */
-        AbstractFile file = getLookup().lookup(AbstractFile.class
-        );
-        if (null != file) {
-            actionsList.add(ViewFileInTimelineAction.createViewSourceFileAction(file));
+        if (this.srcContent instanceof AbstractFile) {
+            actionsList.add(ViewFileInTimelineAction.createViewSourceFileAction((AbstractFile) this.srcContent));
+        } else if (this.srcContent instanceof DataArtifact) {
+            try {
+                if (ViewArtifactInTimelineAction.hasSupportedTimeStamp((DataArtifact) this.srcContent)) {
```
kellykelly3, line 500: The cast to a is too strong, change to BlackboardArtifact if needed.  Both hasSupportedTimeStamp and ViewArtifactInTimelineAction take BlackboardArtifact

Core/src/org/sleuthkit/autopsy/directorytree/DataResultFilterNode.java
```diff
@@ -240,22 +241,107 @@ private DataResultFilterChildren(Node arg, ExplorerManager sourceEm) {
...
+        "# {0} - contentType",
+        "DataResultFilterNode_getViewSourceDisplayName_baseMessage=View Source {0}",
+        "DataResultFilterNode_getViewSourceDisplayName_type_File=File",
+        "DataResultFilterNode_getViewSourceDisplayName_type_DataArtifact=Data Artifact",
+        "DataResultFilterNode_getViewSourceDisplayName_type_OSAccount=OS Account"
+    })
+    private static Action getContentNavigateAction(Content content) {
+        if (content instanceof DataArtifact) {
+            return new ViewArtifactAction(
+                    (DataArtifact) content,
```
kellykelly3, line 327: The content should be cast to a BlackboardArtifact, not a DataAritfact.

Core/src/org/sleuthkit/autopsy/directorytree/DataResultFilterNode.java
```diff
@@ -240,22 +241,107 @@ private DataResultFilterChildren(Node arg, ExplorerManager sourceEm) {
...
+         * @param displayName The display name of the menu item.
+         */
+        ViewOsAccountAction(OsAccount osAccount, String displayName) {
+            super(displayName);
+            this.osAccount = osAccount;
+        }
+
+        @Override
+        public void actionPerformed(ActionEvent e) {
+            DirectoryTreeTopComponent.findInstance().viewOsAccount(osAccount);
```
kellykelly3, line 304: This call should not be happening on the EDT! Call in SwingWorker, make sure that cursors appear to block any further tree actions.

kellykelly3 (changes requested):

Core/src/org/sleuthkit/autopsy/actions/ViewOsAccountAction.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param displayName The display name of the menu item.
+     */
+    public ViewOsAccountAction(OsAccount osAccount, String displayName) {
+        super(displayName);
+        this.osAccount = osAccount;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        SwingUtilities.invokeLater(() -> DirectoryTreeTopComponent.findInstance().viewOsAccount(osAccount));
```
kellykelly3, line 47: viewOsAccount will cause DB calls.  Why are explicitly forcing it onto the EDT?

gdicristofaro: I can change that.

Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -473,52 +502,359 @@ public BlackboardArtifact getArtifact() {
...
+        
+        // view source content in directory and timeline
+        actionsLists.add(getNonNull(
+                getViewSrcContentAction(this.artifact, this.srcContent),
+                getTimelineSrcContentAction(this.srcContent)
+        ));
+
+        // extract with password from encrypted file
+        actionsLists.add(getNonNull(
+                getExtractWithPasswordAction(this.srcContent)
```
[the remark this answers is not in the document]
rcordovano: Ultimately, yes, but not within the scope of the story behind this PR. We are in the process of looking at a general solution (e.g., perhaps the AutopsyDataModeItem class hierarchy) and I think that we should decide what the general solution is rather than fix these issues piecemeal).

PR #7245, opened 2021-09-08
7944 interesting item tree structure changes

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/InterestingHits.java
```diff
@@ -324,14 +302,13 @@ public SetNameNode(String setName) {//, Set<Long> children) {
...
-            int sizeOfSet = interestingResults.getArtifactIds(setName, BlackboardArtifact.Type.TSK_INTERESTING_ARTIFACT_HIT.getDisplayName()).size()
-                    + interestingResults.getArtifactIds(setName, BlackboardArtifact.Type.TSK_INTERESTING_FILE_HIT.getDisplayName()).size();
+            int sizeOfSet = interestingResults.getArtifactIds(setName).size();
             super.setDisplayName(setName + " (" + sizeOfSet + ")");
         }
 
         @Override
         public boolean isLeafTypeNode() {
-            return false;
+            return true;
```
rcordovano, line 311: A SetNameNode would not be a leaf node because it has analysis result node children, right? It's interesting that it doesn't seem to matter that this has been set to false...it seems like this would break IsLeafItemVisitor...

Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeTopComponent.java
```diff
@@ -1271,9 +1271,10 @@ public void viewArtifact(final BlackboardArtifact art) {
             treeNode = getHashsetNode(typesChildren, art);
         } else if (typeID == BlackboardArtifact.Type.TSK_KEYWORD_HIT.getTypeID()) {
             treeNode = getKeywordHitNode(typesChildren, art);
-        } else if (typeID == BlackboardArtifact.Type.TSK_INTERESTING_FILE_HIT.getTypeID()
-                || typeID == BlackboardArtifact.Type.TSK_INTERESTING_ARTIFACT_HIT.getTypeID()) {
-            treeNode = getInterestingItemNode(typesChildren, art);
+        } else if (typeID == BlackboardArtifact.Type.TSK_INTERESTING_FILE_HIT.getTypeID()) {
+            treeNode = getInterestingItemNode(typesChildren, BlackboardArtifact.Type.TSK_INTERESTING_FILE_HIT, art);
+        } else if (typeID == BlackboardArtifact.Type.TSK_INTERESTING_ARTIFACT_HIT.getTypeID()) {
+            treeNode = getInterestingItemNode(typesChildren, BlackboardArtifact.Type.TSK_INTERESTING_FILE_HIT, art);
```
rcordovano, line 1277: It looks like there is a copy-paste error here. The type in the conditional does not match the type passed to getInterestingItemNode().

rcordovano (changes requested):

Core/src/org/sleuthkit/autopsy/directorytree/DirectoryTreeTopComponent.java
gdicristofaro: My mistake.  Should be corrected now.

Core/src/org/sleuthkit/autopsy/datamodel/InterestingHits.java
rcordovano: Never mind, there is a documentation problem here. I will address it later.

rcordovano (approved):

PR #7279, opened 2021-09-21
7978 os account api change

wschaeferB:


Core/src/org/sleuthkit/autopsy/centralrepository/datamodel/CorrelationAttributeUtil.java
```diff
@@ -102,71 +104,80 @@ private static String getEmailAddressAttrDisplayName() {
...
-            return new ArrayList<>();
-        } else {
-            return makeCorrAttrsForSearch(content);
-        }
+    public static List<CorrelationAttributeInstance> makeCorrAttrsToSave(AbstractFile file) {
+        return makeCorrAttrsForSearch(file);
+    }
+
+    public static List<CorrelationAttributeInstance> makeCorrAttrsToSave(AnalysisResult file) {
+        return new ArrayList<>();
```
rcordovano, line 112: You can return Collections.emptyList() here instead of making a new list.

PR #7282, opened 2021-09-23
8005-IllegalArgumentException-in-Chromium-RA-module

markmckinnon:
Fix illegalArgument Exception with corrupt cache files.

### DIFF
--- a/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
+++ b/RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
@@ -1379,8 +1379,6 @@ final class ChromeCacheExtractor {
                     }
                 } catch (TskCoreException | IngestModuleException ex) {
                     throw new TskCoreException(String.format("Failed to get external key from address %s", longKeyAddresses)); //NON-NLS 
-//                } catch (NullPointerException ex) {
-//                    String name = "a";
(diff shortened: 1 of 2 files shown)


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ChromeCacheExtractor.java
```diff
@@ -592,8 +592,13 @@ private void findExternalFiles(String cachePath) throws TskCoreException {
        
         // see if it is cached
         String fileTableKey = cacheFolderName + cacheFileName;
-        if (cacheFileName.startsWith("f_") && externalFilesTable.containsKey(fileTableKey)) {
-            return Optional.of(externalFilesTable.get(fileTableKey));
+
+        if (cacheFileName != null) {
+            if (cacheFileName.startsWith("f_") && externalFilesTable.containsKey(fileTableKey)) {
```
eugene7646, line 597: @markmckinnon Why are the file names specific to those that start with "f_"? Is the idea to detect carved files? If so, our other carver creates carved files that have different naming structure.

markmckinnon: @eugene7646 the f_* files are extenal files based on the documentation in the code that are used to hold cache information.

eugene7646 (approved):

PR #7291, opened 2021-09-27
8004 os account node host name addition

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/OsAccounts.java
```diff
@@ -374,20 +408,70 @@ public void run() {
...
                     Exceptions.printStackTrace(ex);
                 }
             }
         }
+
+        /**
+         * Data concerning an OS Account loaded asynchronously (and not at sheet
+         * creation).
+         */
+        static class AsynchOsAcctData {
```
rcordovano, line 437: I believe this class can be private.

rcordovano (changes requested):

PR #7296, opened 2021-09-28
8041 show scope for os account node

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/OsAccounts.java
```diff
@@ -208,38 +209,50 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 
-                        updateSheet(new NodeProperty<>(
+                        propertiesToUpdate.add(new NodeProperty<>(
                                 Bundle.OsAccounts_accountHostNameProperty_name(),
                                 Bundle.OsAccounts_accountHostNameProperty_displayName(),
                                 Bundle.OsAccounts_accountHostNameProperty_desc(),
                                 hostsString));
                     }
 
+                    updateSheet(propertiesToUpdate.toArray(new NodeProperty<?>[propertiesToUpdate.size()]));
```
rcordovano, line 255: Do we need to put the updateSheet() calls into the EDT or does NetBeans do that when we call setSheet(createSheet())?

gdicristofaro: I'm not seeing anything pushing the sheet call to the EDT so I wrapped the call in a `SwingUtilities.invokeLater` call.

PR #7297, opened 2021-09-28
8038 fix ra domain name usage

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1726,6 +1732,42 @@ private BlackboardArtifact createAssociatedArtifact(String filePathName, Blackbo
         return map;
     }
     
+    /**
+     * Strip the machine sid off of the osAccountSID. The returned string will
```
markmckinnon, line 1736: Change "machine sid" to "relative identifier" based on https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/security-identifiers

markmckinnon (approved):

PR #7301, opened 2021-09-29
Added comments and notes.

bcarrier:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -1726,6 +1731,7 @@ private BlackboardArtifact createAssociatedArtifact(String filePathName, Blackbo
 
         for(OsAccount account: tskCase.getOsAccountManager().getOsAccounts(((DataSource)dataSource).getHost())) {
             Optional<String> userName = account.getLoginName();
+            // @@@ BC: Seems like this should be calling account.getAddr() to get the SID.
```
kellykelly3, line 1734: This is not new code.  Changing this would effect existing code.

@markmckinnon and I felt like this was the correct thing to be doing.

PR #7306, opened 2021-09-29
Making sure file names don't have spaces at the end (8007)

eugene7646:


Core/src/org/sleuthkit/autopsy/coreutils/FileUtil.java
```diff
@@ -169,8 +169,9 @@ public static String copyFolder(String source, String path, String folderName) t
     public static String escapeFileName(String fileName) {
         //for now escaping /:"*?<>| (not valid in file name, at least on Windows)
         //with underscores. We are only keeping \ as it could be part of the path.
-        return fileName.replaceAll("[\\p{Cntrl}/:\"*?<>|]+", "_");
-    } 
+        // Also trim empty space characters at the end of file name.
+        return fileName.replaceAll("[\\p{Cntrl}/:\"*?<>|]+", "_").trim();
```
rcordovano, line 173: String.trim() also removes leading spaces. That does not match up with the comment.

eugene7646: Yes it does, but I think those do not cause errors in file names, only the trailing spaces do. That was the intent of the comment. I can change it if you want.

eugene7646: I have updated the PR

rcordovano (approved):

PR #7317, opened 2021-09-30
8066 prevent null task being submitted to background thread

wschaeferB:


Core/src/org/sleuthkit/autopsy/datamodel/BlackboardArtifactNode.java
```diff
@@ -891,23 +891,13 @@ protected Sheet createSheet() {
...
         if (artifact instanceof AnalysisResult
                 && !(artifactType.getTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_HASHSET_HIT.getTypeID()
                 || artifactType.getTypeID() == BlackboardArtifact.ARTIFACT_TYPE.TSK_KEYWORD_HIT.getTypeID())) {
             scoTask = updateSheetForAnalysisResult((AnalysisResult) artifact, sheetSet);
         } else {
-            /*
-             * Add the name of the source content of the artifact represented by
-             * this node to the sheet. The value of this property is the same as
-             * the display name of the node and this a "special" property that
-             * displays the node's icon as well as the display name.
```
wschaeferB, line 904: This appeared to be code that accidently got duplicated from lines 882-892 so I have deleted it, if for some reason putting the same thing in the sheet right after we previously added it is necessary then I can undelete it.

APriestman (approved):

PR #7389, opened 2021-10-29
8098 Optimize artifact loading

APriestman:
Depends on https://github.com/sleuthkit/sleuthkit/pull/2595

### DIFF
--- a/Core/src/org/sleuthkit/autopsy/mainui/datamodel/BlackboardArtifactSearchParam.java
+++ b/Core/src/org/sleuthkit/autopsy/mainui/datamodel/BlackboardArtifactSearchParam.java
@@ -1,0 +1,72 @@
+/*
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2021 Basis Technology Corp.
+ * Contact: carrier <at> sleuthkit <dot> org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


Core/src/org/sleuthkit/autopsy/mainui/datamodel/BlackboardArtifactSearchParam.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package org.sleuthkit.autopsy.mainui.datamodel;
+
+import java.util.Objects;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+
+/**
+ * Key for data artifact in order to retrieve data from DAO.
+ */
+public class BlackboardArtifactSearchParam {
```
kellykelly3, line 27: I thought we were moving away from using "BlackboardArtifact" to just "Artifact".  I feel like @rcordovano has given me that comment on occation.

rcordovano: @kellykelly3 and @APriestman, I did provide that guidance in the past, for the sake of shorter names, but that was before we had both BlackboardArtifacts and DataArtifacts. I could still see using Artifact, since BlackboardArtifact has always been a bit awkward for me - I read that as an artifact _of_ the blackboard, but I suppose that it can equally mean artifact _on_ the blackboard. Maybe @bcarrier wants to set a policy here...

kellykelly3 (approved):

PR #7399, opened 2021-11-02
8144 communication counts

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/accounts/Accounts.java
```diff
@@ -329,18 +331,35 @@ Long getCount(String accountType) {
             return types;
         }
 
+        /**
+         * Calculates the total count of accounts based on the values in the
+         * map.
+         *
+         * @return The total count.
+         */
+        Long getTotal() {
```
APriestman, line 340: Is this used?

gdicristofaro: No.  I forgot to take it out.  It's gone now.

APriestman (approved):

PR #7436, opened 2021-11-29
8202 ingest prog bars on EDT; ingest job cancel off EDT

rcordovano:


Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
```diff
@@ -773,51 +784,48 @@ private void startStreamingModeAnalysis() {
...
-                 * to the case database. This number will be used to estimate
-                 * how many files remain to be analyzed as each file ingest task
-                 * is completed.
+                 * For ingest job progress reporting purposes, do a count of the
+                 * files the data source processor has added to the case
+                 * database. This number will be used to estimate how many files
+                 * remain to be analyzed as each file ingest task is completed.
+                 * The estimate will likely be an over-estimate, since some of
+                 * the files will have already been "streamed" to this job and
+                 * processed.
```
APriestman, line 801: estimatedFilesToProcess is not an estimate of how many files are left, it's how many files there are total. Every processed file increments the processedFiles counter regardless of whether we've finished the data source. So it may be that we switch to determinate with estimatedFilesToProcess = 100, but processedFiles is already 30 so we'd set the progress bar to 30%.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
```diff
@@ -1248,100 +1246,185 @@ void addDataArtifacts(List<DataArtifact> artifacts) {
...
+     * file, if the job has not been cancelled, if the job has not been
+     * cancelled.
+     *
+     * @param fileName The name of the file.
+     */
+    private void updateFileIngestProgressForFileTaskStarted(String fileName) {
+        if (usingNetBeansGUI && !jobCancelled) {
+            SwingUtilities.invokeLater(() -> {
+                if (processedFiles <= estimatedFilesToProcess) {
+                    fileIngestProgressBar.progress(fileName, (int) processedFiles);
```
APriestman, line 1374: With the synchronization removed, I believe there is a chance that processedFiles will increase between these two lines, leading to the possibility of processedFiles being greater than the maximum for the progress bar.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
```diff
@@ -534,7 +541,7 @@ boolean hasDataArtifactIngestModules() {
     }
 
     /**
-     * Determnines which inges job stage to start in and starts up the ingest
+     * Determnines which ingets job stage to start in and starts up the ingest
```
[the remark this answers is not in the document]
rcordovano: Cool, I'll revise the comment.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
rcordovano: Cool, I'll revise the comment.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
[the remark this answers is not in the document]
rcordovano: Fixed in next commit.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
rcordovano: Fixed in next commit.

Core/src/org/sleuthkit/autopsy/ingest/IngestJobExecutor.java
rcordovano: A code experiment (passing MAX_INT) revealed that if processedFiles > estimatedFilesToProcess, then the progress just pegs at 100%, no exception is thrown. This is the same thing that happens in the else clause. So this code can actually be simplified. The simplification is in the next commit.

PR #7478, opened 2021-12-17
8250 column ordering

gdicristofaro:


Core/src/org/sleuthkit/autopsy/mainui/datamodel/BlackboardArtifactDAO.java
```diff
@@ -153,11 +154,11 @@
       
 
     TableData createTableData(BlackboardArtifact.Type artType, List<BlackboardArtifact> arts) throws TskCoreException, NoCurrentCaseException {
-        Map<Long, Map<BlackboardAttribute.Type, Object>> artifactAttributes = new HashMap<>();
+        Map<Long, Map<BlackboardAttribute.Type, Object>> artifactAttributes = new LinkedHashMap<>();
```
APriestman, line 157: Can you add a comment that we're using a linked hash map to preserve ordering?

gdicristofaro: Sure.

APriestman (approved):

PR #7513, opened 2022-01-06
1717 add delete analysis result event

kellykelly3:


Core/src/org/sleuthkit/autopsy/mainui/nodes/AnalysisResultNode.java
```diff
@@ -32,7 +32,13 @@
 import org.sleuthkit.autopsy.datamodel.utils.IconsUtil;
 import org.sleuthkit.autopsy.mainui.datamodel.AnalysisResultRowDTO;
 import org.sleuthkit.autopsy.mainui.datamodel.AnalysisResultTableSearchResultsDTO;
+<<<<<<< HEAD
```
eugene7646, line 35: @kellykelly3 This probably shouldn't be here ;)

Core/src/org/sleuthkit/autopsy/mainui/datamodel/AnalysisResultDAO.java
```diff
@@ -855,6 +855,13 @@ void clearCaches() {
 
     @Override
     Set<? extends DAOEvent> processEvent(PropertyChangeEvent evt) {
+        
+        if(evt.getPropertyName().equals(Case.Events.ANALYSIS_RESULT_DELETED.toString())) {
+            this.analysisResultCache.invalidateAll();
```
eugene7646, line 860: @kellykelly3 I think this is probably incorrect. AnalysisResultDAO has 3 caches - analysisResultCache, setHitCache, and keywordHitCache. We either somehow need to check what kind of analysis result it is and invalidate the appropriate cache (e.g. KW hits), or invalidate all of them.

kellykelly3: @eugene7646 You are correct, which is why the deleted keyword was not disappearing.  I am going to invalidate them all. You asked that I look at the keyword and tree issues, which I am currently working on.

eugene7646 (approved):

PR #7526, opened 2022-01-21
Delete batches of analysis results by type and configuration (8275)

eugene7646:


Core/src/org/sleuthkit/autopsy/mainui/nodes/actions/DeleteAnalysisResultSetAction.java
```diff
@@ -0,0 +1,115 @@
...
+                        progressMessage = Bundle.DeleteAnalysisResultsAction_progress_allResults(type.getDisplayName());
+                        warningMessage = Bundle.DeleteAnalysisResultsAction_warning_allResults(type.getDisplayName());
+                    } else {
+                        progressMessage = Bundle.DeleteAnalysisResultsAction_progress_allResultsWithConfiguration(type.getDisplayName(), configuration);
+                        warningMessage = Bundle.DeleteAnalysisResultsAction_warning_allResultsWithConfiguration(type.getDisplayName(), configuration);
+                    }
+ 
+                    int response = JOptionPane.showConfirmDialog(
+                            WindowManager.getDefault().getMainWindow(),
+                            warningMessage,
```
kellykelly3, line 87: @eugene7646 Swing is not thread safe. You cannot do gui stuff in a background thread.

You will need to use SwingUtilities.invokeAndWait for the confirmation dialog.

kellykelly3: These calls might need to be in the EDT too.

Core/src/org/sleuthkit/autopsy/mainui/nodes/actions/DeleteAnalysisResultSetAction.java
```diff
@@ -0,0 +1,115 @@
...
+                            Case.getCurrentCase().getSleuthkitCase().getBlackboard().deleteAnalysisResults(type, dsID, configuration);
+                            logger.log(Level.INFO, "Deleted Analysis Results type = {0}, data source ID = {1}, configuration = {2}", new Object[]{type, dsID, configuration});
+                        } catch (TskCoreException ex) {
+                            logger.log(Level.SEVERE, "Failed to delete analysis results of type = "+type+", data source ID = "+dsID+", configuration = "+configuration, ex);
+                        }
+                    }
+
+                    return null;
+                } finally {
+                    progress.finish();
```
kellykelly3, line 108: This probably needs to be called in EDT

kellykelly3 (changes requested):

PR #7535, opened 2022-01-26
8152 Added check for empty string

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/Firefox.java
```diff
@@ -449,9 +449,12 @@ private void getCookie(long ingestJobId) {
...
 
                 if (checkColumn == true) {
-                    bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DATETIME_CREATED,
-                            RecentActivityExtracterModuleFactory.getModuleName(),
-                            (Long.valueOf(result.get("creationTime").toString())))); //NON-NLS
+                    String value = result.get("creationTime").toString();
+                    if(value != null && !value.isEmpty()) {
+                        bbattributes.add(new BlackboardAttribute(ATTRIBUTE_TYPE.TSK_DATETIME_CREATED,
+                                RecentActivityExtracterModuleFactory.getModuleName(),
+                                (Long.valueOf(result.get("creationTime").toString())))); //NON-NLS
```
eugene7646, line 456: @kellykelly3 Should we do this entire operation (Long.valueOf(result.get("creationTime").toString()) on line 452, so that we know for sure it succeeds? Even if the string is not empty, it might still not be a valid Long?

kellykelly3: If you pass a string that is not a long to Long.valueOf a NumberFormatException will be thrown.

I have mixed feelings about your suggestion.  On the surface you are correct, adding an extra check to see if it's a number makes sense, but is it necessary? Probably not.  The creationTime column in cookies.sqlite is type integer.  I suspect what happened in this case is that a "null" value was interpreted as an empty string.

eugene7646 (approved):

PR #7536, opened 2022-01-26
8276 kwh UI updates

gdicristofaro:


Core/src/org/sleuthkit/autopsy/mainui/datamodel/AnalysisResultConfigSearchParam.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.sleuthkit.autopsy.mainui.datamodel;
+
+import java.util.Objects;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+
+/**
+ * Base class for search params for analysis results that filter by configuration.
+ */
+public class AnalysisResultConfigSearchParam extends AnalysisResultSearchParam {
```
eugene7646, line 27: @gdicristofaro I think we should combine AnalysisResultConfigSearchParam and AnalysisResultSearchParam instead of extending. Basically I think AnalysisResultConfigSearchParam should become AnalysisResultSearchParam - all analysis results should have a concept of configuration, though for some analysis results it will be empty. Do you see any reason not to do this?

eugene7646: Yes, AnalysisResultSearchParam will contain configuration parameter, to support all analysis results that use configuration - interesting items, hash set hits, KWS. There will be no AnalysisResultSetSearchParam class.

eugene7646 (approved):

PR #7550, opened 2022-02-07
8283 Added code to RA to assure an os info object gets created

kellykelly3:


RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -649,28 +651,37 @@ private boolean parseAutopsyPluginOutput(String regFilePath, AbstractFile regFil
...
-                            logger.log(Level.SEVERE, String.format("Error adding installed program artifact to blackboard for file %d.", regFile.getId()), ex); //NON-NLS
-                        }
+                        
+                        osInfo.setOsName(version);
+                        osInfo.setInstalltime(installtime);
+                        osInfo.setSystemRoot(systemRoot);
+                        osInfo.setProductId(productId);
+                        osInfo.setRegOwner(regOwner);
+                        osInfo.setRegOrg(regOrg);
+//                        try {
```
gdicristofaro, line 661: I don't think we need this commented code anymore.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -2356,5 +2346,127 @@ private boolean isDomainIdInSAMList(String osAccountSID) {
...
+                    
+                    postArtifact(createArtifactWithAttributes(BlackboardArtifact.Type.TSK_OS_INFO, context.getDataSource(), bbattributes));
+                }
+            } catch (TskCoreException ex) {
+                logger.log(Level.SEVERE, "Failed to create default OS_INFO artifact", ex); //NON-NLS
+            }
+        }
+
+        void setCompName(String compName) {
+            if(this.compName == null || this.compName.isEmpty()) {
```
gdicristofaro, line 2412: Should these setter clauses be something like `var != null && !var.isEmpty()`?

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -2356,5 +2346,127 @@ private boolean isDomainIdInSAMList(String osAccountSID) {
...
+            }
+        }
+
+        void setRegOwner(String regOwner) {
+            if(this.regOwner == null || this.regOwner.isEmpty()) {
+                this.regOwner = regOwner;
+            }
+        }
+
+        public void setRegOrg(String regOrg) {
```
gdicristofaro, line 2465: This could probably be package private.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
```diff
@@ -701,22 +712,10 @@ private boolean parseAutopsyPluginOutput(String regFilePath, AbstractFile regFil
...
-                            if (results.isEmpty()) {
-                                newArtifacts.add(createArtifactWithAttributes(BlackboardArtifact.Type.TSK_OS_INFO, regFile, bbattributes));
-                            } else {
-                                results.get(0).addAttributes(bbattributes);
-                            }
-                        } catch (TskCoreException ex) {
-                            logger.log(Level.SEVERE, String.format("Error adding installed os_info to blackboard for file %d.", regFile.getId()), ex); //NON-NLS
-                        }
+                        
+                        osInfo.setOsName(os);
```
gdicristofaro, line 716: I think this will get associated with `TSK_PROG_NAME` instead of `TSK_VERSION`.

gdicristofaro (changes requested):

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
kellykelly3: Correct.  That was with intention.  I believe the original code was incorrect.

gdicristofaro: Agreed.  I'm getting more accurate results like this.

RecentActivity/src/org/sleuthkit/autopsy/recentactivity/ExtractRegistry.java
gdicristofaro: Is there an issue taking out the `public` modifier for `setRegOrg` or is that needed for something?

gdicristofaro (changes requested):

gdicristofaro (approved):

PR #7559, opened 2022-02-18
8301 accounts NPE fix

gdicristofaro:


Core/src/org/sleuthkit/autopsy/datamodel/accounts/Accounts.java
```diff
@@ -124,8 +123,8 @@
      *
      * @param skCase The SleuthkitCase object to use for db queries.
      */
-    public Accounts(SleuthkitCase skCase) {
-        this(skCase, 0);
+    public Accounts() {
```
eugene7646, line 126: @gdicristofaro I think this is a public API change, correct? If so, then we need to run this by Brian. Perhaps we should instead deprecate the old APIs and add new ones.

PR #7624, opened 2022-05-10
Update opencv to 3416

kellykelly3:


Core/src/org/sleuthkit/autopsy/coreutils/VideoUtils.java
```diff
@@ -234,6 +239,78 @@ static BufferedImage generateVideoThumbnail(AbstractFile file, int iconSize) {
         }
         return bufferedImage == null ? null : ScalrWrapper.resizeFast(bufferedImage, iconSize);
     }
+    
+    boolean canCompressAndScale(AbstractFile file) {
+        return isVideoThumbnailSupported(file);
+    }
+    
+    static void compressVideo(Path inputPath, Path outputPath, ExecUtil.ProcessTerminator terminator) throws Exception {
```
eugene7646, line 260: These new methods need to be public, right?

Core/src/org/sleuthkit/autopsy/coreutils/VideoUtils.java
```diff
@@ -234,6 +239,158 @@ static BufferedImage generateVideoThumbnail(AbstractFile file, int iconSize) {
         }
         return bufferedImage == null ? null : ScalrWrapper.resizeFast(bufferedImage, iconSize);
     }
+    
+    public boolean canCompressAndScale(AbstractFile file) {
```
eugene7646, line 243: This should be a static method

Core/src/org/sleuthkit/autopsy/coreutils/VideoUtils.java
```diff
@@ -234,6 +239,158 @@ static BufferedImage generateVideoThumbnail(AbstractFile file, int iconSize) {
         }
         return bufferedImage == null ? null : ScalrWrapper.resizeFast(bufferedImage, iconSize);
     }
+    
+    public static boolean canCompressAndScale(AbstractFile file) {
+
+        if (PlatformUtil.getOSName().toLowerCase().startsWith("Windows")) {
```
eugene7646, line 245: This always returns false because first we convert to lower case and then compare to "Windows" with capital "W"

kellykelly3: OMG, you are correct.  I need to find where I copied and pasted it from, but it's there too.

eugene7646 (approved):

PR #7662, opened 2022-08-22
CL processor returns error when invalid report config name is used (8409)

eugene7646:


Core/src/org/sleuthkit/autopsy/commandlineingest/CommandLineIngestManager.java
```diff
@@ -77,7 +77,7 @@ public class CommandLineIngestManager extends CommandLineManager {
 
     static int CL_SUCCESS = 0;
     static int CL_RUN_FAILURE = -1;
-    static int CL_PROCESS_FAILURE = 1;
+    static int CL_PROCESS_FAILURE = -2;
```
kellykelly3, line 80: Shouldn't these be final?

kellykelly3: And private?

PR #7700, opened 2022-10-31
8425 snap & homebrew

gdicristofaro:


eugene7646:
We have decided to keep this as draft PR until we have a better understanding of what it takes to maintain these for every release going forward, and until we have a clear customer or use case that benefits.

Issue #7347, opened 2021-10-14
Enhancement: "Add Data Source" for "Local files and folders" should support removing an individual entry


moshekaplan:
In the "Add Data Source" for "Local files and folders", instead of only having "Add" and "Clear", it should be possible to select an individual entry from the list of files and folders and remove it from the list

![image](https://user-images.githubusercontent.com/810279/137248739-34f365c3-cf46-4d66-9877-8e2c1398087c.png)

markmckinnon:
This is referenced in PR https://github.com/sleuthkit/autopsy/pull/7705

markmckinnon:
Addressed in Release 4.21.0 of Autopsy. If there are any questions/issue please create a new issue or reopen this one.

PR #7705, opened 2022-11-10
Add ability to delete a logical file/directory from DSP panel

markmckinnon:
Add ability to delete a logical filer/directory from DSP panel.  This was requested in https://github.com/sleuthkit/autopsy/issues/7347


markmckinnon:
Sorry referenced the wrong issue should be https://github.com/sleuthkit/autopsy/issues/7347

PR #7733, opened 2023-02-07
Alternative binary viewer

hajdam:
Suggestion for replacement of hex viewer with BinEd library https://bined.exbin.org/library/

preview

If this is not desired, it could be probably converted to plugin instead.

Possible improvements that could be included:

- Reintroduction of removed 4k paging if it is useful and needed feature
- Encoding switching
- Visualization of unprintable characters
- Different layout / grouping of hex codes
- Basic search/find functionality
- Status bar


hajdam:
As there is no response, plugin will be created instead: https://bined.exbin.org/autopsy-plugin/

PR #7744, opened 2023-03-15
added autocompletion class

kellykelly3:


eugene7646:
Merged into develop via separate PR

Issue #7345, opened 2021-10-14
Enhancement: File Search Results should also display search query


moshekaplan:
The option for "Tools-> File Search by Attributes" should display the search criteria in the window, so that a user can easily see what search query generated the output.

varun-khachane:
Hi, I would like to work on this issue. Can you assign it to me?

markmckinnon:
You are now assigned it.

varun-khachane:
@markmckinnon is there any guidelines or templates to make changes in the GUI?

varun-khachane:
@moshekaplan @markmckinnon Can you help me with this issue so that I can raise the PR soon and close this issue. Also can you provide me with some channel where I can communicate with you regarding the errors while solving this issue.

PR #7751, opened 2023-04-27
Changes done in FileSearchPanel.java for issue #7345

varun-khachane:
Added a data structure to store all the data from thee filters which will be used to display it on the panel.


varun-khachane:
@markmckinnon I have raised the PR for issue #7345. Please go through it and share your insights.

varun-khachane:
Can you please tell e why the checks are failing and guide me to resolve them.

markmckinnon:
This is only partial complete from what the request was.  My thought is that they want the search criteria displayed in the GUI for that tab.  I would also not add what the actual query that the code uses but what was entered in the search screen.  Does that make sense?

PR #7785, opened 2023-06-02
7074 Add ContentProvider Infrastructure

gdicristofaro:


Core/src/org/sleuthkit/autopsy/casemodule/CaseMetadata.java
```diff
@@ -614,6 +681,127 @@ private void readFromFile() throws CaseMetadataException {
             throw new CaseMetadataException(String.format("Error reading from case metadata file %s", metadataFilePath), ex);
         }
     }
+    
+    private Element getChildElOrNull(Element parent, String childTag) {
```
bcarrier, line 685: Add JavaDoc

PR #7823, opened 2023-07-20
CT-7122 Module Output and Only Load Valid Provider Cases

gdicristofaro:


Core/src/com/basistech/df/cybertriage/autopsy/CTIntegrationMissingDialog.java
```diff
@@ -0,0 +1,164 @@
...
+import javax.swing.JComponent;
+import org.sleuthkit.autopsy.coreutils.Logger;
+
+/**
+ * Provides directions with how to enable CT integration with Autopsy when
+ * trying to open a CT exported case.
+ */
+public class CTIntegrationMissingDialog extends javax.swing.JDialog {
+
+    private static final String DOCS_PAGE_URL = "http://docs.cybertriage.com";
```
bcarrier, line 35: I'm setting up the online docs so that I can get you a specific URL

Core/src/org/sleuthkit/autopsy/casemodule/Case.java
```diff
@@ -177,6 +178,7 @@ public class Case {
     private static final String CASE_ACTION_THREAD_NAME = "%s-case-action";
     private static final String CASE_RESOURCES_THREAD_NAME = "%s-manage-case-resources";
     private static final String NO_NODE_ERROR_MSG_FRAGMENT = "KeeperErrorCode = NoNode";
+    private static final String CT_PROVIDER_PREFIX = "CTCustomContentProvider_";
```
bcarrier, line 181: This will need to get changed when we rename the provier.

markmckinnon:
@bcarrier the URL fits fine.  


content_provider_url

PR #7964, opened 2025-04-24
Fix export CSV for emails run program artifacts

markmckinnon:
Fix errors with exporting list as a csv for Email and run program artifacts

## Summary by CodeRabbit

* **Bug Fixes**
  * CSV exports now consistently use UTF-8 encoding to prevent character issues.
  * Email artifacts now record attachment presence explicitly as "Yes" or "No" instead of leaving blank values.
  * Attachment attribute label updated to "Has Attachment" for clearer display.

* **Chores**
  * Adjusted execution order of extractors to refine ingest workflow sequencing.


bcarrier:
@markmckinnon Can you resolve the merge conflict? I think it came in from another PR from you I just merged?

PR #7972, opened 2025-07-10
Add Account_Id to insert statement when inserting a record with an Account_Id.

markmckinnon:
Update Account_ID column if the table has an account_id. when the CR record is inserted.  This has not been done since 4.19.x version.  

## Summary by CodeRabbit

* **Chores**
  * Enhanced bulk import logic to adapt per-destination to whether account fields are present, improving robustness and throughput.
* **Bug Fixes**
  * Ensures records with missing or invalid account identifiers are bound and inserted correctly during bulk operations, preventing insertion failures and preserving data integrity.


bcarrier:
@coderabbitai review

bcarrier:
@coderabbitai review

bcarrier:
@markmckinnon What's the context behind this? Wouldn't the DB get upgraded when it is opened and therefore always have the account ID? Are there paths where the upgrade isn't forced?

markmckinnon:
@bcarrier the database had account_id but the column never had any data in it since 4.19.x.   The column will have empty values for anything that was inserted since 4.19.x for tables with account_id.  This fixes any new data that is inserted going forward.

PR #7985, opened 2025-11-03
To create a macOS bundle application

errodriguez:
An image and a BASH script to facilitate creating a macOS bundle so it can be recognised as an application and be launched from Spotlight instead of only running "bin/autopsy" from Terminal.


bcarrier:
@coderabbitai review

